PR #1, opened 2019-04-03
[Student][MBL-11324] Submission Details Skeleton View

JordanMarshall:


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsUpdate.kt
```diff
@@ -39,18 +39,27 @@ class SubmissionDetailsUpdate : UpdateInit<SubmissionDetailsModel, SubmissionDet
...
+                    setOf(SubmissionDetailsEffect.LoadData(model.canvasContext.id, model.assignmentId)))
+            }
             is SubmissionDetailsEvent.SubmissionClicked -> {
-                val submissionType = getSubmissionContentType(
-                        model.rootSubmission?.dataOrNull?.submissionHistory?.find { it?.id == event.submissionId },
+                if (event.submissionAttemptId == model.selectedSubmissionAttemptId) {
+                    Next.noChange<SubmissionDetailsModel, SubmissionDetailsEffect>()
+                } else {
+                    val submissionType = getSubmissionContentType(
+                        model.rootSubmission?.dataOrNull?.submissionHistory?.find { it?.id == event.submissionAttemptId },
```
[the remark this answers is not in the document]
CalvinKern: This also requires an update to the `SubmissionDetailsUpdateTest`

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/ui/SubmissionDetailsView.kt
```diff
@@ -17,34 +17,197 @@
...
+    parent: ViewGroup,
+    private val canvasContext: CanvasContext,
+    private val fragmentManager: FragmentManager
+) : MobiusView<SubmissionDetailsViewState, SubmissionDetailsEvent>(
+    R.layout.fragment_submissiont_details,
+    layoutInflater,
+    parent
+) {
 
+    private var consumer: Consumer<SubmissionDetailsEvent>? = null
```
CalvinKern: Would it be reasonable to move the consumer to MobiusView or will it used that infrequently in subclasses like this that it's not worth it?

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/ui/SubmissionDetailsView.kt
```diff
@@ -17,34 +17,197 @@
...
+        submissionVersionsSpinner.adapter =
+            ArrayAdapter(context, R.layout.spinner_submission_versions, submissions.map { it.second }).apply {
+                setDropDownViewResource(R.layout.spinner_submission_versions_dropdown)
+            }
+        submissionVersionsSpinner.setSelection(selectedIdx, false)
+        submissionVersionsSpinner.onItemSelectedListener = object : AdapterView.OnItemSelectedListener {
+            override fun onNothingSelected(parent: AdapterView<*>?) = Unit
+            override fun onItemSelected(parent: AdapterView<*>?, view: View?, position: Int, id: Long) {
+                val submissionId = submissions[position].first
+                consumer?.accept(SubmissionDetailsEvent.SubmissionClicked(submissionId))
```
CalvinKern: Could change the `submissionId` variable name to `submissionAttempt` to be more consistent.

apps/student/src/main/res/layout/fragment_submissiont_details.xml
```diff
@@ -0,0 +1,193 @@
...
+        android:fillViewport="true">
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:gravity="center"
+            android:orientation="vertical"
+            android:padding="32dp"
+            android:visibility="gone"
+            tools:visibility="visible">
```
CalvinKern: I think you need to either move your visibilities up to the scrollview, or move the errorContainer id to the linear layout. That's the visibility that is getting updated in the view, and this linear layout has no id so it seems like it will never get set to visible

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsPresenter.kt
```diff
@@ -0,0 +1,97 @@
...
+            }
+            tabData += SubmissionDetailsTabData.FileData(
+                name = name,
+                files = this,
+                selectedFileId = model.selectedAttachmentId ?: 0
+            )
+        }
+
+        // Grade/Rubric tab
+        if (rootSubmission.isGraded || rootSubmission.rubricAssessment.isNotEmpty()) {
```
CalvinKern: `isGraded` checks if there is a grade given, not if it is an assignment that will be graded.
I thought we were always going to show the 'grade' tab for assignments that will be graded, but if we decided that we do not need to show this tab when there is not a grade then this is okay. Otherwise, you could pull from the assignment grading type and just check for `NOT_GRADED`?

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionDetailsRenderTest.kt
```diff
@@ -0,0 +1,272 @@
...
+
+    @Test
+    fun spinnerShowsCorrectSelectedSubmission() {
+        val firstSubmission = Submission(
+            attempt = 1,
+            submittedAt = Calendar.getInstance().apply { set(2050, 0, 30, 23, 59, 0) }.time
+        )
+        val secondSubmission = Submission(
+            attempt = 2,
+            submittedAt = Calendar.getInstance().apply { set(2050, 0, 31, 23, 59, 0) }.time
```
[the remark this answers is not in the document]
JordanMarshall: Calendar doesn't have a companion object so we can't add statically-accessible extension functions like that. We could do something like `Calendar.getInstance().getTime(2050, 0, 31, 23, 59, 0)`, but it might make more sense to add new functionality to our existing DateHelper utility, e.g. `DateHelper.makeDate(2050, 0, 31, 23, 59, 0)`.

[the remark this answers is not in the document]
CalvinKern: Ah, that is unfortunate. Up to you if we make the helper, just thought it might be a little more convenient than this pattern.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsUpdate.kt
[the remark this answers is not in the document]
JordanMarshall: Excellent catch.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/ui/SubmissionDetailsView.kt
JordanMarshall: Let's leave it as-is for now until we have a better feel for what becomes the norm in our views. I'd rather deal with the minor duplication until we know it's the right move.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/ui/SubmissionDetailsView.kt
JordanMarshall: 👍

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/ui/SubmissionDetailsView.kt
CalvinKern: Super reasonable

apps/student/src/main/res/layout/fragment_submissiont_details.xml
JordanMarshall: Another excellent catch. This was wrapped in a ScrollView to ensure the user could still access the retry button in landscape or on small devices, but while the ID was moved from the child to the parent, the visibility settings were not. I'll fix that and also update the UI test to assert that the content itself is also visible, not just the container.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsPresenter.kt
JordanMarshall: After discussing this with David it was decided that the grade tab will only be shown when there is a grade or rubric assessment.

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionDetailsRenderTest.kt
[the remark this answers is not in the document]
JordanMarshall: It will undoubtedly be helpful for future tests so I'll go ahead and make it.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsPresenter.kt
JordanMarshall: After further discussion with product and design, we settled on showing all the tabs all the time to keep things consistent for the user.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/ui/SubmissionDetailsView.kt
```diff
@@ -38,15 +38,15 @@ import com.instructure.student.mobius.assignmentDetails.submissionDetails.Submis
...
+import kotlinx.android.synthetic.main.fragment_submission_details.*
 
 class SubmissionDetailsView(
     layoutInflater: LayoutInflater,
     parent: ViewGroup,
     private val canvasContext: CanvasContext,
     private val fragmentManager: FragmentManager
 ) : MobiusView<SubmissionDetailsViewState, SubmissionDetailsEvent>(
-    R.layout.fragment_submissiont_details,
+    R.layout.fragment_submission_details,
```
CalvinKern: lol, nice catch

PR #3, opened 2019-04-03
Fix open source builds

bootstraponline:


open_source.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -ex
+
+# Open source users: Run this script before building.
```
CalvinKern: Is this going to be clear to people to run this or should we include a note about this in the readme?

open_source_data/soseedygrpc/files.list
```diff
@@ -0,0 +1,6 @@
+# Private info for iOS data seeding server
```
joehoag: Should this say "Android" rather than "iOS"?

open_source.sh
bootstraponline: That's a good point, I'll updated the readme.

open_source_data/soseedygrpc/files.list
bootstraponline: There's some weird history about `soseedygrpc`

tl;dr it was only ever used on iOS and isn't actually used anymore by anyone

PR #4, opened 2019-04-04
Fix open source builds

bootstraponline:


open_source_data/student/google-services.json
```diff
@@ -6,7 +6,7 @@
   "client": [
     {
       "client_info": {
-        "mobilesdk_app_id": "2",
+        "mobilesdk_app_id": "1:111111111111:android:1111111111111111",
```
joehoag: That's not a real id, is it?

bootstraponline: Google crashes unless the fake id matches a certain pattern. The id is still fake.

PR #5, opened 2019-04-05
[Student][Teacher][MBL-12193] Preview HTML files

CalvinKern:
Instead of downloading HTML files, preview them in a webview so that we 
can handle relative files that are referenced.

See Jira ticket for test plan.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/router/RouteMatcher.kt
+++ b/apps/student/src/main/java/com/instructure/student/router/RouteMatcher.kt
@@ -256,12 +256,11 @@ object RouteMatcher : BaseRouteMatcher() {
                 handleFullscreenRoute(context, route)
             } else {
                 val isGroupRoute = "groups" == route.uri?.pathSegments?.get(0)
-                if (route.queryParamsHash.containsKey(RouterParams.PREVIEW)) {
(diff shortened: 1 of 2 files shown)


apps/student/src/main/java/com/instructure/student/router/RouteMatcher.kt
```diff
@@ -354,15 +353,37 @@ object RouteMatcher : BaseRouteMatcher() {
...
+        if (activity == null) {
+            return
+        }
+
+        // If we're trying to open an HTML file, don't download it. It could be referencing other files
+        // through a relative URL which we won't be able to access. Instead, just showing the file in
+        // a webview will load the file the user is trying to view and will resolve all relative paths
+        if (filename.toLowerCase().endsWith(".htm") || filename.toLowerCase().endsWith(".html")) {
+            var needsAuth = true
+            var htmlUrl = "${ApiPrefs.protocol}://${ApiPrefs.domain}"
```
Blackchai: Super not a big deal, but you could swap this for ApiPrefs.fullDomain

CalvinKern: Super done

PR #9, opened 2019-04-11
[Teacher][MBL-12224] Add mobius and modules moflow to teacher

Blackchai:
For now, I've just duplicated things over to the teacher app, removing student specific elements. I'll also create a tech debt ticket to help us remember to take another look at our common mobius code and see what should be pulled out into a new or existing module.

### DIFF
--- a/apps/teacher/src/main/java/com/instructure/teacher/mobius/common/modules/ModulesListModels.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/mobius/common/modules/ModulesListModels.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2019 - present Instructure, Inc.
+ *
+ *     This program is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU General Public License as published by
+ *     the Free Software Foundation, version 3 of the License.
+ *
+ *     This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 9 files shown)


apps/teacher/src/main/java/com/instructure/teacher/mobius/common/modules/ModulesListModels.kt
```diff
@@ -0,0 +1,38 @@
...
+package com.instructure.teacher.mobius.common.modules
+
+import com.instructure.canvasapi2.models.CanvasContext
+import com.instructure.canvasapi2.models.ModuleItem
+import com.instructure.canvasapi2.utils.DataResult
+
+sealed class ModulesListEvent {
+    object PullToRefresh : ModulesListEvent()
+    data class ModuleClicked(val moduleItem: ModuleItem) : ModulesListEvent()
+    data class DataLoaded(val modulesResult: DataResult<ModuleItem>) : ModulesListEvent()
```
JordanMarshall: Are we planning to exhaust pagination for this?

Blackchai: Oh right, thanks

JordanMarshall: I think we'll tackle this at a later point.

PR #10, opened 2019-04-12
[Student][MBL-11328] View URL submissions

JordanMarshall:


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/UrlSubmissionFragment.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    override fun onStart() {
+        super.onStart()
+        Glide.with(requireContext()).load(previewUrl).into(urlPreviewImageView)
+        urlTextView.text = url
+        urlTextView.setTextColor(ThemePrefs.buttonColor)
+        val launchUrl = { _: View ->
+            val title = getString(R.string.urlSubmission)
+            val intent = InternalWebViewActivity.createIntent(requireActivity(), url, title, true)
```
CalvinKern: There is an issue opening links that are *not* https, should we support non secure sites?
I made a submission with url http://www.google.com and everything appears fine, but when I click on the url to open it in an interactive webview it shows a load error for not showing clear text. Seems like we should allow non https sites to be clicked on, but I could see it being a business requirement to enforce secure connections.

JordanMarshall: Hmmm, that sounds like a product/business decision; we'll want to open a dialog about that. 
I'm assuming the change would happen in `InternalWebViewActivity` and that the current behavior affects more than just this feature, so it's likely not a blocker for this PR.

PR #11, opened 2019-04-12
Cloud build metrics docs

bootstraponline:


bootstraponline:
> Might we want to move the cloud_build_metrics folder to some more centralized location, like the mobile_shared repo?

I have thought about that. I'd want to discuss with the team first. This code doesn't really have an owner right now so I'm worried about moving it off repo. `espresso` in `mobile-shared` is actually used by all the Android apps.

PR #13, opened 2019-04-16
[Student][MBL-11330] View Text Submissions

JordanMarshall:


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/TextSubmissionFragment.kt
```diff
@@ -0,0 +1,110 @@
...
+        var formatted = textSubmissionWebView.formatHtml(src)
+
+        /* If the source content begins with a paragraph tag, the WebView automatically applies some vertical padding.
+           For other content, we need to apply the padding ourselves. */
+        val verticalPadding = if (src.startsWith("<p")) "0px" else "8px"
+
+        // Set padding by updating the relevant CSS properties
+        formatted = formatted.replaceFirstAfter("html,body", "padding: 0;", "padding: $verticalPadding 8px;")
+        return formatted.replaceFirstAfter("#content", "padding: 0px 10px 10px;", "")
+    }
```
CalvinKern: This seems like a useful function that we would want to reuse. Maybe make another function on the webview to adding padding if needed?

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/TextSubmissionFragment.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    override fun onStop() {
+        super.onStop()
+        textSubmissionWebView.stopLoading()
```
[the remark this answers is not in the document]
JordanMarshall: No, but `loadRawHtml` is called on onStart which will load the page again. afaik there's no way to resume loading once stopped.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/TextSubmissionFragment.kt
JordanMarshall: CanvasWebView is certainly used in enough places that it's worth looking into. In this case the code was built for the relatively controlled content we'd expect to see in a text submission, so I'm not sure how well it would translate into a general solution without further testing.

CalvinKern: Reasonable.

PR #16, opened 2019-04-17
[Student][MBL-11325] Submission content empty state

StephenBrough:
See [ticket](https://instructure.atlassian.net/browse/MBL-11325) for testing


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/SubmissionDetailsEmptyEffectHandler.kt
```diff
@@ -0,0 +1,27 @@
...
+ */
+package com.instructure.student.mobius.assignmentDetails.submissionDetails.content.emptySubmission
+
+import com.instructure.student.mobius.assignmentDetails.submissionDetails.content.emptySubmission.ui.SubmissionDetailsEmptyContentView
+import com.instructure.student.mobius.common.ui.EffectHandler
+
+class SubmissionDetailsEmptyEffectHandler : EffectHandler<SubmissionDetailsEmptyContentView, SubmissionDetailsEmptyEvent, SubmissionDetailsEmptyEffect>() {
+    override fun accept(effect: SubmissionDetailsEmptyEffect) {
+        when(effect) {
+            is SubmissionDetailsEmptyEffect.ShowSubmitDialogView -> view?.showSubmitDialogView(effect.assignmentId, effect.course)
```
CalvinKern: This should probably be the whole assignment passed to the view, not just the id. That's what was needed for the dialog in the AssignmentDetailsView (which we should probably try to figure out how to share between these two screens?)

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/SubmissionDetailsEmptyUpdate.kt
```diff
@@ -0,0 +1,41 @@
...
+    override fun performInit(model: SubmissionDetailsEmptyContentModel): First<SubmissionDetailsEmptyContentModel, SubmissionDetailsEmptyEffect> {
+        return First.first(model, setOf())
+    }
+
+    override fun update(model: SubmissionDetailsEmptyContentModel, event: SubmissionDetailsEmptyEvent): Next<SubmissionDetailsEmptyContentModel, SubmissionDetailsEmptyEffect> {
+        return when(event) {
+            SubmissionDetailsEmptyEvent.SubmitAssignmentClicked -> {
+                // If a user is trying to submit something to an assignment and the assignment is null, something is terribly wrong
+                val submissionTypes = model.assignment.getSubmissionTypes()
+                if(submissionTypes.size == 1) {
```
CalvinKern: You'll need just a bit more logic here, we would have to show the dialog in the case of online_upload being the online type IF the canvas instance has Arc enabled. i.e.,
`if(submissionTypes.size == 1 && !(submissionTypes.contains(Assignment.SubmissionType.ONLINE_UPLOAD) && model.isArcEnabled)) {`

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/ui/AssignmentDetailsView.kt
```diff
@@ -131,6 +130,7 @@ class AssignmentDetailsView(
     }
 
     private fun renderLoadedState(state: AssignmentDetailsViewState.Loaded) {
+        submissionRubricButton.setVisible(true) // Moved this here from the render function so it doesn't show up at the very top of view while things are loading
```
CalvinKern: Sorry bud, but it still shows up on top of everything while loading (pull to refresh shows this really well)

apps/student/src/main/res/layout/fragment_assignment_details.xml
```diff
@@ -412,9 +412,10 @@
                     android:gravity="center_vertical"
                     android:orientation="horizontal"
                     android:paddingStart="16dp"
-                    android:paddingEnd="16dp">
+                    android:paddingEnd="16dp"
+                    android:visibility="gone">
 
-                    <TextView
+                <TextView
```
CalvinKern: I think the indentation on this was correct before...

apps/student/src/main/res/layout/fragment_submission_details_empty_content.xml
```diff
@@ -0,0 +1,103 @@
...
+        android:layout_height="wrap_content"
+        android:orientation="vertical"
+        app:layout_constraintGuide_percent="0.05" />
+
+    <androidx.constraintlayout.widget.Guideline
+        android:id="@+id/textRight"
+        android:layout_width="1dp"
+        android:layout_height="wrap_content"
+        android:orientation="vertical"
+        app:layout_constraintGuide_percent="0.95" />
```
CalvinKern: Interesting. I love the use of the constraint layout components here, but curious why the choice for doing percentage margins rather than dp margins?

apps/student/src/main/res/layout/fragment_submission_details_empty_content.xml
```diff
@@ -0,0 +1,103 @@
...
+    <Button
+        android:id="@+id/submitButton"
+        style="@style/ButtonStyle"
+        android:layout_width="248dp"
+        android:text="@string/submitAssignment"
+        android:visibility="invisible"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintHorizontal_bias="0.496"
+        app:layout_constraintLeft_toLeftOf="parent"
+        app:layout_constraintRight_toRightOf="parent" />
```
CalvinKern: Is this horizontal bias needed? It should be able to center it correctly based off the left and right constraints. Also, is there a reason for not using the guidelines for the left and right constraints instead of parent?

apps/student/src/test/java/com/instructure/student/test/assignment/details/AssignmentDetailsPresenterTest.kt
```diff
@@ -312,7 +312,7 @@ class AssignmentDetailsPresenterTest : Assert() {
     }
 
     @Test
-    fun `Submit button reads "Submit Assignment" if there are no existing subnmissions`() {
+    fun `Submit button reads "Submit Assignment" if there are no existing submissions`() {
```
CalvinKern: I prefer the typo 🙃

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/SubmissionDetailsEmptyContentPresenter.kt
```diff
@@ -0,0 +1,124 @@
...
+    return format(pattern)
+}
+
+fun OffsetDateTime.getTime(context: Context): String {
+    val amPm = hashMapOf(
+            0L to context.getString(R.string.lowercaseAM),
+            1L to context.getString(R.string.lowercasePM)
+    )
+
+    val pattern = DateTimeFormatterBuilder().appendPattern("h:mm").appendText(ChronoField.AMPM_OF_DAY, amPm).toFormatter()
```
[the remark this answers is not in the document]
StephenBrough: I'll run this by Dave and see what he thinks.

As for the weird map thing, that's what ThreeTen is doing behind the scenes. Not sure if that's for localization, but I'm pretty sure we're doing it the way you've written elsewhere and we haven't had any problems, so I'll just make the switch to be consistent 👍

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/SubmissionDetailsEmptyEffectHandler.kt
StephenBrough: Huh, I thought I copied that directly from the Assignment details view. I'll make that change!

CalvinKern: Ah, I had changed it when I did the dialog view when I realized we hadn't accounted for Arc. So you probably just got it before the change was made.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/ui/AssignmentDetailsView.kt
StephenBrough: NOOOO! I thought for sure... I'll take a look.

CalvinKern: Looks like your other PR has resolved this. You should be able to just remove this line now from this PR. :)

apps/student/src/main/res/layout/fragment_submission_details_empty_content.xml
StephenBrough: Hah, it was actually a copy-paste from either Student or Teacher. Funny, we do different things between the apps and one of them uses ConstraintLayout.

CalvinKern: lol, nice. Not sure what the intent behind percentage was in the first place, but we may want to have strict dp margin instead:
`app:layout_constraintGuide_begin="16dp"`
Not sure if that gives the same thing, but it would at least be consistent across different device sizes. Not sure how much of an issue this is though, I'll let you decide what has to be done.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/SubmissionDetailsEmptyUpdate.kt
StephenBrough: Hmm... this one is a bit of work to get `isArcEnabled` passed all the way down. Kind of gross, but I could pass it through Submission details and then into the empty Submission content. Might also make sense to put it on the Assignment object... thoughts?

apps/student/src/main/res/layout/fragment_submission_details_empty_content.xml
StephenBrough: Nope! I think I may have messed with the visual editor a little and made some of these changes. Changing back!

apps/student/src/main/res/layout/fragment_submission_details_empty_content.xml
StephenBrough: I think consistency is good. I'll change it to 16dp.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/SubmissionDetailsEmptyUpdate.kt
CalvinKern: I think adding a method to assignment to determine `isArcEnabled` is reasonable. Then we can reuse that logic wherever we need it.

StephenBrough: It looks like getting that field requires an API call. We might not want to do that every time we access the field. I can add a field to Assignment that will get assigned when we first get it in the Assignment details effect handler.

PR #17, opened 2019-04-18
[Teacher][Accessibility][MBL-11805, MBL-12179, MBL-11802, MBL-11808]: Accessibility fixes

joehoag:
There are a few things going on here:

- Now excluding/ignoring accessibility issues with pspdfkit and date/time picker.  This required a fairly horrendous hack.
- Mopped up a few random accessibility failures.
- Accessibility testing now enabled by default for Teacher tests.  (UPDATE: Backed off of this.)

I also created tickets https://instructure.atlassian.net/browse/MBL-12258 and https://instructure.atlassian.net/browse/MBL-12259 for us to revisit / actually address the accessibility problems with pspdfkit and date/time picker at a later date if we so choose.  But ignoring them for now allows us to enable accessibility testing by default for the teacher app.

### DIFF
--- a/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/utils/TeacherTest.kt
+++ b/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/utils/TeacherTest.kt
@@ -20,12 +20,23 @@ import com.instructure.canvas.espresso.CanvasTest
 import com.instructure.teacher.BuildConfig
 import com.instructure.teacher.activities.LoginActivity
 import com.instructure.teacher.ui.pages.*
+import org.junit.Before
 
 abstract class TeacherTest : CanvasTest() {
 
     override val activityRule = TeacherActivityTestRule(LoginActivity::class.java)
 
     override val isTesting = BuildConfig.IS_TESTING
+
+    // Overriding this to turn on accessibility tests for all teacher tests
+    @Before
+    override fun preLaunchSetup() {
+        // Enable accessibility checks
+        enableAndConfigureAccessibilityChecks()
+
+        // Call the super method
+        super.preLaunchSetup()
+    }
 
     /**
      * Required for auto complete of page objects within tests
--- a/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/utils/TeacherTest.kt
+++ b/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/utils/TeacherTest.kt
@@ -28,16 +28,6 @@ abstract class TeacherTest : CanvasTest() {
 
     override val isTesting = BuildConfig.IS_TESTING
 
-    // Overriding this to turn on accessibility tests for all teacher tests
-    @Before
-    override fun preLaunchSetup() {
-        // Enable accessibility checks
-        enableAndConfigureAccessibilityChecks()
-
-        // Call the super method
-        super.preLaunchSetup()
-    }
-
     /**
      * Required for auto complete of page objects within tests
      */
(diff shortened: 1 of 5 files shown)


joehoag:
Looks like we have massive teacher UI test failures.  Looking into it...

joehoag:
Well, this is bizarre... I'm getting a bunch of contrast-related accessibility issues on the FTL device that I didn't get locally.  Stuff like this:

```AppCompatTextView{id=2131296927, res-name=noInstructionsTextView, visibility=VISIBLE, width=328, height=59, has-focus=false, has-focusable=false, has-window-focus=true, is-clickable=true, is-enabled=true, is-focused=false, is-focusable=false, is-layout-requested=false, is-selected=false, layout-params=android.widget.LinearLayout$LayoutParams@bcb4f3e, tag=null, root-is-layout-requested=false, has-input-connection=false, x=16.0, y=43.0, text=Help your students with this assignment by adding instructions., input-type=0, ime-target=false, has-links=false}: TextView does not have required contrast of 4.500000. Actual contrast is 3.670513```

:-(.  I may need to back off of turning on accessibility checking by default in the teacher app until I can get a handle on this.

**Comment by joehoag** [comment]
OK, I've backed off of turning on accessibility checking by default for the teacher app.  I'll create a ticket to chase the accessibility issues found on the FTL device.

File: automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/CanvasTest.kt [hunk 1]
```diff
@@ -26,13 +32,68 @@ abstract class CanvasTest : InstructureTest() {
...
+            }
+
+            // Matches if the view being tested, or any of its parents going up 5
+            // levels, has any of the excluded strings in its resource name.
+            override fun matches(item: Any): Boolean {
+                when(item) {
+                    is AccessibilityViewCheckResult -> {
+                        var view = item.view
+                        var upCount = 0
+                        while(view != null && upCount < 5)
```
CalvinKern: What is the reason for checking only 5 levels up?
May be more robust to use `withParent(isExcludedNamedView(listOf(...)))` to capture all the way up the hierarchy and just have this matches function check if the current view's resource name matches.

apps/teacher/src/androidTest/java/com/instructure/teacher/ui/utils/TeacherTest.kt
```diff
@@ -20,6 +20,7 @@ import com.instructure.canvas.espresso.CanvasTest
 import com.instructure.teacher.BuildConfig
 import com.instructure.teacher.activities.LoginActivity
 import com.instructure.teacher.ui.pages.*
+import org.junit.Before
```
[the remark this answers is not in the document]
joehoag: Maybe.  I'll remove it if it is.

automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/CanvasTest.kt
joehoag: The 5 levels proved to be enough to do the job.  And I wanted to avoid checking the whole hierarchy, just for reasons of efficiency.

If we did decide to scan all ancestors, is withParent() recursive?  I thought that withParent() only checked the immediate parent.

CalvinKern: Hmm, that's a good point, I don't think withParent() is recursive from the wording. If you're confident that 5 levels should catch everything we need then I'm okay with it.

joehoag:
Thanks for the review!

PR #18, opened 2019-04-18
[Student][MBL-12211] Force refresh on initial load of grades screen

CalvinKern:
There was an issue where assignment group weights were not being refreshed (except by unchecking the ‘what if’ grade box) and caused the grades screen to use stale weights for a long time.

There is a 'weighted course' in my sandbox that can be used to your hearts content while testing.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/adapter/GradesListRecyclerAdapter.kt
+++ b/apps/student/src/main/java/com/instructure/student/adapter/GradesListRecyclerAdapter.kt
@@ -159,6 +159,7 @@ open class GradesListRecyclerAdapter(
                 }
 
                 if (isAllGradingPeriodsSelected) {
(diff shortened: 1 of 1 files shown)


StephenBrough:
Is there a specific path to repro the issue? I logged into your sandbox, changed a grade weight, and (after doing a pull to refresh), the app showed the same grade as web.

CalvinKern:
lol, I didn't realize that this screen had a pull to refresh. Maybe not worth fixing then?
The way I could repro was by opening the app, looking at the grades, killed the app, changed the grade weights on web, opened the app and was still getting the old values.

StephenBrough:
Killing the app and reopening it should pull down new values, so I'd say this is still valid. I'll try to repro it using those steps.

PR #27, opened 2019-04-25
[Parent][Accessibility][MBL-12290]: Fix accessibility issues from Parent app UI tests

joehoag:
This will also enable accessibility testing by default for all apps.


apps/parent/src/main/java/com/instructure/parentapp/activity/NavigationActivity.kt
```diff
@@ -308,6 +308,11 @@ class NavigationActivity : BaseParentActivity(), MasqueradingDialog.OnMasqueradi
                 adapter = UserSpinnerAdapter(this@NavigationActivity, users)
                 setSelection(ParentPrefs.selectedStudentIdx)
             }
+
+            // Give the spinner a minimum height of 48dp.
+            // Unfortunately, I didn't see a way to do this in the layout file.
```
StephenBrough: Oddly enough, it looks like we don't have the spinner in any layout file anywhere. It's always being added dynamically.

PR #63, opened 2019-05-21
[Student][MBL-12283] Rubric Tooltip View

JordanMarshall:


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/drawer/rubric/ui/SubmissionRubricTooltipPopup.kt
```diff
@@ -0,0 +1,164 @@
...
+            val rootView = anchorView.rootView
+            val lp = rootView.layoutParams
+            if (lp is WindowManager.LayoutParams && lp.type == WindowManager.LayoutParams.TYPE_APPLICATION) {
+                // This covers regular app windows and Dialog windows.
+                return rootView
+            }
+            // For non-application window types (such as popup windows) try to find the main app window
+            // through the context.
+            var context = anchorView.context
+            while (context is ContextWrapper) {
```
CalvinKern: This seems dangerous. 
Is there a guarantee that `context.baseContext` will eventually get to an activity or something that isn't a ContextWrapper?
If not, we may just want a safety counter to make sure we don't loop infinitely.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/drawer/rubric/ui/SubmissionRubricTooltipPopup.kt
```diff
@@ -0,0 +1,164 @@
...
+        outParams.token = anchorView.applicationWindowToken
+
+        val offsetX = anchorView.width / 2
+
+        val offsetBelow = anchorView.height  // Place below the view in most cases.
+        val offsetAbove = 0  // Place above the view if the tooltip does not fit below.
+
+        outParams.gravity = Gravity.START or Gravity.TOP
+
+        val tooltipOffset = context.DP(5).toInt()
```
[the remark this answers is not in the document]
JordanMarshall: They serve different purposes but happen to be the same value. This one defines the offset of the tooltip from the anchor view while the other defines the dimensions of the tooltip tail. Regardless, I'll refactor this into a member property. We _could_ turn it into a dimension resource but afaik this will be the same value for all configurations so using a resource would just add unnecessary overhead.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/drawer/rubric/ui/SubmissionRubricTooltipPopup.kt
JordanMarshall: Aside from being converted to Kotlin, this function is unmodified from the original androidx [source](https://github.com/aosp-mirror/platform_frameworks_support/blob/237c8946756af4b0fe9d0fa3965593e247d53698/appcompat/src/main/java/androidx/appcompat/widget/TooltipPopup.java#L198). Regarding your concern, the only way I could see this getting into an infinite loop is if we somehow had an infinite loop of ContextWrappers, which I'm guessing would break the app long before this code could get executed.

CalvinKern: Huh. Well I'll be. If it's good enough for androidx source then it's good enough for us!

PR #64, opened 2019-05-22
[Student][MBL-12283] Rubric Presenter

JordanMarshall:


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/drawer/rubric/SubmissionRubricPresenter.kt
```diff
@@ -17,10 +17,96 @@
...
+
+    private const val customRatingId = "_custom_rating_id_"
+
     override fun present(model: SubmissionRubricModel, context: Context): SubmissionRubricViewState {
-        TODO()
+        val rubric = model.assignment.rubric ?: emptyList()
+        val assessments = model.submission.rubricAssessment
+
+        // Show empty state if the assignment does not use a rubric
+        if (rubric.isEmpty()) return SubmissionRubricViewState(listOf(RubricListData.Empty))
```
CalvinKern: Just want to make sure that this rubric pane will only ever contain rubric grade data, so we won't show any assignment grades if there's no rubric even if there's a grade?
Or is there going to be a different presenter for normal grades that aren't in a rubric?

JordanMarshall: After a lengthy discussion with product and design it was decided that this will exclusively be a Rubric tab and there will not be a grade tab.

PR #65, opened 2019-05-23
[Student][MBL-11327] Add discussion content to new assignment flow

CalvinKern:


CalvinKern:
Noticed there is a problem with linking to a group discussion from the discussion content. Will fix in this PR.

PR #71, opened 2019-05-29
[Student][MBL-12275] Wire up persistence to assignment submissions

StephenBrough:


libs/pandautils/src/main/java/com/instructure/pandautils/dialogs/UploadFilesDialog.kt
```diff
@@ -55,7 +55,6 @@ import org.greenrobot.eventbus.EventBus
 import org.greenrobot.eventbus.Subscribe
 import java.io.File
 import java.util.*
-import kotlin.collections.ArrayList
```
CalvinKern: Best change in the PR 🙃

PR #73, opened 2019-05-29
[All][MBL-12449] Remove special 504 handling from StatusCallback

JordanMarshall:
Repro:
- Install apps (student, teacher, and parent)
- Log into an account
- Kill the app and clear its cache
- Enable airplane mode on the device
- Open the app
	
Old behavior:
- Student: Only shows navigation tabs and blank pages
- Teacher: Pages load indefinitely
- Parent: Shows splash screen indefinitely

New behavior:
- Student: All pages display. Some still load indefinitely, others display empty state
- Teacher: Dashboard shows empty state, other pages still load indefinitely
- Parent: Still shows splash screen indefinitely

### DIFF
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/weave/ApiAsync.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/weave/ApiAsync.kt
@@ -40,27 +40,12 @@ fun <T> apiAsync(managerCall: ManagerCall<T>): Deferred<DataResult<T>> {
 
     val callback = object : StatusCallback<T>() {
 
-        var succeededOrFailed = false
-
         override fun onResponse(response: Response<T>, linkHeaders: LinkHeaders, type: ApiType) {
-            succeededOrFailed = true
-            if (response.isSuccessful) {
-                @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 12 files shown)


libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/weave/ApiAsync.kt
```diff
@@ -40,27 +40,12 @@ fun <T> apiAsync(managerCall: ManagerCall<T>): Deferred<DataResult<T>> {
...
-                deferred.complete(DataResult.Fail(failure))
-            }
-        }
-
-        override fun onFinished(type: ApiType) {
-            if (!succeededOrFailed && type != ApiType.CACHE) {
-                deferred.complete(DataResult.Fail(Failure.Network("StatusCallback: 504 Error")))
-            }
+            @Suppress("UNCHECKED_CAST")
+            deferred.complete(DataResult.Success(response.body() as T))
```
CalvinKern: Do we need to handle passing back data result failures here or does `onResponse` only get called for successful requests?

JordanMarshall: `onResponse` only gets called for successful requests

PR #80, opened 2019-05-30
[Student][MBL-12460] Make a11y descriptions around edit dashboard better

StephenBrough:


apps/student/src/main/res/layout/fragment_favoriting.xml
```diff
@@ -31,6 +31,12 @@
         app:popupTheme="@style/ToolBarPopupStyle"
         app:theme="@style/ToolBarStyle"/>
 
+    <TextView
+        android:id="@+id/editDashboardDescriptor"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="@string/editDashboardDescriptor"/>
+
```
CalvinKern: Were there designs for this? Looks a little out of place to me. Would look better with some padding and larger text maybe?

StephenBrough: Yes and yes. The ticket has the design spec and I just pushed up something that is about as close as I can get to it.

PR #93, opened 2019-06-11
[Teacher][MBL-11773] Fix dupe PTR indicator in people list

Blackchai:


apps/teacher/src/main/java/com/instructure/teacher/fragments/PeopleListFragment.kt
```diff
@@ -188,12 +188,15 @@ class PeopleListFragment : BaseSyncFragment<User, PeopleListPresenter, PeopleLis
 
     override fun onRefreshFinished() {
         swipeRefreshLayout.isRefreshing = false
-        emptyPandaView.visibility = View.GONE
+//        emptyPandaView.visibility = View.GONE
```
CalvinKern: Should probably just delete this instead of commenting it out.

Blackchai: Ugghhh. I commented it out to confirm that it wasn't necessary and then forgot to delete after testing it 🤦‍♂ ... Thanks Calvin.

PR #94, opened 2019-06-12
[Student][MBL-11332] External tools submission view

StephenBrough:


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsEffectHandler.kt
```diff
@@ -38,15 +38,18 @@ class SubmissionDetailsEffectHandler : EffectHandler<SubmissionDetailsView, Subm
...
+            val submission = SubmissionManager.getSingleSubmissionAsync(effect.courseId, effect.assignmentId, ApiPrefs.user!!.id, true).await()
             val assignment = AssignmentManager.getAssignmentAsync(effect.assignmentId, effect.courseId, true).await()
 
+
             // We need to know if they can make submissions through arc, only for file uploads - This is for empty submissions
             val isArcEnabled = if (assignment.isSuccess && assignment.dataOrThrow.getSubmissionTypes().contains(Assignment.SubmissionType.ONLINE_UPLOAD)) {
                 effect.courseId.isArcEnabled()
             } else false
 
-            consumer.accept(SubmissionDetailsEvent.DataLoaded(assignment, submission.await(), isArcEnabled))
```
Blackchai: Do we always need to get this ltiUrl? Just wondering if this is something we should always be attempting.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/ui/SubmissionDetailsView.kt
```diff
@@ -193,6 +194,7 @@ class SubmissionDetailsView(
             is SubmissionDetailsContentType.QuizContent -> QuizSubmissionViewFragment.newInstance(type.url)
             is SubmissionDetailsContentType.TextContent -> TextSubmissionViewFragment.newInstance(type.text)
             is SubmissionDetailsContentType.DiscussionContent -> DiscussionSubmissionViewFragment.newInstance(type.previewUrl ?: "")
+            is SubmissionDetailsContentType.ExternalToolContent -> LTIWebViewFragment.newInstance(LTIWebViewFragment.makeRoute(type.canvasContext, type.url, hideToolbar = true))!!
```
Blackchai: No way to get around the !! on the fragment instantiation?

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/SubmissionManager.kt
```diff
@@ -201,13 +201,16 @@ object SubmissionManager {
     }
 
     @JvmStatic
-    fun getLtiFromAuthenticationUrl(url: String, callback: StatusCallback<LTITool>, forceNetwork: Boolean) {
+    fun getLtiFromAuthenticationUrl(url: String?, callback: StatusCallback<LTITool>, forceNetwork: Boolean) {
```
Blackchai: What is the case for passing in a null value here?

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/ui/SubmissionDetailsView.kt
StephenBrough: `LTIWebViewFragment` has a case where `makeRoute` could return null if the route isn't valid. Rather than doing some bigger refactoring, or adding in additional methods, I just added the double bang since it shouldn't ever return null in _this_ case. Looks dumb, but it was quick, haha.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsEffectHandler.kt
StephenBrough: Should only be for assignments with external tool marked as the submission type. I'll make a change there.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/SubmissionManager.kt
StephenBrough: The url is nullable on the Assignment model, which I think it was being handled as a platform type previously, so a null value could have crashed the app. 

I suppose I could make a check earlier to see if the url is null and just not call into this method if it is, but if it _is_ null, then I believe an error is shown in the LTIWebViewFragment automatically. Sounds like it depends on how we want to handle the error when the url is null.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsEffectHandler.kt
Blackchai: thanks!

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/ui/SubmissionDetailsView.kt
Blackchai: :+1:

PR #95, opened 2019-06-12
[Student][MBL-12229] File upload persistence

CalvinKern:
Save file uploads to our database to persist for the new submission flow.
Also adds a progress body for file uploads, updates sent through eventbus ProgressEvent so that anything can receive events.

To test:
This is a pain to actually test. It's not hooked up to any screens yet, so you have to go manually change some files to be able to test it directly.

First - change UploadFilesDialog.kt at line 457 (bottom of `uploadFiles` function) to look like this:
```
            if(bundle != null) {
                dialogCallback?.invoke(EVENT_ON_UPLOAD_BEGIN)
                dialogAttachmentCallback?.invoke(EVENT_ON_UPLOAD_BEGIN, fileList.first())
//                intent.putExtras(bundle)
//                activity?.startService(intent)
                dismiss()
            }
```
Second - change AddSubmissionFragment.kt at line 281 (top of `setupListeners` function) to look like this:
```
        fileUpload.setOnClickListener {
            // TODO: Test new submission service here
            UploadFilesDialog.show(fragmentManager, UploadFilesDialog.createAssignmentBundle(null, canvasContext as Course, assignment)) { _, file ->
//                context?.let { SubmissionService.startFileSubmission(it, canvasContext, assignment.id, assignment.name, assignment.groupCategoryId, arrayListOf(file!!, file, file, file, file, file)) } // to test file submissions, can repeat `file` as many times as desired
//                context?.let { SubmissionService.retryFileSubmission(it, canvasContext, 29) } // to test retries, you need the submission id from the db (set a breakpoint in the submission service and use it)
            }
        }
```

In AddSubmissionFragment, you can uncomment whichever line you want to use for the current thing your testing. You can pull the db files from the device and view them in a db browser, or you can set a breakpoint at the bottom of the SubmissionFileUploadReceiver in SubmissionService.kt to make calls to view the database like:
`db.fileSubmissionQueries.getFilesForSubmissionId(submissionId).executeAsList()`

You can use that to verify that files were input correctly, deleted correctly, set to errors correctly, etc...

Some of the test cases that would be good to hit:
* Submitting one large file or a file over a slower network (to be able to see the progress update in the notification)
* Submitting multiple files
* Submitting multiple files that have an error (the only error I could get is File Quota Exceeded, I can help with getting that test case ready; errors for wrong file type should be caught before submitting the files to the service; the file upload service also stops uploading files as soon as one errors)
* Retrying a submission with some files that uploaded

### DIFF
--- a/apps/student/src/main/db/com/instructure/student/FileSubmission.sq
+++ b/apps/student/src/main/db/com/instructure/student/FileSubmission.sq
@@ -57,6 +57,11 @@ SELECT *
 FROM fileSubmission
 WHERE dbSubmissionId = ?;
 
+getFilesWithoutAttachmentsForSubmissionId:
+SELECT *
+FROM fileSubmission
+WHERE dbSubmissionId = ? AND attachmentId IS NULL;
+
 deleteFilesById:
 DELETE
 FROM fileSubmission
@@ -74,7 +79,7 @@ WHERE id = ?;
 
 setAttachmentId:
 UPDATE fileSubmission
-SET attachmentId = ?
+SET attachmentId = ?, errorFlag = ?, error = ?
 WHERE id = ?;
 
 getLastInsert:
--- a/apps/student/src/main/db/com/instructure/student/FileSubmission.sq
+++ b/apps/student/src/main/db/com/instructure/student/FileSubmission.sq
@@ -1,0 +1,81 @@
+ --
+ -- Copyright (C) 2019 - present Instructure, Inc.
+ --
+ --     Licensed under the Apache License, Version 2.0 (the "License");
+ --     you may not use this file except in compliance with the License.
+ --     You may obtain a copy of the License at
+ --
+ --     http://www.apache.org/licenses/LICENSE-2.0
+ --
+ --     Unless required by applicable law or agreed to in writing, software
+ --     distributed under the License is distributed on an "AS IS" BASIS,
+ --     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ --     See the License for the specific language governing permissions and
+ --     limitations under the License.
+ --
+
+import com.instructure.canvasapi2.models.CanvasContext;
+-- import com.instructure.student.db.sqlColAdapters.Date;
+import com.instructure.student.db.sqlColAdapters.ErrorColAdapter;
+import com.instructure.pandautils.models.FileSubmitObject;
+
+CREATE TABLE fileSubmission (
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    dbSubmissionId INTEGER NOT NULL,
+    attachmentId INTEGER,
+    name TEXT,
+    size INTEGER,
+    contentType TEXT,
+    fullPath TEXT,
+    error TEXT,
+    errorFlag INTEGER as Boolean NOT NULL DEFAULT 0
+);
+
+CREATE INDEX fileSubmission_dbSubmissionId ON fileSubmission(dbSubmissionId);
+
+--
+insertFile:
+INSERT INTO fileSubmission (dbSubmissionId, name, size, contentType, fullPath)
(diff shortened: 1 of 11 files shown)


apps/student/src/main/db/com/instructure/student/FileSubmission.sq
```diff
@@ -0,0 +1,86 @@
...
+ --
+ --     Unless required by applicable law or agreed to in writing, software
+ --     distributed under the License is distributed on an "AS IS" BASIS,
+ --     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ --     See the License for the specific language governing permissions and
+ --     limitations under the License.
+ --
+
+import com.instructure.canvasapi2.models.CanvasContext;
+-- import com.instructure.student.db.sqlColAdapters.Date;
```
StephenBrough: Looks like you don't need this import at all :D

apps/student/src/main/db/com/instructure/student/FileSubmission.sq
```diff
@@ -0,0 +1,86 @@
...
+SELECT *
+FROM fileSubmission
+WHERE dbSubmissionId = ?;
+
+getFilesWithoutAttachmentsForSubmissionId:
+SELECT *
+FROM fileSubmission
+WHERE dbSubmissionId = ? AND attachmentId IS NULL;
+
+deleteFilesById:
```
StephenBrough: Should probably be `deleteFileById`. Looks like there are a few other methods whose names don't line up with the query, here and in Submission.sq. We'll probably want to fix those either here or before the 6.6 release.

CalvinKern: Ah, nice catch. Not sure what else could be changed in FileSubmission.sq besides `getLastInsert` to `getLastInsertId` (as well as in Submission.sq) and maybe `setAttachmentId` to something like `setFileAttachmentIdAndError`.

I could remove the `insertFileWithError` as that's not getting used and I doubt it will.

What else did you feel could be changed with these sq files?

StephenBrough: I think that's probably fine there 👍 

Some of the queries imply retrieving a single entity, but we do a `SELECT *` on it. Most of it is constrained to some kind of ID (the non-primary key ones), but we didn't mark the ID as `UNIQUE`, and we aren't using a `LIMIT`, so it's possible we could retrieve multiple items. I don't have the experience to know how important those are though, if we are pretty sure the IDs will be unique.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/ProgressResponseBody.kt
```diff
@@ -0,0 +1,55 @@
...
+        var uploaded: Long = 0
+
+        FileInputStream(file).use { stream ->
+            var read: Int = stream.read(buffer)
+            while (read != -1) {
+
+                // Send out updates if we have a submission id
+                if (submissionId != null) {
+                    val event = ProgressEvent(submissionId, uploaded, contentLength())
+                    EventBus.getDefault().postSticky(event)
```
StephenBrough: _Classic._

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/ProgressResponseBody.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    override fun contentLength() = file.length()
+
+    override fun contentType() = MediaType.parse("application/octet-stream")
+
+    override fun writeTo(sink: BufferedSink) {
+        val buffer = ByteArray(DEFAULT_BUFFER_SIZE)
+        var uploaded: Long = 0
+
+        FileInputStream(file).use { stream ->
```
StephenBrough: I like this 👍

PR #97, opened 2019-06-14
[Student][Automation][MBL-12359]: LoginE2ETest

joehoag:
Also added @E2E annotation, and 3 flavors of flank.yml files.

I fixed a few accessibility issues that arose as well.


apps/student/src/main/res/layout/navigation_drawer.xml
```diff
@@ -241,7 +243,9 @@
                 <androidx.appcompat.widget.SwitchCompat
                     android:id="@+id/navigationDrawerColorOverlaySwitch"
                     android:layout_width="wrap_content"
-                    android:layout_height="wrap_content"/>
+                    android:layout_height="wrap_content"
+                    android:minHeight="48dp"
+                    android:minWidth="48dp"/>
 
             </LinearLayout>
```
joehoag: These were some accessibility fixes (touch target too small) that I had to make in the nav drawer.  I couldn't tell the difference before and after, but see the ticket for a before/after comparison.

apps/student/src/main/res/layout/viewholder_course_header.xml
```diff
@@ -35,6 +35,8 @@
         style="@style/TextFont.Medium"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
+        android:minHeight="48dp"
+        android:minWidth="48dp"
         android:layout_gravity="bottom|end"
         android:background="?android:selectableItemBackground"
         android:paddingBottom="2dp"
```
joehoag: Another accessibility fix (touch target too small again) in the courses header.

apps/student/flank_nightly.yml
```diff
@@ -0,0 +1,45 @@
+gcloud:
```
joehoag: I'm open to a better name than "nightly".  Maybe "multi", "devices", "multiple_devices", "device_array", ... ?

CalvinKern: I think "nightly" is okay, but feels a little weird as nothing in this file explicitly deals with having it run "nightly". "devices" might be the wrong keyword to think of too, "api" might be more appropriate as that's the property changing. Potentially something like "flank_api_sweep"
I'll keep thinking too as I review the rest of this.

CalvinKern: Would we want to test against landscape at all for these as well? If so, then maybe "devices" would be more reasonable.

apps/student/src/main/res/layout/navigation_drawer.xml
CalvinKern: These accessibility fixes look good to me too. 👍

apps/student/flank_nightly.yml
CalvinKern: "flank_device_suite"?

joehoag: "flank_sweep"?

joehoag:
I committed a change to rename "flank_nightly" to "flank_sweep".

PR #109, opened 2019-06-20
[Student][MBL-12554] Stub out some interaction tests

CalvinKern:


apps/student/src/androidTest/java/com/instructure/student/ui/interaction/DiscussionsInteractionTest.kt
```diff
@@ -0,0 +1,117 @@
...
+import org.junit.Test
+
+class DiscussionsInteractionTest : StudentTest() {
+    override fun displaysPageObjects() {
+        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
+    }
+
+    @Stub
+    @Test
+    fun testDiscussionCreate() {
```
joehoag: What is the meaning of testXXX without any suffix?  (E.g., you have testDiscussionCreate() and then testDiscussionCreate_withAttachment and _disabledWhenNotPermitted.)  Does a lack of suffix just imply "the base case"?

CalvinKern: That's what I was figuring. We have the base test for creating a discussion. Then we have the specific cases of _withAttachment, etc...

I wasn't sure if that was better for readability or if it would be better to make the specific cases be testDiscussion_create, testDiscussion_createWithAttachment, etc...

joehoag: I'm sure that no matter what you do, some people will like it and some won't.  :-)  I like what you have here.  You could consider using something like "_base" or "_normal" instead of no suffix at all, but people will probably get the idea either way.

CalvinKern: Ooh, I actually really like the idea of adding "_base". Makes it seem more consistent. I'll add that to the PR.

PR #110, opened 2019-06-24
[Student][MBL-12522] Annotation Meta Data

Blackchai:


apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/AssignmentDetailsRenderTest.kt
```diff
@@ -49,6 +53,7 @@ class AssignmentDetailsRenderTest : StudentRenderTest() {
     }
 
     @Test
+    @TestMetaData(Priority.P2, FeatureCategory.ASSIGNMENTS, TestCategory.RENDER)
```
joehoag: Are these really P2?  I thought we were part of a P0/P1 push.

joehoag: Sorry -- my mistake.  These were pre-existing P2 tests.

libs/panda_processor/src/main/java/com/instructure/panda_processor/FileGenerator.kt
```diff
@@ -0,0 +1,127 @@
...
+            annotationList.filter { it.priority == Priority.P3 && it.stubbed }.size,
+
+            annotationList.filter { it.priority == Priority.P0 && !it.stubbed }.size,
+            annotationList.filter { it.priority == Priority.P1 && !it.stubbed }.size,
+            annotationList.filter { it.priority == Priority.P2 && !it.stubbed }.size,
+            annotationList.filter { it.priority == Priority.P3 && !it.stubbed }.size
+    )
+
+    private fun calculateFeatureCounts(annotationList: List<TestMetaData>) = FeatureCounts(
+            annotationList.filter { FeatureCategory.ASSIGNMENTS in listOf(it.featureCategory, it.secondaryFeature) && it.stubbed }.size,
```
joehoag: Seems like there must be a way to iterate through FeatureCategory elements instead of enumerating them all like this.

libs/panda_processor/src/main/java/com/instructure/panda_processor/CustomFileBuilder.kt
```diff
@@ -0,0 +1,126 @@
...
+    val p3Stubbed: Int,
+
+    val p0Written: Int,
+    val p1Written: Int,
+    val p2Written: Int,
+    val p3Written: Int
+)
+
+data class FeatureCounts(
+    val assignmentsStubbed: Int,
```
joehoag: Similar to my comment below for calculateFeatureCounts, I just think that there must be a better way to do this other than expecting all of these explicit parameters in order.  Like, maybe we could track a map of (metric name) -> (metric count).  This would probably cause some churn to your code, but it might make it more maintainable in the long run.

Blackchai: Thanks! Check out the changes.

libs/panda_processor/src/main/java/com/instructure/panda_processor/CustomFileBuilder.kt
```diff
@@ -0,0 +1,73 @@
...
+
+class CustomFileBuilder(
+        totalWritten: Int,
+        totalStubbed: Int,
+        testCounts: HashMap<TestCategory, TestCount>,
+        priorityCounts: HashMap<Priority, TestCount>,
+        featureCounts: HashMap<FeatureCategory, TestCount>) {
+
+    private val contentTemplate = """
+        *** Total Test Counts ***
```
Blackchai: I'm going to leave this txt format as is for now, since its just for the POC. The final version will be a nicely formatted and far more visual file.

libs/panda_processor/src/main/java/com/instructure/panda_processor/CustomFileBuilder.kt
```diff
@@ -0,0 +1,73 @@
...
+
+
+        *** Test Priority ***
+        P0 Test Count = stubbed(${priorityCounts[Priority.P0]?.stubbed}) / written(${priorityCounts[Priority.P0]?.written})
+        P1 Test Count = stubbed(${priorityCounts[Priority.P1]?.stubbed}) / written(${priorityCounts[Priority.P1]?.written})
+        P2 Test Count = stubbed(${priorityCounts[Priority.P2]?.stubbed}) / written(${priorityCounts[Priority.P2]?.written})
+        P3 Test Count = stubbed(${priorityCounts[Priority.P3]?.stubbed}) / written(${priorityCounts[Priority.P3]?.written})
+
+        *** Test Feature Coverage ***
+        Assignments = stubbed(${featureCounts[FeatureCategory.ASSIGNMENTS]?.stubbed}) / written(${featureCounts[FeatureCategory.ASSIGNMENTS]?.written})
```
joehoag: It's probably possible to abstract these sections to a method that iterates through FeatureCategory / Priority / TestCategory and produces a the appropriate section in string form, so that you didn't have to manually enumerate like this.  But maybe that is a change for a future version.

joehoag:
The updated code is looking much better.  Thanks for addressing the issues raised.

There are opportunities for additional abstraction and simplification.  I'll leave it up to you to decide whether to pursue those now or in a future iteration.

FWIW, I know that we've collectively gone back and forth a lot on functionality and code structure, but overall I think that the end result here will be pretty awesome.  It will be really nice to generate this report any time that we want to (or perhaps after each Bitrise test run).

libs/panda_processor/src/main/java/com/instructure/panda_processor/CustomFileBuilder.kt
Blackchai: Yeah I totally agree, see my comment above though. The end game is to have a nicer visual file be generated, probably something in html or something suited to a dashboard. This text file was created just to give us something that we can look at for the results.

Blackchai:
Thanks Joe! I appreciate all the feedback!

PR #111, opened 2019-06-24
[Student][MBL-12506] Rubric updates

JordanMarshall:


apps/student/src/main/res/layout/adapter_rubric_criterion.xml
```diff
@@ -14,85 +14,118 @@
...
-            android:paddingEnd="12dp"
-            android:paddingBottom="8dp"
-            android:textColor="@color/black"
-            tools:text="This is a sample comment"/>
+            android:layout_marginTop="-14dp"
+            android:layout_marginBottom="-8dp"
+            android:background="?attr/selectableItemBackgroundBorderless"
+            android:minWidth="48dp"
+            android:minHeight="48dp"
+            android:paddingTop="14dp"
```
CalvinKern: Seems a little strange to have the padding and margin for the top cancel each other out. Are they really needed? (margin top is -14dp and padding top is 14dp)

JordanMarshall: Yes, it's rather unorthodox but it allows the button to meet the minimum touch area for a11y while also being placed correctly per the UI designs.

PR #117, opened 2019-06-26
[Student][Accessibility][MBL-11103]: Accessibility / Talkback fix in assignment list

joehoag:


apps/student/src/main/java/com/instructure/student/binders/AssignmentBinder.kt
```diff
@@ -54,6 +53,9 @@ class AssignmentBinder : BaseBinder() {
             } else {
                 holder.points.visibility = View.VISIBLE
                 BaseBinder.setupGradeText(context, holder.points, assignment, submission, courseColor)
+                // Set accessibility text
+                val outOf = context.resources.getString(R.string.outOf)
+                holder.points.contentDescription = holder.points.text.toString().replace("/", " $outOf ")
```
CalvinKern: Is this going to hold up for all languages? As in, are there any languages that wouldn't display a score as "x/x"?
If there is a possibility for that, we should probably construct the whole string with the points.

JordanMarshall: iirc we fixed a nearly identical issue for a different part of the app a while ago and we should have the relevant string for it somewhere, something like `%1$s out of %2$s points`. Part of the problem is that the string we get from BaseBinder is probably not correctly internationalized which complicates creating a proper fix here.

joehoag: The underlying logic (BaseBinder.getGrade()) seems to always use a "/", regardless of language.

CalvinKern: Hmm, yeah. That method looks to use a hard coded "/". So it does seem like this fix will work for now, though I feel like setting this content description might fit better inside the `setupGradeText()` method where the text is getting set.

That way if we ever revisit support for i18n here it's hopefully more clear that this string will need to support i18n through arguments as well.
Thoughts @JordanMarshall @Blackchai ?

joehoag: If we did it in setupGradeText, we would still need to depend on the "/" for our re-formatting for contentDescription.  The only way to do this "right" is to refactor getGrade() so that it returns left and right hand sides, and then have setupGradeText insert the "/".

joehoag:
Took Calvin's advice and moved the contentDescription logic into BaseBinder.setupGradeText().

PR #121, opened 2019-06-27
[Automation][MBL-12703]: Fix dittoMode propagation to espresso lib

joehoag:
The espresso lib changes can be viewed at https://github.com/instructure/mobile-shared/commit/185781e70ba0dbde009989c66317bd55f947c5fe.

The change in ApiPrefs.kt is potentially controversial.

The reason that the Parent UI tests kept failing: Our test environments return an empty string for terms of service, causing that button not to show up in the legal page, which causes our tests to fail.  We had specified "-PdittoMode=live" in our Bitrise config for the Parent UI tests, which meant that the termsOfUse-related tests were all going to fail.  Up until now, though, we had broken dittoMode propagation, which meant that we were NOT running the tests in live mode.  But as soon as I made this fix, we started running the tests in live mode, and thus started failing every time.

I've made the short term fix of removing "-PdittoMode=live" from our bitrise config for running parent UI tests.  We'll need to have another conversation about how to deal with the fact that we can never remake the parent ditto tapes again because the termsOfUse tests will fail ever after.

Follow-up: It turns out that we have control over what we return from the termsOfService call, so I'll make that tweak so that we are no longer returning nothing.


libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/ApiPrefs.kt
```diff
@@ -68,7 +70,7 @@ object ApiPrefs : PrefManager(PREFERENCE_FILE_NAME) {
             return selectedLocale.takeUnless { it == ACCOUNT_LOCALE }
                     ?: user?.effective_locale
                     ?: user?.locale
-                    ?: DEVICE_LOCALE
+                    ?: ConfigurationCompat.getLocales(Resources.getSystem().configuration)[0].language
         }
```
joehoag: DEVICE_LOCALE gets interpreted as "root", which was causing my tests to fail -- effective_locale would flip-flop back and forth between "en" and "root".  By taking the next step here and actually providing a real device locale (language), the confusion was addressed.

JordanMarshall: This _might_ affect the behavior of the dev language menu, but it looks like that menu isn't working correctly right now so it's not a blocking issue.

PR #127, opened 2019-06-27
[Student][Teacher][MBL-11335] Add floating recording view new student assignments 

Blackchai:
Couple of key notes:
-This also effects teacher. The floating recording view is now in pandautils and has been made _slightly_ more generic.
-For now, this doesn't actually submit the media comment. This PR was getting quite large already, and given how submission comments are handled there would have been a good deal of worked remaining to add that.


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsModels.kt
```diff
@@ -20,16 +20,29 @@ package com.instructure.student.mobius.assignmentDetails.submissionDetails
...
         SubmissionDetailsEvent()
 }
 
 sealed class SubmissionDetailsEffect {
+    object ShowAudioRecordingView : SubmissionDetailsEffect()
+    object ShowVideoRecordingView : SubmissionDetailsEffect()
+    object ShowVideoRecordingPlaybackError : SubmissionDetailsEffect()
+    object ShowMediaCommentError : SubmissionDetailsEffect()
+    object MediaCommentDialogClosed : SubmissionDetailsEffect()
+    data class ShowVideoRecordingPlayback(val file: File) : SubmissionDetailsEffect()
```
CalvinKern: Do we need an effect for ShowAudioRecordingPlayback too? Or should this effect be renamed to support both video and audio?

Blackchai: So the floating view for audio supports the playback itself, without any outside interaction. While the video view opens up an activity to show the video replay in fullscreen. If you think its confusing I could add some comments?

CalvinKern: Oh ok, that makes sense. I don't think it's too confusing, just wanted to make sure that it wasn't getting lost. But adding a comment would never hurt, just to make it clear that audio handles itself when we come back to this in a year.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/drawer/comments/ui/SubmissionCommentsFragment.kt
```diff
@@ -0,0 +1,51 @@
...
+        MobiusFragment<SubmissionCommentsModel, SubmissionCommentsEvent, SubmissionCommentsEffect, SubmissionCommentsView, SubmissionCommentsViewState>() {
+    lateinit var data: SubmissionDetailsTabData.CommentData
+
+    override fun makeEffectHandler() = SubmissionCommentsEffectHandler(requireContext())
+    override fun makeUpdate() = SubmissionCommentsUpdate()
+    override fun makeView(inflater: LayoutInflater, parent: ViewGroup) = SubmissionCommentsView(inflater, parent)
+    override fun makePresenter() = SubmissionCommentsPresenter
+
+    // TODO update Comment Data
+    override fun makeInitModel() = SubmissionCommentsModel(ArrayList<SubmissionComment>(), Submission(), 1L, 1L, data.assignmentId, false)
```
CalvinKern: IMO, if you're passing in bogus values for the submission id and the course id, they should be defaults inside of the model constructor. If these will be set here in a later PR, then just ignore this comment :)

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/drawer/comments/ui/SubmissionCommentsView.kt
```diff
@@ -0,0 +1,72 @@
...
+        dialog.setOnShowListener {
+            dialog.audioComment.setOnClickListener {
+                consumer?.accept(SubmissionCommentsEvent.AddAudioCommentClicked)
+                dialog.cancel()
+            }
+            dialog.videoComment.setOnClickListener {
+                consumer?.accept(SubmissionCommentsEvent.AddVideoCommentClicked)
+                dialog.cancel()
+            }
+        }
```
CalvinKern: I feel like you should be sending the MediaCommentDialogClosed event in an onDismissListener for the dialog. And then you wouldn't need the close as a shared event and could clean up that flow from the SubmissionDetailsLoop.

Otherwise, it seems like you'll have a bug once you try to incorporate `isMediaCommentEnabled` into a presenter when a user closes the dialog by clicking outside of it.

CalvinKern: Actually, you'll probably want to do it in onCancelListener.

apps/student/src/main/res/layout/dialog_media_comment_picker.xml
```diff
@@ -0,0 +1,56 @@
...
+        android:text="@string/addVideoComment"
+        android:textSize="16sp"
+        android:drawableStart="@drawable/vd_video"
+        android:paddingTop="14dp"
+        android:paddingBottom="14dp"
+        android:paddingStart="18dp"
+        android:paddingEnd="18dp"
+        android:drawablePadding="18dp"
+        android:background="?attr/selectableItemBackground"
+        tools:visibility="visible" />
```
CalvinKern: Looks like you're only using one of these buttons now, but this whole file looks like placeholder so ¯\_(ツ)_/¯

apps/student/src/test/java/com/instructure/student/test/assignment/details/submissionDetails/commentTab/SubmissionCommentsUpdateTest.kt
```diff
@@ -0,0 +1,136 @@
...
+        val expectedModel = initModel.copy(isMediaCommentEnabled = true)
+        updateSpec
+                .given(givenModel)
+                .whenEvent(SubmissionCommentsEvent.MediaCommentDialogClosed)
+                .then(
+                        assertThatNext(
+                                hasModel(expectedModel)
+                        )
+                )
+    }
```
CalvinKern: Should this test specify that there should be no effects sent as well?
assertThatNext(
  hasModel(expectedModel),
  NextMaterhcers.hasNoEffects()
)

apps/teacher/build.gradle
```diff
@@ -195,7 +195,6 @@ dependencies {
     implementation project(path: ':login-api-2')
     implementation project(path: ':chips')
     implementation project(path: ':rceditor')
-    implementation 'com.wonderkiln:camerakit:0.13.4'
```
CalvinKern: 🤣

libs/pandares/src/main/res/values-sw600dp/dimens.xml
```diff
@@ -0,0 +1,22 @@
...
+  ~     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~     See the License for the specific language governing permissions and
+  ~     limitations under the License.
+  ~
+  -->
+<resources>
+    <dimen name="recordingViewHeight">254dp</dimen>
+    <dimen name="recordingViewWidth">390dp</dimen>
+    <dimen name="mediaRecordingButton">64dp</dimen>
+</resources>
```
CalvinKern: The old file these were pulled from was values-sw760dp and this one is values-sw600dp . Do they need the same dp value?

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsModels.kt
Blackchai: :+1:

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/drawer/comments/ui/SubmissionCommentsFragment.kt
Blackchai: Yeah, atm none of this is being used since I'm not actually setting up or presenting this screen. I can update it to be something different if you want, but they were just there to satisfy this function.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/drawer/comments/ui/SubmissionCommentsView.kt
Blackchai: ahh good catch, yeah I need to add that event to this dialog. Thanks!

apps/student/src/test/java/com/instructure/student/test/assignment/details/submissionDetails/commentTab/SubmissionCommentsUpdateTest.kt
Blackchai: TBH I just haven't written mobius tests in a minute. I'm not expecting any effects, so I suppose I should add that matcher.

libs/pandares/src/main/res/values-sw600dp/dimens.xml
Blackchai: hmmmm I thought I matched the new directory. I would have sworn they were the same. I'll double check.

Blackchai: nice catch!

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/drawer/comments/ui/SubmissionCommentsFragment.kt
Blackchai: Ah, so yeah, since they will be set in a later PR I'll just leave 'em.

PR #141, opened 2019-07-02
[Automation][MBL-12692]: Moved from mobileqa.test to mobileqa.beta

joehoag:


apps/parent/src/androidTest/java/com/instructure/parentapp/ui/pages/LoginSignInPage.kt
```diff
@@ -97,7 +97,7 @@ class LoginSignInPage: BasePage() {
         enterEmail(user.loginId)
         enterPassword(user.password)
         clickLoginButton()
-        authorizeButton().perform(webClick())
+        // authorizeButton().perform(webClick()) // Not necessary on mobileqa.beta
```
joehoag: This was really puzzling -- you no longer had to press the "authorize android ..." button after signing in.  (There was no such button presented.)  If anyone finds this alarming, I'll be happy to dig into it.

JordanMarshall: 👍 I had totally forgotten that the test domain still used the authorize button, but this is an expected change.

PR #143, opened 2019-07-02
[MBL-12555] Add git workflow doc

JordanMarshall:


CalvinKern:
Would it be worth putting the docs directory inside of `apps` so we can edit in Android Studio or should we rely on external editors for the markdown and keep docs agnostic of `apps` so it could be used for any kind of document?

JordanMarshall:
Android Studio doesn't have a working markdown preview so it's not much more useful than other text editors in that regard. imo we should keep the docs dir at the root level but I could be persuaded otherwise.

CalvinKern:
Ah, yeah. I've used a third party markdown editor in Android Studio before. That's the main reason I was asking. I'm fine to use another editor for this and keep it out of our app namespace. But the option is there if we want it (Markdown Navigator by Vladmir Schneider).

PR #151, opened 2019-07-16
[Teacher][MBL-12823] Make modules with embedded Studio vids work again

StephenBrough:
See ticket for testing creds.

### DIFF
--- a/apps/teacher/src/main/java/com/instructure/teacher/features/modules/list/ui/ModuleListView.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/features/modules/list/ui/ModuleListView.kt
@@ -118,11 +118,12 @@ class ModuleListView(
                 Route(null, QuizDetailsFragment::class.java, canvasContext, args)
             }
             ModuleItem.Type.ExternalUrl, ModuleItem.Type.ExternalTool -> {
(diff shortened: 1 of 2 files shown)


apps/teacher/src/main/java/com/instructure/teacher/features/modules/list/ui/ModuleListView.kt
```diff
@@ -118,11 +118,12 @@ class ModuleListView(
                 Route(null, QuizDetailsFragment::class.java, canvasContext, args)
             }
             ModuleItem.Type.ExternalUrl, ModuleItem.Type.ExternalTool -> {
-                val args = InternalWebViewFragment.makeBundle(
-                    item.externalUrl.orEmpty(),
-                    item.title.orEmpty()
+                val args = LTIWebViewFragment.makeLTIBundle(
```
JordanMarshall: For external urls this displays an empty page with the message "An unexpected error occurred", so we probably want to split up this branch condition and retain the original behavior for those urls. Also, this probably falls outside the scope of the ticket, but it looks like this change causes two of the LTIs in the repro account to load with an incorrect initial viewport. Specifically, "Vimeo LTI Video" and "Youtube LTI Video" load in a zoomed-in state and the user must manually zoom out to see the entire video.

PR #160, opened 2019-07-18
[Student][MBL-12672]: Initial attempt at incorporating Firebase Remote Config

joehoag:
Initially, we're only doing this in the Student app.

We may want to do a few more things before charging into Remote Config usage:
- Decide on a naming scheme for remote config parameters.  Should we name specifically for app (e.g., "student_rubric_v2"), with the possibility of naming for all apps (e.g., "all_rubric_v2")?
- Add a "remote config settings" menu/activity in debug builds to allow us to locally override remote config parameter values.

### DIFF
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/RemoteConfigUtils.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/RemoteConfigUtils.kt
@@ -1,0 +1,96 @@
+package com.instructure.canvasapi2.utils
+
+import com.google.firebase.remoteconfig.FirebaseRemoteConfig
+import com.google.firebase.remoteconfig.FirebaseRemoteConfigSettings
+
+/**
+ * All access to remote config data from our code must go through these enum values.
+ * [rc_name] is the name of the parameter as specified in the remote config console.
+ * [safeValueAsString] is the "safe to use" value for the parameter, specified in string form.  This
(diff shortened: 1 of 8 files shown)


libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/RemoteConfigUtils.kt
```diff
@@ -0,0 +1,96 @@
...
+                val updated = task.getResult()
+                if(updated != null && updated) {
+                    remoteConfig.getKeysByPrefix("").forEach() {
+                        //println("rc: on completion: putting ($it, ${remoteConfig.getString(it)}) into prefs")
+                        prefs.putString(it, remoteConfig.getString(it))
+                        //println("rc: on completion: retrieved ${prefs.getString(it)}")
+                    }
+                }
+            }
+        }
```
joehoag: We could consider launching a repeating task that re-grabs the remote config values once an hour or something like that.  As it stands, this code will just grab them at app startup.  I don't know what our requirements are, or if there is a use case that would necessitate occasional re-grabbing.

CalvinKern:
I've confirmed that this doesn't impede our app in regards to China. As in, if this request fails (by being blocked) the app still functions properly.

joehoag:
> I've confirmed that this doesn't impede our app in regards to China. As in, if this request fails (by being blocked) the app still functions properly.

Yes, if our request fails, we will fall back to previous values, or the defined "safe defaults" if the request never succeeds.

PR #164, opened 2019-07-19
[Student][MBL-12668] Show grades when logged in as an observer

StephenBrough:
See ticket for some testing creds. Jordan's account also has one that's good for testing.

We only grab the first observee in a course and display their grades.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/adapter/GradesListRecyclerAdapter.kt
+++ b/apps/student/src/main/java/com/instructure/student/adapter/GradesListRecyclerAdapter.kt
@@ -174,20 +174,21 @@ open class GradesListRecyclerAdapter(
                         // Get the first student this user is observing
                         val student = awaitApi<List<Enrollment>> { EnrollmentManager.getObserveeEnrollments(true, it) }.filter { it.courseId == course.id }.filter { it.observedUser != null }[0].observedUser
                         // Get Assignment Groups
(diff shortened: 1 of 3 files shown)


apps/student/src/main/java/com/instructure/student/adapter/GradesListRecyclerAdapter.kt
```diff
@@ -103,8 +108,104 @@ open class GradesListRecyclerAdapter(
...
+                return@launch
+            }
+
+            for (enrollment in course.enrollments!!) {
+                if (enrollment.isStudent && enrollment.multipleGradingPeriodsEnabled) {
+                    // We load current term
+                    currentGradingPeriod = GradingPeriod(
+                        id = enrollment.currentGradingPeriodId,
+                        title = enrollment.currentGradingPeriodTitle
+                    )
```
CalvinKern: The setting of currentGradingPeriod used to happen inside the next if check (if currentGradingPeriod == null || title == null).

Looks like it all still works here, but now doing a PTR always resets the the current grading period, instead of keeping the grading period that was selected before the PTR.

If there wasn't really a reason to pull this out of the if, can we add it back inside the next check to keep that data?
Though, I don't know how that handles when the selected grading period gets deleted on PTR (can that even happen?!)

StephenBrough: Weird... I must have missed this when I was reorganizing it. CONSIDER IT DONE.

StephenBrough:
I definitely like the idea of making the API calls more efficient. I'm guessing it'll work the same in Student as it does in Parent, I'll look into using that call.

PR #172, opened 2019-07-25
[Student][MBL-12909] Update empty states for rubric and files tabs in submission details

StephenBrough:
See ticket for the specs Jordan posted.

### DIFF
--- a/libs/pandautils/src/main/res/values/strings.xml
+++ b/libs/pandautils/src/main/res/values/strings.xml
@@ -168,7 +168,8 @@
     <string name="noAssignments">No Assignments</string>
     <string name="noAssignmentsSubtext">It looks like assignments haven\'t been created in this space. Time to go out and explore.</string>
     <string name="noAssignmentsTeacher">It looks like no assignments have been published yet.</string>
-    <string name="noFiles">There are no files for this assignment</string>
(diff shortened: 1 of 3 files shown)


libs/pandautils/src/main/res/values/strings.xml
```diff
@@ -168,7 +168,7 @@
     <string name="noAssignments">No Assignments</string>
     <string name="noAssignmentsSubtext">It looks like assignments haven\'t been created in this space. Time to go out and explore.</string>
     <string name="noAssignmentsTeacher">It looks like no assignments have been published yet.</string>
-    <string name="noFiles">No Files</string>
+    <string name="noFiles">There are no files for this assignment</string>
```
CalvinKern: Seems like this string is used in other places that are not in an assignment. Probably need a new resource for the submission files string.

StephenBrough: Good catch, changing now!

Blackchai: This is also in pandautils! Move it to pandares please if you are going to change it :)

Blackchai: Nevermind :3

PR #183, opened 2019-07-30
[Student][MBL-12921]: Fixed problem with assignment details scrolling at startup

joehoag:
So, the problem only occurs when there is a description for the assignment, which means that the description webview needs to be rendered.  It seems that, when the webview is rendered, it demands focus and therefore causes a scroll event to occur.

This is a pretty simple fix, but let me know if you see any potential problems with it.


joehoag:
Thanks for the detailed (offline) review and discussion!

PR #184, opened 2019-07-30
[Student][MBL-12935][MBL-12937] Handle submission state in assignment

CalvinKern:
This shows “graded” if the assignment was graded, “submitted” if it was submitted”, “missing” if there is no student submission and it was marked as missing or is past due, and “not submitted” otherwise.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsPresenter.kt
+++ b/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsPresenter.kt
@@ -26,6 +26,7 @@ import com.instructure.canvasapi2.utils.NumberHelper
 import com.instructure.canvasapi2.utils.isRtl
 import com.instructure.canvasapi2.utils.isValid
 import com.instructure.pandautils.discussions.DiscussionUtils
+import com.instructure.pandautils.utils.AssignmentUtils2
 import com.instructure.student.R
 import com.instructure.student.Submission
 import com.instructure.student.mobius.assignmentDetails.ui.AssignmentDetailsViewState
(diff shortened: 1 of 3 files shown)


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsPresenter.kt
```diff
@@ -84,12 +85,23 @@ object AssignmentDetailsPresenter : Presenter<AssignmentDetailsModel, Assignment
...
+        val (submittedLabelRes, submittedColorRes, submittedIconRes) = if (assignment.isSubmitted) {
+            Triple(
+                if (assignmentState == AssignmentUtils2.ASSIGNMENT_STATE_GRADED) R.string.gradedSubmissionLabel else R.string.submitted,
+                R.color.alertGreen,
+                R.drawable.vd_submitted
+            )
         } else {
-            Triple(R.string.notSubmitted, R.color.defaultTextGray, R.drawable.vd_unsubmitted)
+            if (assignment.submission?.missing == true ||
+                (assignment.dueAt != null && assignmentState == AssignmentUtils2.ASSIGNMENT_STATE_MISSING)) {
```
joehoag: Is assignment.dueAt != null the same as "past due"?  Wouldn't you need to compare the current time with the dueAt value?

CalvinKern: Getting the assignment state does that check to see if the date is passed, but it will also return missing if there is no due at. Since we only want to show missing if it's missing and due, we have to double check that here. That way we can show "not submitted" instead of missing for no due date assignments.

joehoag: 👍 Got it.

PR #187, opened 2019-07-30
[Student][MBL-12920] Add choose file to submit media dialog

Blackchai:


apps/student/src/main/res/layout/dialog_submission_picker_media.xml
```diff
@@ -49,4 +49,19 @@
...
         android:background="?attr/selectableItemBackground" />
 
+    <TextView
+        android:id="@+id/submissionEntryMediaFile"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        style="@style/TextFont.Regular"
+        android:text="@string/chooseMediaFile"
+        android:textSize="16sp"
+        android:drawableStart="@drawable/vd_utils_media"
```
CalvinKern: Can we use the new icon that's properly sized :)

CalvinKern:
Looks like we're passing the allowable file types in when doing media recordings. Probably shouldn't as the online file upload is the only one that needs to abide by the allowable file types.

apps/student/src/main/res/layout/dialog_submission_picker_media.xml
Blackchai: roger that!

PR #188, opened 2019-07-31
[Student][MBL-12913] Added navigation on failed submission comment notifications

CalvinKern:


apps/student/src/main/java/com/instructure/student/mobius/common/ui/SubmissionService.kt
```diff
@@ -344,27 +346,35 @@ class SubmissionService : IntentService(SubmissionService::class.java.simpleName
             // Set up notifications
             FileUploadService.createNotificationChannel(notificationManager, COMMENT_CHANNEL_ID)
             val notification = NotificationCompat.Builder(this@SubmissionService, COMMENT_CHANNEL_ID)
-                .setOngoing(true)
```
Blackchai: Do we want to remove this?

PR #190, opened 2019-07-31
[Student][MBL-12918] File type icons

Blackchai:
Added support for the various file types in the panda picker and the upload status view.


CalvinKern:
There seems to be a minor issue with the way that audio files. They are using the video/media icon as the content type for `.3gp` is resolving to `video/3gpp`. We either need to not store the content type for audio files and add `.3gp` as one of the audio extensions in FileUtils, ignore this issue and let audio files show the video/media icon, or change the file type that the audio recorder creates so that it will resolve to an audio content type.

PR #194, opened 2019-08-01
[Student][MBL-12912] Made web view in assignment more scrollable

CalvinKern:
This makes it so that the webview in the assignment details can be scrolled down, but when scrolling back up it triggers the PTR as the webview scrolls. Better than it was before, but still not great. We'll need to spend more time later figuring out a way to get PTR, nested scroll view, and webview to work better together.


CalvinKern:
We are shelving this for now as more effort is required to get this working correctly. Attempted solutions in this PR made LTI's work better at the expense of normal assignment descriptions.

PR #199, opened 2019-08-02
[Student][Comments][MBL-12915]: Added a render UI test for comments

joehoag:
Also MBL-12969: Fixed some ProgressBar-related PickerSubmissionUploadRenderTest tests to not run on API 23.


apps/student/src/androidTest/java/com/instructure/student/ui/interaction/CommentsInteractionTest.kt
```diff
@@ -0,0 +1,36 @@
...
+
+class CommentsInteractionTest: StudentTest() {
+    override fun displaysPageObjects() = Unit // Not used for interaction tests
+
+    @Stub
+    @Test
+    @TestMetaData(Priority.P0, FeatureCategory.COMMENTS, TestCategory.INTERACTION, true, FeatureCategory.ASSIGNMENTS)
+    fun testComments_addComment() {
+        // Test adding a comment to an assignment comment stream
+    }
```
joehoag: Just stubbing this for now.

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
```diff
@@ -0,0 +1,225 @@
...
+                        get() = "Pending message"
+                    override val progress: Double?
+                        get() = 0.45
+
+                }
+        )
+    }
+
+    @Test
+    @TestMetaData(Priority.P1,FeatureCategory.COMMENTS,TestCategory.RENDER,secondaryFeature = FeatureCategory.ASSIGNMENTS)
```
joehoag: I don't really know what our rules are for setting test priority values.  I'm open to suggestions/alternatives.

libs/panda_annotations/src/main/java/com/instructure/panda_annotations/TestMetaData.kt
```diff
@@ -34,7 +34,8 @@ enum class Priority {
 
 enum class FeatureCategory {
     ASSIGNMENTS, SUBMISSIONS, LOGIN, COURSE, DASHBOARD, GROUPS, SETTINGS, PAGES, DISCUSSIONS, MODULES,
-    INBOX, GRADES, FILES, EVENTS, PEOPLE, CONFERENCES, COLLABORATIONS, SYLLABUS, TODOS, QUIZZES, NOTIFICATIONS, ANNOTATIONS, ANNOUNCEMENTS, NONE
+    INBOX, GRADES, FILES, EVENTS, PEOPLE, CONFERENCES, COLLABORATIONS, SYLLABUS, TODOS, QUIZZES, NOTIFICATIONS,
+    ANNOTATIONS, ANNOUNCEMENTS, COMMENTS, NONE
```
joehoag: I just added a COMMENTS feature category.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/CommentsInteractionTest.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.instructure.panda_annotations.TestMetaData
+import com.instructure.student.ui.utils.StudentTest
+import org.junit.Test
+
+class CommentsInteractionTest: StudentTest() {
+    override fun displaysPageObjects() = Unit // Not used for interaction tests
+
+    @Stub
+    @Test
+    @TestMetaData(Priority.P0, FeatureCategory.COMMENTS, TestCategory.INTERACTION, true, FeatureCategory.ASSIGNMENTS)
```
CalvinKern: The other category should probably be SUBMISSIONS rather than ASSIGNMENTS. It also seems like we do the primary/secondary in hierarchy order for other tests, so maybe switch them as well so FeatureCategory.SUBMISSIONS is the feature category, and FeatureCategory.COMMENTS is the secondary category?

joehoag: 👍 I'll do that.

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
```diff
@@ -0,0 +1,225 @@
...
+                tint = 1000
+        )
+
+        // Just another example of how painful it is to test pending comments.
+        // This is dependent on a generated PendingSubmissionComment interface.
+        pendingCommentItem = CommentItemState.PendingCommentItem(
+                authorName = user.shortName!!,
+                avatarUrl = user.avatarUrl!!,
+                sortDate = Date(119, 6, 4, 13, 20),
+                pendingComment = object : PendingSubmissionComment {
```
CalvinKern: Instead of implementing an object here, you could instead create a `PendingSubmissionComment.Impl(...)`. This makes it so you don't have to implement everything through the getter, but just as constructor params instead.

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
CalvinKern: FeatureCategory.ASSIGNMENTS to FeatureCategory.SUBMISSIONS and swap order

CalvinKern: I'm never sure what to put for our priorities. Though we'll probably want these down to a p2? I think we want to cover p0 and p1 through end to end tests, but I can't remember. @Blackchai input?

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
joehoag: Good idea.  This object isn't actually being used right now, but this change will improve our lot once we actually tackle the rendering of pending messages.

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
joehoag: 👍

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
```diff
@@ -0,0 +1,225 @@
...
+
+    private fun loadPageWithViewState(state: SubmissionCommentsViewState) {
+        val emptyEffectRunner = object : WorkRunner {
+            override fun dispose() = Unit
+            override fun post(runnable: Runnable) = Unit
+        }
+        val fragment = SubmissionCommentsFragment().apply {
+            overrideInitViewState = state
+            loopMod = { it.effectRunner { emptyEffectRunner } }
+            data = SubmissionDetailsTabData.CommentData( // ?? I don't know what this does, but I need to provide it
```
CalvinKern: Looks like this is used to populate the initial mobius model. Since you're providing an overrideInitViewState, that model isn't getting used to create a view state, though it is still getting created thus requiring the data to be set.

apps/student/src/androidTest/java/com/instructure/student/ui/pages/renderPages/SubmissionCommentsRenderPage.kt
```diff
@@ -0,0 +1,110 @@
...
+            is CommentItemState.PendingCommentItem -> commentState.pendingComment.message
+            else -> throw IllegalStateException("Unknown comment item state type ${commentState::class.java.simpleName}")
+        }
+    }
+
+    private fun getDateString(commentState: CommentItemState) : String? {
+        return when(commentState) {
+            is CommentItemState.CommentItem -> commentState.dateText
+            is CommentItemState.SubmissionItem -> commentState.dateText
+            is CommentItemState.PendingCommentItem -> commentState.sortDate.toLocaleString() // (Maybe??)
```
CalvinKern: In the `PendingCommentBinder.kt` it looks like the date isn't used at all for pending comments. If the error flag is set it should be "Error" otherwise it should be "Sending...".
Since we're not testing anything with pending comments in the render test, maybe unimportant?

joehoag: Yeah, I figure that we can tackle this setting when we finally try to display pending comments.  I'll add a comment.

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
Blackchai: Yeah I'm pretty torn on priority atm. Imo, the only priorities that _really_ matter atm are the ones that map to the p0/p1 test cases that we imported from test rails. Thus far, we've been mostly doing p2/p3 for render tests, and I think that makes sense. I do think we need to have a discussion at some point regarding priorities and what we can gain from tracking them on new test cases.

joehoag:
Committed an update with the requested changes.

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
joehoag: OK, I'll change the p1s to p2s.

joehoag:
Thanks for the thoughtful review!

PR #200, opened 2019-08-02
[Student][MBL-12529] Add db schema to project

CalvinKern:
Also made some columns non null, and added some progress columns to a submission for future work.

Testing this requires a clean install (or delete all app data on existing app) since the database schema changed for the non null and added columns.


apps/student/src/test/java/com/instructure/student/test/assignment/details/AssignmentDetailsEffectHandlerTest.kt
```diff
@@ -115,19 +115,22 @@ class AssignmentDetailsEffectHandlerTest : Assert() {
...
-            "Entry text",
-            OffsetDateTime.now().minusDays(daysAgo),
-            "Assignment Name",
-            assignment.id,
-            course,
-            submissionType,
-            failed,
-            null,
-            userId
+            id = submissionId,
```
StephenBrough: I like this. With these larger parameter lists with db calls, it's helpful to use named parameters 👍

apps/student/src/main/db/com/instructure/student/Submission.sq
```diff
@@ -76,5 +78,12 @@ UPDATE submission
 SET errorFlag = ?
 WHERE id = ?;
 
+updateProgress:
```
StephenBrough: Do we need the file related fields for all submissions?

CalvinKern: We don't need it for text/url submissions, but do for media/file submissions. But I don't think we'll even be updating progress unless there's files (text/url don't report progress, they just upload and finish then we delete the entry from the database)

PR #202, opened 2019-08-02
[Student][MBL-12949] Workaround for minus letter grades

Blackchai:


Blackchai:
> This works great, but do we want to apply this fix to the assignments list and the grades page as well?

Yeah probs, I'll add that right quick.

joehoag:
👍 The latest changes properly read A, A- and B+ in the grades/assignment details pages.

apps/student/src/main/java/com/instructure/student/binders/BaseBinder.kt
```diff
@@ -127,8 +128,12 @@ open class BaseBinder {
                 textView.setBackgroundDrawable(createGradeIndicatorBackground(context, color))
                 // Set accessibility text
                 val outOf = context.resources.getString(R.string.outOf)
-                textView.contentDescription = grade?.replace("/", " $outOf ")
-
+                val a11yGradeString = getAccessibleGradeString(grade ?: "", context)
+                if (a11yGradeString.isNotEmpty()) {
+                    textView.contentDescription = a11yGradeString
+                } else {
+                    textView.contentDescription = grade?.replace("/", " $outOf ")
```
joehoag: Don't we only get here when grade is empty or null?

libs/pandautils/src/main/java/com/instructure/pandautils/utils/A11yUtils.kt
```diff
@@ -40,4 +41,12 @@ val AccessibilityManager.hasVisualFeedback get() = isServiceEnabled(Accessibilit
 val AccessibilityManager.isSwitchAccessEnabled: Boolean
     get() = getEnabledAccessibilityServiceList(AccessibilityServiceInfo.FEEDBACK_ALL_MASK).any { "com.android.switchaccess.SwitchAccessService" in it.id }
 
+fun getAccessibleGradeString(grade: String, context: Context): String {
```
joehoag: This seems like a misleading name for this function.  Maybe "getMinusAdjustedString" or something like that?  Or else return the original string instead of "" when it doesn't contain "-".

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/ui/gradeCell/GradeCellViewState.kt
```diff
@@ -118,14 +129,16 @@ sealed class GradeCellViewState {
...
+            return GradeData(
                 graphPercent = graphPercent,
                 accentColor = accentColor,
                 score = score,
                 showPointsLabel = true,
                 outOf = outOfText,
                 outOfContentDescription = outOfContentDescriptionText,
                 grade = grade,
+                gradeContentDescription = accessibleGradeString,
+                gradeCellContentDescription = gradeCellContentDescription,
```
joehoag: It wouldn't hurt to add a comment for each of these describing what it represented.  Although you do have some comments above by their respective initializations, so maybe that will suffice.

joehoag:
Screen Shot 2019-08-05 at 11 49 20 AM

Nit: The "B-" in the total grade at the top of the grades page reads as "B".  Do you think that's important enough to address?

joehoag:
Also, it looks like quite a few GradeCellTest unit tests failed.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/ui/gradeCell/GradeCellViewState.kt
Blackchai: I also feel like they are named pretty well, so I think it should be sufficient, if you are okay with that.

joehoag: 👍

joehoag:
Looks like the Grades total grade is now being read correctly.  

But I did see a couple more things that we might want to address.  (I'm sorry, man, I should do a better job of presenting all of these at once.) 

Screen Shot 2019-08-05 at 1 16 30 PM

"15 pts" gets read as "15 P T S"

Screen Shot 2019-08-05 at 1 15 48 PM

"13/15" gets read as "thirteen fifteenths".  I *think* that that used to work correctly.

joehoag:
(Note: I will create a 6.6.1 ticket for the "P T S" problem.  Trevor will attempt a fix for the "thirteen fifteenths" problem.)

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/ui/gradeCell/GradeCellViewState.kt
Blackchai: Thanks!

joehoag:
Epilog: I'm actually a little concerned over the amount of special-casing that we had to employ to get all of the edge-cases right.  It has a brittle vibe about it.  To be sure, no slicker solutions occur to me, and this was probably the right thing to do to get this release out the door.  

But maybe we want to create a tech-debt ticket to revisit this?

PR #210, opened 2019-08-05
[Student][MBL-12985] Fix url encoded text submissions in database

CalvinKern:
Instead, only encode when we're actually submitting the text to the API.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/mobius/common/ui/SubmissionService.kt
+++ b/apps/student/src/main/java/com/instructure/student/mobius/common/ui/SubmissionService.kt
@@ -60,6 +60,8 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import org.threeten.bp.OffsetDateTime
 import java.io.File
+import java.io.UnsupportedEncodingException
+import java.net.URLEncoder
 
 
 class SubmissionFileUploadReceiver(private val dbSubmissionId: Long) : BroadcastReceiver() {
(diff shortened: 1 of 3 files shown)


apps/student/src/main/java/com/instructure/student/mobius/common/ui/SubmissionService.kt
```diff
@@ -148,7 +150,11 @@ class SubmissionService : IntentService(SubmissionService::class.java.simpleName
         dbSubmissionId = db.getLastInsert().executeAsOne()
 
         showProgressNotification(assignmentName, dbSubmissionId)
-        val result = apiAsync<Submission> { SubmissionManager.postTextSubmission(context, assignmentId, text, it) }
+        val textToSubmit = try {
```
StephenBrough: Should we have a test for this?

CalvinKern: I was thinking a test for this would be nice, but we aren't currently testing anything in the submission service except for launching. By nature, classes that inherit from IntentService are not testable: https://developer.android.com/training/testing/integration-testing/service-testing.html
>>> Note: The ServiceTestRule class does not support testing of IntentService objects. If you need to test a IntentService object, you should encapsulate the logic in a separate class and create a corresponding unit test instead.

May be worth trying to move a lot of this logic to helper classes that are testable, but that will have to be later tech-debt to be able to release today. Unless you strongly feel like it should go in now, in which case I'll jump on this.

StephenBrough: Nah, I think we can put it off until later. I think we're still figuring out how the service stuff will be organized, so we'll tackle all the testing stuffs then.

PR #211, opened 2019-08-06
[Student] Fix quiz question timer crash

JordanMarshall:
See https://fabric.io/instructure/android/apps/com.instructure.candroid/issues/5c991dc8f8b88c29639899d1

To repro: Open the quiz details page for a timed quiz and double-tap "start" or "resume quiz" fast enough to open two instances of the quiz questions fragment. After the second fragment has opened and the timer is counting down, back out to the quiz details page and wait for the app to crash.


StephenBrough:
Looks like we still open two fragments, but the crash is OUTTA THERE.

PR #217, opened 2019-08-07
[student][MBL-13019][RC-6.6] Fix assignment deets discussion crash

Blackchai:
This crash appears to be caused by blueprint courses, which don't copy over the author. See fabric for domains and user ids to masquerade with to test.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsPresenter.kt
+++ b/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsPresenter.kt
@@ -19,6 +19,7 @@ package com.instructure.student.mobius.assignmentDetails
 import android.content.Context
 import androidx.core.content.ContextCompat
 import com.instructure.canvasapi2.models.Assignment
+import com.instructure.canvasapi2.models.DiscussionParticipant
 import com.instructure.canvasapi2.models.DiscussionTopicHeader
 import com.instructure.canvasapi2.models.Quiz
 import com.instructure.canvasapi2.utils.DateHelper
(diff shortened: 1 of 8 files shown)


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsPresenter.kt
```diff
@@ -287,13 +288,21 @@ object AssignmentDetailsPresenter : Presenter<AssignmentDetailsModel, Assignment
...
+            val authoredDate = DateHelper.getMonthDayAtTime(context, discussionTopicHeader.postedDate, context.getString(R.string.at)) ?: context.getString(R.string.discussions_unknown_date)
+            val attachmentIconVisibility = discussionTopicHeader.attachments.isNotEmpty()
+
+            DiscussionHeaderViewState.Loaded(authorAvatarUrl, authorName, authoredDate, attachmentIconVisibility)
+        }
+    }
 
-        return DiscussionHeaderViewState(authorAvatarUrl, authorName, authoredDate, attachmentIconVisibility)
+    private fun isDiscussionAuthorNull(author: DiscussionParticipant?): Boolean {
+        return author?.id == 0L && author.displayName == null && author.avatarImageUrl == null && author.htmlUrl == null
```
joehoag: Don't all of the author dereferences need to be "author?." -- with a question mark?  Does the initial author?.id check guarantee that author is non-null in all subsequent checks?  (I guess, if it's compiling successfully, that must be the case.)

Blackchai: Yeah that confused me a bit too, I think that it would be necessary if those were ORs instead of ANDs. It actually gives a warning if you leave the safe calls in haha.

CalvinKern: So, this does work with the way that the API returns us the data, but if it ever started sending an actual null as the response, this won't work as desired.
This first check `author?.id == 0L` when author is null will resolve `null == 0L` to false, meaning it will go to the else branch up above and use the null/default values for the author.

Nothing bad will happen, it just won't respect the NoAuthor state in that example unless the first check in `isDiscussionAuthorNull` is to actually check if `author == null`.

JordanMarshall: After the first check the compiler smart-casts `author` to a non-nullable type

PR #238, opened 2019-08-13
[Student][Teacher][MBL-10782] Update inbox reply functionality

Blackchai:
While digging into this I noticed that monologues weren't working at all, so those have also been fixed. See the ticket for a detailed description regarding the reply behavior changes. The TL;DR is that we try to match web behavior as closely as possible.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/fragment/InboxComposeMessageFragment.kt
+++ b/apps/student/src/main/java/com/instructure/student/fragment/InboxComposeMessageFragment.kt
@@ -163,8 +163,13 @@ class InboxComposeMessageFragment : ParentFragment() {
         recipientsView.viewTreeObserver.addOnGlobalLayoutListener(object : ViewTreeObserver.OnGlobalLayoutListener {
             override fun onGlobalLayout() {
                 if (isReply) {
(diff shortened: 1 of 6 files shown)


StephenBrough:
When replying to a single person in a group message, I'm seeing "<Name of sender>, +1". In other conversations, between two people, I see both names of the participants. Is that supposed to happen?

StephenBrough:
Also kind of weird that reply and reply all in a conversation with two people has the same functionality.

Blackchai:
> Also kind of weird that reply and reply all in a conversation with two people has the same functionality.

I agree, but that is how Canvas web works too :(.

Blackchai:
> When replying to a single person in a group message, I'm seeing ", +1". In other conversations, between two people, I see both names of the participants. Is that supposed to happen?

I'm not sure, I didn't investigate that aspect of inbox as part of this. Do you think that's something that has changed or just a potential bug you've noticed?

StephenBrough:
> I'm not sure, I didn't investigate that aspect of inbox as part of this. Do you think that's something that has changed or just a potential bug you've noticed?

Ah, I thought that was new behavior, but just tested it out on master and it's existing behavior.

PR #241, opened 2019-08-14
[Student][E2E][MBL-12976,MBL-12714]: E2E tests for inbox and settings

joehoag:


apps/student/src/androidTest/java/com/instructure/student/ui/e2e/SettingsE2ETest.kt
```diff
@@ -1,24 +1,50 @@
 package com.instructure.student.ui.e2e
```
StephenBrough: Just caught this... we should have copyrights in all our test files (looks like basically all the E2E ones don't have copyrights, and some of the other test files don't have one). We could either create a ticket to go through them, or you could do it in this PR.

joehoag: Thanks -- nice catch.  I've pushed some changes that put copyright notices at the top of all files in this PR (as well as cleaning up some imports), and I'll be on the lookout to make sure that files in future PRs have them.

joehoag:
Thanks for another review!

PR #243, opened 2019-08-15
[Student][MBL-13042] Add retry/cancel to submissions

CalvinKern:
Two extra things in this PR:
1. Fix a bug with onActivityResult event bus messages. Not removing the event was causing the event to leak to other things listening to that event.
2. Moved inserting into the database into the start service call. This will allow multiple submissions to be in progress WITH meaningful messages to users. Before, no ‘in progress’ messages would be visible to users and the submission service wouldn’t update to in progress until it was handling that submission. Now it will update everything to in progress right away, updating the actual progress of uploads as it happens. (This wasn’t an issue before as we were quickly moving through file uploads to the file upload service, meaning our submission service could quickly handle all the incoming intents. This doesn’t work as nice now that file uploads are synchronous in the submission service)

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/mobius/common/ui/SubmissionService.kt
+++ b/apps/student/src/main/java/com/instructure/student/mobius/common/ui/SubmissionService.kt
@@ -188,7 +188,7 @@ class SubmissionService : IntentService(SubmissionService::class.java.simpleName
         )
 
         // Don't show the notification in the foreground so it doesn't disappear when this service dies
-        showProgressNotification(submission.assignmentName, submission.id, false)
+        showProgressNotification(submission.assignmentName, submission.id, inForeground = false)
 
         // Handle broadcasts from file upload service
         // Register the receiver on the application context so this service can die and handle the next submission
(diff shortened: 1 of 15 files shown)


apps/student/src/main/java/com/instructure/student/mobius/common/ui/SubmissionService.kt
```diff
@@ -645,6 +533,30 @@ class SubmissionService : IntentService(SubmissionService::class.java.simpleName
         private const val CHANNEL_ID = "submissionChannel"
         const val COMMENT_CHANNEL_ID = "commentUploadChannel"
 
+        private fun getUserId() = ApiPrefs.user!!.id
+
+        private fun insertNewSubmission(assignmentId: Long, context: Context, files: List<FileSubmitObject> = emptyList(), insertBlock: (StudentDb) -> Unit): Long {
+            val db = Db.getInstance(context)
+            deleteSubmissionsForAssignment(assignmentId, db)
+            insertBlock.invoke(db)
```
StephenBrough: This could also be `insertBlock(db)`

StephenBrough:
Got a few questions, I'll talk to you when you get back in :D

StephenBrough:
The stuffs: 
If there’s already a notification, new notifications won’t show (the notification will update, but it doesn’t show up on screen until you swipe down)

(Probably a feature request) Clicking delete shows a “Deleted” toast… maybe something a bit more descriptive, like “Submission Deleted”? Or maybe a snackbar? 

(Probably a feature request) Have some kind of confirmation dialog for canceling the submission

Should the failed submission notification take you to the assignment details?

PR #246, opened 2019-08-16
[Student][UITests][MBL-12977]: Add support for landscape test run

joehoag:
Also fixed some test failures that were occurring in landscape mode.

### DIFF
--- a/libs/login-api-2/src/main/res/layout-land/activity_login_landing_page.xml
+++ b/libs/login-api-2/src/main/res/layout-land/activity_login_landing_page.xml
@@ -138,8 +138,8 @@
 
         <ImageView
             android:id="@+id/helpButton"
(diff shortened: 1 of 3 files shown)


libs/login-api-2/src/main/res/layout-land/activity_login_landing_page.xml
```diff
@@ -138,8 +138,8 @@
 
         <ImageView
             android:id="@+id/helpButton"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
+            android:layout_width="48dp"
+            android:layout_height="48dp"
```
joehoag: This was an accessibility issue that had already been fixed in the portrait version of the layout.

PR #256, opened 2019-08-22
[Student][NewRelic][MBL-13102]: Incorporate New Relic into student app

joehoag:
Notes:
-- We are just going to get "out-of-the-box" functionality from New Relic for now -- mainly interaction tracing and network tracking.  No custom metrics yet.
-- The enabling of New Relic is controlled by a Remote Config parameter.  For now, it will be on for 6% of our users, but we can tweak that (or even turn it all the way off) via Remote Config.

The remote config parameter looks like this:
Screen Shot 2019-08-22 at 12 26 08 PM

The RemoteConfigPool condition looks like this:
Screen Shot 2019-08-22 at 12 26 26 PM

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/activity/LoginActivity.kt
+++ b/apps/student/src/main/java/com/instructure/student/activity/LoginActivity.kt
@@ -19,16 +19,23 @@ package com.instructure.student.activity
 import android.content.Context
 import android.content.Intent
 import android.os.Build
+import android.os.Bundle
+import android.util.Log
 import androidx.core.content.ContextCompat
 import android.webkit.CookieManager
 import com.instructure.student.BuildConfig
 import com.instructure.student.R
 import com.instructure.canvasapi2.models.User
 import com.instructure.canvasapi2.utils.ApiPrefs
+import com.instructure.canvasapi2.utils.RemoteConfigParam
+import com.instructure.canvasapi2.utils.RemoteConfigUtils
 import com.instructure.interactions.router.Route
 import com.instructure.loginapi.login.activities.BaseLoginInitActivity
 import com.instructure.pandautils.services.PushNotificationRegistrationService
 import com.instructure.pandautils.utils.Utils
+import com.newrelic.agent.android.FeatureFlag
+import com.newrelic.agent.android.NewRelic
+
 
 class LoginActivity : BaseLoginInitActivity() {
 
@@ -65,6 +72,24 @@ class LoginActivity : BaseLoginInitActivity() {
 
     override fun isTesting(): Boolean {
         return BuildConfig.IS_TESTING
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        val startNewRelic =
+                RemoteConfigUtils.getString(RemoteConfigParam.USE_REMOTE_CONFIG)?.equals("true",ignoreCase = true) ?: false
+
+        Log.v("LoginActivity","startNewRelic=$startNewRelic")
+
+        if(startNewRelic) {
+            NewRelic.enableFeature(FeatureFlag.NetworkRequests)
+            NewRelic.enableFeature(FeatureFlag.NetworkErrorRequests)
+            NewRelic.withApplicationToken(BuildConfig.NEWRELIC_APP_TOKEN)
+                    .withCrashReportingEnabled(false) // might interfere with Crashlytics
+                    .withInteractionTracing(true)
+                    .withLoggingEnabled(false)
+                    .start(this.applicationContext)
+        }
+        super.onCreate(savedInstanceState)
     }
 
     /**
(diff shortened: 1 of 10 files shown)


instructure-cx:
Scan submitted to Checkmarx

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/RemoteConfigUtils.kt
```diff
@@ -13,7 +13,8 @@ enum class RemoteConfigParam(val rc_name: String, val safeValueAsString: String)
     TEST_BOOL("test_bool", "false"),
     TEST_FLOAT("test_float", "0f"),
     TEST_STRING("test_string", "hey there"),
-    TEST_LONG("test_long", "42")
+    TEST_LONG("test_long", "42"),
+    USE_REMOTE_CONFIG("all_remote_config", "false")
```
joehoag: "all" meaning "all apps".

joehoag: Whoa... looks like my tests have some problems.  Looking...

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1004228&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/student/src/main/java/com/instructure/student/activity/LoginActivity.kt
```diff
@@ -67,6 +74,24 @@ class LoginActivity : BaseLoginInitActivity() {
...
+        val startNewRelic =
+                RemoteConfigUtils.getString(RemoteConfigParam.USE_REMOTE_CONFIG)?.equals("true",ignoreCase = true) ?: false
+
+        Log.v("LoginActivity","startNewRelic=$startNewRelic")
+
+        if(startNewRelic) {
+            NewRelic.enableFeature(FeatureFlag.NetworkRequests)
+            NewRelic.enableFeature(FeatureFlag.NetworkErrorRequests)
+            NewRelic.withApplicationToken(BuildConfig.NEWRELIC_APP_TOKEN)
+                    .withCrashReportingEnabled(false) // might interfere with Crashlytics
```
joehoag: We could consider turning on crash reporting.  New Relic claims that its crash reporting will clash with Crashlytics, but I didn't really see any problems when I had it on during our New Relic trial.  It would be one more New Relic "out-of-the-box" thing that we could use.

apps/student/src/main/java/com/instructure/student/activity/LoginActivity.kt
```diff
@@ -67,6 +74,24 @@ class LoginActivity : BaseLoginInitActivity() {
         return BuildConfig.IS_TESTING
     }
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        val startNewRelic =
+                RemoteConfigUtils.getString(RemoteConfigParam.USE_REMOTE_CONFIG)?.equals("true",ignoreCase = true) ?: false
```
CalvinKern: Can the `RemoteConfigUtils.getBoolean()` method not be used here? Might be slightly cleaner.

apps/student/src/main/java/com/instructure/student/activity/LoginActivity.kt
CalvinKern: I'd be in favor of just having one crash service being used. I'd hate for us to have to start checking both places consistently to make sure we don't miss any crashes. If new relic has good crash reporting, we could consider switching over. Could be really interesting to use for China releases if new relic isn't blocked there (yet to determine if firebase crashlytics is blocked in China).

apps/student/src/main/java/com/instructure/student/activity/LoginActivity.kt
joehoag: I'm trying to do something safe here.  I *think* that RemoteConfigUtils.getBoolean() will throw if the string does not contain a boolean value, and I want to avoid any chance of that.  

(That begs the question, "Should we *ever* use anything besides getString()?"  Maybe not, unless you're super-confident that the Remote Config value will always be set to a valid value. 
 Or, I suppose that we could change getBoolean(), etc..., to swallow exceptions and return some nominal error value if the string format is not digestible.)

This should just get used once at startup, so the extra few cycles shouldn't matter.

apps/student/src/main/java/com/instructure/student/activity/LoginActivity.kt
joehoag: Sure.  Maybe further down the road, we could set up a separate remote config value to turn New Relic crash reporting on/off.  It might be nice to just collect a few and see how they compare/contrast with Crashlytics.

apps/student/src/main/java/com/instructure/student/activity/LoginActivity.kt
CalvinKern: Oh, good point. I'd imagine we want all those "helpers" to be safe to use for inappropriate data types. Otherwise, we could do these casts and checks in multiple places if a flag covers a large area.

Looks like the source for `String.toBoolean()` shouldn't ever result in a crash. It will check if string is null and then if it equalsIgnoreCase to "true".
But again, this is a really minor nitpick and isn't really necessary to change. Just developer preference :)

joehoag: I'm going to create a ticket to make sure that all of those helper methods are non-throwing.  (I believe you that getBoolean is non-throwing, but I'll need to look into the other two.)  Then we cal feel a little better about using them.

PR #276, opened 2019-08-29
[Teacher][MBL-11707] Add posted status to SpeedGrader grade fragment

JordanMarshall:


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1004857&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/teacher/src/main/java/com/instructure/teacher/fragments/SpeedGraderGradeFragment.kt
```diff
@@ -80,6 +92,9 @@ class SpeedGraderGradeFragment : BasePresenterFragment<SpeedGraderGradePresenter
     }
 
     override fun updateGradeText() {
+        // Show 'grade hidden' icon if the submission is graded but there is no postAt date
+        hiddenIcon.setVisible(presenter.submission?.isGradePosted)
```
CalvinKern: Pretty sure this should be an `== false` check on isGradePosted. Right now it's hiding the `hiddenIcon` when a student's grade is hidden.
Also, do we want to show this for students that don't have a grade yet when the `+` icon is showing? If so, it will need to be adjusted so they don't overlap each other.

JordanMarshall: This was bad logic on my part since the behavior we actually want from `isGradePosted` is too complex to be reduced to a single boolean value. I'll remove it from the model and add the appropriate logic inline. Also, I'll update the parameter of `isVisible` to be non-nullable so we can avoid unintended behavior like this in the future.

PR #285, opened 2019-09-04
[All][MBL-13129] Use Refresh Tokens

StephenBrough:
All 401s will go through the CanvasAuthenticator from now on (aside from 'accounts/self' calls, since they frequently return 401 when we check if the user can masquerade). This will handle any authentication refreshes or necessary log outs.

To test:
 - Check mobile dev developer key for the one created by Nate (login as site admin, click on Developer Keys in the menu)
 - Set the client id/secret using the number above the 'Show Key' button, and the string shown after pressing the 'Show Key' button respectively (I set their ApiPrefs right in CanvasAuthenticator, before it goes to make the refresh call)
 - Log in to a user on mobile dev on the app in test
 - Log into web as that user, check for a token that has an expiration time on it
 - Wait 1 hour
 - See if the app still works for that user (navigate around the app)
 - Check on web if that tokens expiration date has been updated

You probably also want to test pre-existing users to make sure that all works, since we will be using their infinite token until they log out and log back in (I tested Student and Teacher, both seemed fine).

GOOD LUCK


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1005370&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/ApiPrefs.kt
```diff
@@ -41,16 +41,33 @@ object ApiPrefs : PrefManager(PREFERENCE_FILE_NAME) {
...
 
+    @JvmStatic
+    var accessToken by StringPref()
+
+    @JvmStatic
+    var refreshToken by StringPref()
+
+    // TODO - Remove in favor of just access token once refresh is 100%
+    @JvmStatic
+    fun getValidToken(): String = if (accessToken.isNotEmpty()) accessToken else token
```
CalvinKern: May be worth making the getters private for both `token` and `accessToken` until this is removed, just to protect ourselves from ourselves.

libs/login-api-2/src/main/java/com/instructure/loginapi/login/api/MobileVerifyAPI.java
```diff
@@ -56,6 +57,7 @@ public Response intercept(Chain chain) throws IOException {
                         }
                     }
                 })
+                .authenticator(new CanvasAuthenticator())
```
CalvinKern: Do we want mobile verify to handle refresh tokens? Isn't this hitting a static domain without a logged in user?

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/ApiPrefs.kt
StephenBrough: Sounds like a good idea. I don't _think_ there are a lot of places directly pulling the access token from `ApiPrefs`, so it shouldn't be too much of a change.

libs/login-api-2/src/main/java/com/instructure/loginapi/login/api/MobileVerifyAPI.java
StephenBrough: I think you're right, we don't want it using the authenticator. Shouldn't be any authentication going on with mobile verify.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/ApiPrefs.kt
StephenBrough: Hmm... they're both delegates, and making them private would make their setters private too. 

Thinking about it some more, this might not be too big of an issue, since we shouldn't be messing with which token we're using once this is pushed up.

CalvinKern: Ah, that's a reasonable point. It would have been nice for the extra safety, but if it's not really possible without a bigger change then I'm okay without it.

Like you said, we really shouldn't be touching these tokens except for right now.

apps/parent/src/main/java/com/instructure/parentapp/activity/SignInActivity.kt
```diff
@@ -77,7 +77,7 @@ class SignInActivity : BaseLoginSignInActivity() {
                 }
                 index = url.indexOf(token)
                 if (index != -1) {
-                    ApiPrefs.token = url.substring(index + token.length)
+                    ApiPrefs.accessToken = url.substring(index + token.length)
```
CalvinKern: Hmm, this seems odd to grab the expiring access token from this url. I can't seem to figure out how to get here though, it looked like maybe creating a new account but that didn't trigger it for me (did the normal oauth token grab). Thoughts on this?

StephenBrough: Hmm... I'm thinking this is an old path used when Airwolf was still a thing. It might not actually be used anymore.

CalvinKern: Yeah, I can't figure out how to get down this path. And doing a search in canvas-lms comes up with no results for the url that's matched here `canvas/tokenReady`. Seems like this might just be old code.

PR #292, opened 2019-09-09
[Parent][MBL-13081] Fix 'No Submission' on non-submittable assignments

StephenBrough:
On paper assignments will still show "In Class", but all other non-online submissions will not show anything. They will also show 'Submitted' if there is a grade for the assignment.

Before, they would show a big red 'Missing' label if they were past due.

### DIFF
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/utils/AssignmentUtils2.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/utils/AssignmentUtils2.kt
@@ -43,7 +43,7 @@ object AssignmentUtils2 {
 
         // Case - Assignment does not take submissions, but is not 'on paper' (not graded, etc) and it has not been graded
         // Result - DUE
-        if (!assignment.isOnlineSubmissionType && !assignment.getSubmissionTypes().contains(Assignment.SubmissionType.ON_PAPER) && submission?.grade == null) {
+        if (assignment.turnInType == Assignment.TurnInType.NONE && submission?.grade == null) {
             return ASSIGNMENT_STATE_DUE
         }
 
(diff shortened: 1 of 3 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1005740&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

libs/pandautils/src/main/java/com/instructure/pandautils/utils/AssignmentUtils2.kt
```diff
@@ -42,6 +41,12 @@ object AssignmentUtils2 {
             return ASSIGNMENT_STATE_UNKNOWN
         }
 
+        // Case - Assignment does not take submissions, but is not 'on paper' (not graded, etc) and it has not been graded
+        // Result - DUE
+        if (!assignment.isOnlineSubmissionType && !assignment.getSubmissionTypes().contains(Assignment.SubmissionType.ON_PAPER) && submission?.grade == null) {
```
Blackchai: I think I would rather use assignment.turnInType != Assignment.TurnInType.NONE than the double check for type and isOnline.

StephenBrough: Hah, there's an `expectsSubmissions` method that would probably be a little bit cleaner, but it's private, and I don't know if I want to make it public. That check should work though, but it would be checking if the turnInType == NONE, rather than != NONE.

Blackchai: I don't know if the expects submission function is proper. I like the turn in type check since its also being used in a lot of places, but either is fine.

PR #312, opened 2019-09-13
[Login] Fixed false positive and added more refresh events

Blackchai:


instructure-cx:
Scan submitted to Checkmarx

PR #313, opened 2019-09-13
[Student][MBL-13200] Assignment details page: Fix submission status for graded-but-not-submitted assignments

JordanMarshall:
If an assignment had been graded but was never submitted, the assignment detail page would show that status as "Submitted". This commit fixes the status for those cases and now shows either "Not Submitted" or "Missing".


instructure-cx:
Scan submitted to Checkmarx

PR #316, opened 2019-09-13
[Student][RemoteConfig][MBL-12852]: Added simple Remote Config override capability

joehoag:
I added a UI for applying local overrides to Remote Config parameters, found in the settings menu next to Feature Flag functionality.  This allows developers to mess with the local settings of Remote Config parameters w/o messing with our global settings.

Some notes:
--The UI is drop-dead simple (and possibly ugly and inefficient).  But it only shows up in debug builds, and it gets the job done.
--I believe that parameter override(s) will persist until the next time that the app re-reads the Remote Config params from firebase.  That's about once per hour according to our current settings.
--I put as much of the logic as I could in the pandautils lib, so it shouldn't be too hard to duplicate this functionality in teacher and parent.

### DIFF
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2019 - present Instructure, Inc.
+ *
+ *     This program is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU General Public License as published by
+ *     the Free Software Foundation, version 3 of the License.
+ *
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package com.instructure.pandautils.fragments
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.EditText
+import android.widget.TextView
+import androidx.fragment.app.Fragment
+import androidx.recyclerview.widget.RecyclerView
+import com.instructure.canvasapi2.utils.RemoteConfigParam
+import com.instructure.canvasapi2.utils.RemoteConfigPrefs
+import com.instructure.pandautils.R
+import com.instructure.pandautils.utils.onTextChanged
+import com.instructure.pandautils.utils.setupAsBackButton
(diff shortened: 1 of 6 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1006245&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

joehoag:
UPDATE: It appears that your local overrides will be more or less permanent.  You will only get a Firebase Remote Config update if the parameter is updated from the Firebase console.  So you would have to manually undo your override.

libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
```diff
@@ -0,0 +1,71 @@
...
+import com.instructure.canvasapi2.utils.RemoteConfigParam
+import com.instructure.canvasapi2.utils.RemoteConfigPrefs
+import com.instructure.pandautils.R
+import com.instructure.pandautils.utils.onTextChanged
+import com.instructure.pandautils.utils.setupAsBackButton
+import kotlinx.android.synthetic.main.fragment_remote_config_params.*
+
+class RemoteConfigParamsFragment : Fragment() {
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
```
Blackchai: We could use the sweet new way of setting layouts for Fragments, Fragment(R.layout.fragment_remote_config_params)

libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    override fun onBindViewHolder(holder: RecyclerView.ViewHolder, position: Int) {
+        val param = params[position]
+        with (holder.itemView) {
+            val nameLabel = findViewById<TextView>(R.id.param_name)
+            val valueEditText = findViewById<EditText>(R.id.param_value)
+            nameLabel.text = param.rc_name + ": "
+            valueEditText.text.insert(0, RemoteConfigPrefs.getString(param.rc_name, param.safeValueAsString))
+            valueEditText.onTextChanged { newText ->
+                RemoteConfigPrefs.putString(param.rc_name, newText)
```
Blackchai: I might just be misunderstanding, but it doesn't seem ideal to set this pref every time the text changes. Would it make more sense to add a button to confirm or do we not care about anything other than the final result?

joehoag: That was why I made reference to "inefficiency" in the description for this PR.  This could be made smarter and more efficient.  If you think that that is important for a developer-facing feature, I could spend a few more hours spiffing this up.

PR #320, opened 2019-09-17
[Student][Teacher][Parent] Fix onFail to not crash with the wrong type

CalvinKern:
Wasn't able to reproduce the crash, but onFail was apparently not actually safe checking types. Hopefully this makes it work as intended.

### DIFF
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/DataResult.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/DataResult.kt
@@ -42,8 +42,7 @@ sealed class DataResult<out A> {
         return this
     }
 
-    @Suppress("UNCHECKED_CAST")
-    inline fun <FAILURE : Failure> onFail(block: (failure: FAILURE) -> Unit) : DataResult<A> {
+    inline fun <reified FAILURE : Failure> onFail(block: (failure: FAILURE) -> Unit) : DataResult<A> {
         (this as? Fail)?.let { result -> (result.failure as? FAILURE)?.let { block(it) } }
         return this
     }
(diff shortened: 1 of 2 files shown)


instructure-cx:
Scan submitted to Checkmarx

Blackchai:
> Wasn't able to reproduce the crash, but onFail was apparently not actually safe checking types. Hopefully this makes it work as intended.

Do we _also_ want to remove the explicit type cast in the CanvasAuthenticator?

CalvinKern:
> Do we also want to remove the explicit type cast in the CanvasAuthenticator?
I would prefer not to. We only want to try and reauth if there was a authorization error. Not if there was an exception or if they have a network error.

It would make this way less likely to crash again if we remove the explicit type though, so if you feel we should, we can.

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1006504&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

Blackchai:
> > Do we also want to remove the explicit type cast in the CanvasAuthenticator?
> 
> I would prefer not to. We only want to try and reauth if there was a authorization error. Not if there was an exception or if they have a network error.
> 
> It would make this way less likely to crash again if we remove the explicit type though, so if you feel we should, we can.

I hate how we have to do the "quote" reply here haha. 

So my thinking is that we actually _might_ want to be re-authing if they do have a network error/etc. It would suck if they lose network connection during a refresh and have to log back in again. But we definitely wouldn't want to get stuck in a loop.... Maybe for now we leave it, and see how often its getting hit in our analytics.

joehoag:
Nice fix + tests!

PR #326, opened 2019-09-30
[Teacher][MBL-13285] Allow teachers with only unpublished courses

CalvinKern:
To test:

Create an unpublished course with a new teacher that is only for unpublished courses. Login to the app. The current version will take you the the "Not a teacher" page, this patch allows those teachers to login to the app again.


CalvinKern:
> It looks like it should work. Maybe run a few course-related E2E tests as part of the verification process? Maybe DashboardE2ETest and AssignmentsE2ETest?

Verified locally that those tests still pass. Our E2E tests should still be okay with the seeding changes made here.

PR #343, opened 2019-10-11
[Student][Builds][MBL-13398]: Track student build stats in Sumologic

joehoag:
For now, we just report build data for successful Bitrise runs.

UPDATE:  I also stubbed out a couple of tests that won't pass consistently until we fix MBL-13405.

### DIFF
--- a/apps/student/build.gradle
+++ b/apps/student/build.gradle
@@ -50,7 +50,6 @@ androidExtensions {
     experimental = true
 }
 
-def buildType
 android {
     compileSdkVersion Versions.COMPILE_SDK
     buildToolsVersion Versions.BUILD_TOOLS
@@ -366,9 +365,10 @@ class TimingsListener implements TaskExecutionListener, BuildListener {
         def bitriseWorkflow = System.getenv("BITRISE_TRIGGERED_WORKFLOW_ID")
         def bitriseApp = System.getenv("BITRISE_APP_TITLE")
         def bitriseBranch = System.getenv("BITRISE_GIT_BRANCH")
+        def bitriseBuildNumber = System.getenv("BITRISE_BUILD_NUMBER")
         def fileSizeInMB = file.length() == 0 ? 0 : (file.length() / (1024.0 * 1024.0)).round(3)
 
-        // Instruct the builder
+        // Instruct the builder as to how to build our payload
         def payloadBuilder = new groovy.json.JsonBuilder()
         payloadBuilder event: "buildComplete",
             buildTime: cumulativeTime,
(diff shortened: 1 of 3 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1008762&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/student/build.gradle
```diff
@@ -281,4 +283,131 @@ dependencies {
     implementation Libs.NEW_RELIC
 }
 
+// A listener that will allow us to monitor task timings and, eventually, APK size
+class TimingsListener implements TaskExecutionListener, BuildListener {
```
CalvinKern: Is there any way to declare this class in a standalone file, so we can reuse this for teacher/parent if we want to? Not very familiar with gradle scripts like this, so I'm not sure if this is possible.

joehoag: That's a good idea.  I was going to tackle that issue when I extended the same logic to teacher and (maybe) parent.

apps/student/build.gradle
```diff
@@ -281,4 +283,131 @@ dependencies {
...
+
+    @Override
+    void afterExecute(Task task, TaskState taskState) {
+        def ms = TimeUnit.MILLISECONDS.convert(System.nanoTime() - startTime, TimeUnit.NANOSECONDS)
+        timings.put(task.name, ms)
+
+        //task.project.logger.warn "${task.path} took ${ms}ms"
+        if (task.name.contains("assemble")) {
+            sawAssembleTask = true
+        }
```
CalvinKern: Not sure what this `sawAssembleTask` flag is used for (dead variable?) but may be worth an OR check if the name contains "bundle" for when we switch over to bundled releases (though we will only generate bundles for release builds, not sure if this needs to consider that).

joehoag: Right now it's not used at all.  I toyed around with the idea of only reporting information if a build occurred.  Later I pivoted to just reporting from successful bitrise builds.  I can take this out.

joehoag: But I think you're write that I might have to change my apk-finding logic after your aab changes are merged in.

apps/student/build.gradle
CalvinKern: Cool. If it works now, then we can leave it like this and tackle migrating when we include other apps.

apps/student/build.gradle
CalvinKern: If it's doing this on bitrise, we will also be building a universal apk (for ease of testing). And that's probably more important to note for size than the resulting aab (as it still needs to get converted into apks and isn't actually the size we install for users).

apps/student/build.gradle
joehoag: Working on a solution...

CalvinKern:
Looks good to me! Nice work getting that class separated out!

joehoag:
Thanks for the review!

PR #366, opened 2019-11-02
Allow HTTP urls

ErikLjungdahl:
Should fix the bug where the WebView can't open HTTP URLs and only gives the error: net::ERR_CLEARTEXT_NOT_PERMITTED

Haven't actually tested it though.


StephenBrough:
Thanks for the PR!

This is actually not a bug and is intended behavior. Due to security concerns, the WebViews in Canvas mobile apps do not support cleartext.

PR #367, opened 2019-11-04
[Parent][MBL-13482] Add app secrets

CalvinKern:
This appears to be done the same exact way as we have been doing. The only secrets we really care about right now are the google-services.json, everything else is no longer being used. The docs for getting firebase (with the google-services.json) working say to put it in the platform specific modules (android, iOS) which allows us to keep using gradle with our buildSrc directory to merge in private data for the app. 

This also allows us to use all the dependencies in GlobalDependencies so that we can stay consistent for sdk targeting, generic android tools, and other things that should stay in sync between the apps.

To test:
There's not much to test here. Really just make sure that the flutter app can build. I did make some local changes to main.dart to verify that we are linked properly into firebase. You could as well by following some instructions here: https://firebase.google.com/docs/flutter/setup#try_out_an_example_app_with


instructure-cx:
Scan submitted to Checkmarx

apps/flutter_parent/android/buildSrc
```diff
@@ -0,0 +1 @@
+../../../buildSrc/
```
CalvinKern: Symbolic link to our buildSrc with our GlobalDependencies and with our private data utils.

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1010839&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

CalvinKern:
@JordanMarshall Very good point. I just did some performance testing and added it to our flutter research spreadsheet in a new tab. At the worst, 5 extra seconds were added because of this PR. On average, it adds 2 seconds. Overall, it seems like the performance will take a hit, but only on first builds. After that it seems to stabilize (for subsequent builds and for reloads).

PR #370, opened 2019-11-06
[Flutter-Parent] Add more built_value models

StephenBrough:


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1011102&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/flutter_parent/lib/models/user.dart
```diff
@@ -0,0 +1,61 @@
...
+  static Serializer<User> get serializer => _$userSerializer;
+
+  @override
+  String toString() => name;
+
+//  @override
+//  int get hashCode => id;
+//
+//  @override
+//  bool operator ==(other) => this.id == other.id;
```
CalvinKern: you can remove these hashcode, operator, and tostring functions, since built_value generates those for us. (unless there's a need to have tostring return the name only)

apps/flutter_parent/lib/models/user.dart
```diff
@@ -0,0 +1,61 @@
...
+  String get effectiveLocale;
+
+  String toJson() {
+    return json.encode(jsonSerializers.serializeWith(User.serializer, this));
+  }
+
+  static User fromJson(String jsonString) {
+    return jsonSerializers.deserializeWith(
+        User.serializer, json.decode(jsonString));
+  }
```
CalvinKern: I like how simple this is and makes it easy to serialize/deserialize models, though with Dio we'll usually get decoded json already. You can view SchoolDomain to see how this is handled with Dio vs directly using http, as it uses a helper function in the serializers file to convert the dynamic json object to the model object.

CalvinKern:
Also, you should format those files :)

(the formatter puts empty lines between each property, which imo makes it easier to read)

apps/flutter_parent/lib/models/user.dart
CalvinKern: Rather, you can view `accounts_api.dart` to see how it's getting deserialized.

CalvinKern:
Also, you need to add each of these models to the serializer's annotation, otherwise it won't know how to handle those models.

apps/flutter_parent/lib/models/user.dart
StephenBrough: Yep! I figure I'd leave them in since they were there before, but I don't think we need them.

StephenBrough:
> Also, you should format those files :)
> 
> (the formatter puts empty lines between each property, which imo makes it easier to read)

Hmm... the formatter doesn't do that for me, says it thinks there's nothing to do.

CalvinKern:
>> Also, you should format those files :)
>>
>>(the formatter puts empty lines between each property, which imo makes it easier to read)

>Hmm... the formatter doesn't do that for me, says it thinks there's nothing to do.

Strange. Welp, if I ever work in those files I'm going to format it the way my formatter wants to. lol. Not sure what settings are different between us.

apps/flutter_parent/lib/models/course.dart
```diff
@@ -80,19 +106,23 @@ abstract class Course implements Built<Course, CourseBuilder> {
...
+      ..enrollments = ListBuilder<Enrollment>()
+      ..name = ''
+      ..needsGradingCount = 0
+      ..hideFinalGrades = false
+      ..isPublic = false
+      ..applyAssignmentGroupWeights = false
+      .._isFavorite = false
+      ..accessRestrictedByDate = false
+      ..hasWeightedGradingPeriods = false
+      ..restrictEnrollmentsToCourseDates = false;
```
CalvinKern: 100% behind this idea to include initial builders, that'll be really nice for testing. But why is this private?

apps/flutter_parent/lib/models/course_assignment.dart
```diff
@@ -0,0 +1,143 @@
...
+  @BuiltValueField(wireName: "anonymous_peer_reviews")
+  bool get anonymousPeerReviews;
+
+  @BuiltValueField(wireName: "moderated_grading")
+  bool get moderatedGrading;
+
+  @BuiltValueField(wireName: "anonymous_grading")
+  bool get anonymousGrading;
+
+  bool get isArcEnabled;
```
CalvinKern: Studio

apps/flutter_parent/lib/models/course_assignment.dart
```diff
@@ -0,0 +1,143 @@
+library course_assignment;
+
+import 'package:built_value/built_value.dart';
+import 'package:built_value/serializer.dart';
+
+import 'submission.dart';
+
+part 'course_assignment.g.dart';
+
+abstract class CourseAssignment implements Built<CourseAssignment, CourseAssignmentBuilder> {
```
CalvinKern: If you're going to push up assignments now, wanna rename it to Assignment instead of CourseAssignment?

apps/flutter_parent/lib/models/course_assignment.dart
```diff
@@ -0,0 +1,143 @@
...
+  int get id;
+
+  @nullable
+  String get name;
+
+  @nullable
+  String get description;
+
+  //  @JsonKey(name: "submission_types")
+  // List<String> submissionTypesRaw = [];
```
CalvinKern: If you're going to keep this comment in, should at least update these "@JsonKey(name:" to use the built value annotation instead. (and add the get property to the declaration and remove the default value as built value doesn't allow that)

apps/flutter_parent/lib/models/course.dart
StephenBrough: I just copied the example they had. Not sure if being private is a requirement though.

apps/flutter_parent/lib/models/course_assignment.dart
StephenBrough: Ah, that's what this was for, lol, I wasn't sure if it was for something specific or was the actual assignment model. EZ.

apps/flutter_parent/lib/models/serializers.dart
```diff
@@ -16,14 +16,27 @@ library serializers;
...
 /// If changes are made, run `flutter pub run build_runner build` from the project root. Alternatively, you can
 /// have it watch for changes and automatically build if you run `flutter pub run build_runner watch`.
 @SerializersFor([
   SchoolDomain,
+  Course,
+  CourseAssignment,
+  Enrollment,
+  Grade,
+  Submission,
+  User
```
CalvinKern: Super pet peeve, can you alphabetize these names? It'll keep future additions from having any kind of conflict and looks better then just always appending to the end.

apps/flutter_parent/lib/models/submission.dart
```diff
@@ -0,0 +1,123 @@
...
+
+  @BuiltValueField(wireName: 'submission_history')
+  BuiltList<Submission> get submissionHistory;
+
+  //  val attachments: ArrayList<Attachment> = arrayListOf(),
+  @nullable
+  String get body;
+
+  //  @JsonKey(name: "rubric_assessment")
+  //  var rubricAssessment: HashMap<String, RubricCriterionAssessment> = hashMapOf(),
```
CalvinKern: Change jsonKey here if we're going to keep the comment in.

apps/flutter_parent/lib/models/course_assignment.dart
StephenBrough: I think I'll just remove them. I think it'll be easier to grow our models rather than put stuff in we may or may not need right now.

apps/flutter_parent/lib/models/course.dart
CalvinKern: Doesn't seem like it's ever getting used then...
Should either be non private or removed.

apps/flutter_parent/lib/models/course_assignment.dart
CalvinKern: Eh, okay. We probably won't need all that extra garbage for a long time so that's reasonable. Would be nice to already have it though for when we do need it, but shouldn't be too hard to add back.

apps/flutter_parent/lib/models/course.dart
StephenBrough: It seems to get used when it runs the build script. I didn't look into exactly how it works, but it's outlined here: https://github.com/google/built_value.dart/pull/731/files. Someone asked for default values and this was their solution.

apps/flutter_parent/lib/models/serializers.dart
StephenBrough: Yes!

apps/flutter_parent/lib/models/course.dart
StephenBrough: It'll get run when a builder gets created, so I'm guessing it looks like it's not being used because we don't use it in our code.

CalvinKern: Oh, that's super cool, nice find!

apps/flutter_parent/lib/models/assignment.dart
```diff
@@ -0,0 +1,118 @@
+library course_assignment;
```
CalvinKern: Missing license, also missed changing course_assignment to assignment here.

apps/flutter_parent/lib/models/assignment.dart
```diff
@@ -0,0 +1,118 @@
...
+  @BuiltValueField(wireName: "anonymous_peer_reviews")
+  bool get anonymousPeerReviews;
+
+  @BuiltValueField(wireName: "moderated_grading")
+  bool get moderatedGrading;
+
+  @BuiltValueField(wireName: "anonymous_grading")
+  bool get anonymousGrading;
+
+  bool get isArcEnabled;
```
CalvinKern: studio

apps/flutter_parent/lib/models/course_assignment.g.dart
```diff
@@ -0,0 +1,785 @@
+// GENERATED CODE - DO NOT MODIFY BY HAND
```
CalvinKern: We don't need this file anymore :)

PR #373, opened 2019-11-07
[Parent][MBL-13480] Add a lot of networking stuff

CalvinKern:
* Adds a PagedList object that can be used to return paged data sets to widgets from apis.
* Adds support for api preferences, so users, tokens, domains, etc… can now be stored app wide.
* Adds a graphql file, commented out for now as we can’t use it for parent needs, but is a lot of good stuff for when we can use graphql in parent.
* Adds the authentication apis (still on http, will convert to Dio once login UI is there to verify it works)
* Adds assignment api (also gets submissions for grade information)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1011233&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/flutter_parent/lib/main.dart
```diff
@@ -13,10 +13,13 @@
 /// along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import 'package:flutter/material.dart';
+import 'package:flutter_parent/api/utils/api_prefs.dart';
 import 'package:flutter_parent/parent_app.dart';
 import 'package:flutter_parent/utils/service_locator.dart';
 
 void main() async {
   setupLocator();
+  await ApiPrefs.init();
```
StephenBrough: I assume this is fast enough, but will this ever have a noticeable impact on loading up our app?

CalvinKern: That's a good question. I haven't looked at how long it takes, but it's just establishing a connection to the shared preferences on the device. We did this in the hack week project and nobody seemed to find any issues.

An alternative to this would be to make each call into ApiPrefs be asynchronous, and it would just do an init first thing before returning/updating values.

StephenBrough: 👍 If we didn't notice anything, then it's probably not a big deal.

PR #376, opened 2019-11-08
[Parent][MBL-13522] Web login working

CalvinKern:
Haven’t tried supporting snickerdoodles yet, tests will need to be done anyways for that.

Also, there’s some tricky rerouting logic in the base android app when the web view request fails, we don’t know when that happens in flutter. We’ll have to take a deeper look at supporting all auth flows. At least for now we’ll be able to continue development, unblocked by being able to log in.


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1011355&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/flutter_parent/lib/api/auth_api.dart
```diff
@@ -12,53 +12,66 @@
...
-  Future<MobileVerifyResult> mobileVerify(String domain) async {
+  static Future<MobileVerifyResult> mobileVerify(String domain) async {
     var userAgent = ApiPrefs.getUserAgent();
     var encodedAgent = Uri.encodeQueryComponent(userAgent);
     var encodedDomain = Uri.encodeQueryComponent(domain);
-    var url = "https://canvas.instructure.com/api/v1/mobile_verify.json?domain=$encodedDomain&user_agent=$encodedAgent";
-    var response = await http.get(url, headers: {'User-Agent': userAgent});
+
+    var response = await Dio().get(
+      "https://canvas.instructure.com/api/v1/mobile_verify.json",
```
joehoag: Is it OK to hard-code this url?

apps/flutter_parent/lib/l10n/res/intl_messages.arb
```diff
@@ -1,5 +1,5 @@
 {
-  "@@last_modified": "2019-10-30T13:15:17.774455",
+  "@@last_modified": "2019-11-08T16:05:00.369028",
   "alertsLabel": "Alerts",
```
joehoag: Is this a generated file?

apps/flutter_parent/lib/api/auth_api.dart
CalvinKern: Yeah. We will always hit this url for doing mobile verify. The only reason this would change is if we deploy into China. But if that even needs a change hasn't been decided yet.

apps/flutter_parent/lib/l10n/res/intl_messages.arb
CalvinKern: Yes. All of these 'res' files are generated. The only place we will be 'adding' strings is in the AppLocalizations class.

In fact, these probably don't need to be regenerated each time someone makes a change. This is all generated from the command:
`flutter pub run intl_translation:extract_to_arb --output-dir=lib/l10n/res lib/l10n/app_localizations.dart`

We will probably have an automated process and not have to worry about this normally.

apps/flutter_parent/test/screens/domain_search_screen_test.dart
```diff
@@ -16,20 +16,24 @@ import 'dart:async';
...
 void main() {
-  _setupLocator(MockInteractor interactor) {
+  _setupLocator(MockInteractor interactor, {WebLoginInteractor webInteractor}) {
     final _locator = GetIt.instance;
     _locator.reset();
     _locator.registerFactory<DomainSearchInteractor>(() => interactor);
+    _locator.registerFactory<WebLoginInteractor>(() => webInteractor ?? _MockWebLoginInteractor());
   }
 
   testWidgets("default state", (tester) async {
```
joehoag: Can we change all testWidgets() calls to testWidgetsWithAccessibilityChecks() calls?

CalvinKern: Oops, good call out.  I'll fix that.

apps/flutter_parent/test/screens/domain_search_screen_test.dart
```diff
@@ -321,22 +336,32 @@ void main() {
     await tester.tap(find.text("NEXT"));
     await tester.pumpAndSettle();
 
-    // TODO: Enable once WebLoginPage is implemented
-    //expect(find.byType(WebLoginPage), findsOneWidget);
+    expect(find.byType(WebLoginScreen), findsOneWidget);
   });
 
   testWidgets("Navigates to Login page from keyboard submit button", (WidgetTester tester) async {
```
joehoag: So would this be considered a good template for writing Flutter interaction tests?  In this case you use a MockInteractor and MockWebLoginInteractor. Is is always just a case of providing the correct MockInteractors?

apps/flutter_parent/test/screens/web_login_screen_test.dart
```diff
@@ -0,0 +1,164 @@
...
+    when(interactor.mobileVerify(domain))
+        .thenAnswer((_) => Future.value(MobileVerifyResult((b) => b..result = VerifyResultEnum.generalError)));
+    _setupLocator(webInteractor: interactor);
+
+    await tester.pumpWidget(TestApp(WebLoginScreen(domain)));
+    await tester.pumpAndSettle();
+
+    expect(find.byType(WebView), findsOneWidget);
+    expect(find.byType(Dialog), findsOneWidget);
+    expect(find.text('This app is not authorized for use.'), findsOneWidget);
```
joehoag: Is there a way to use the equivalent of string resource values here instead of string constants?  So that the test will still pass if we change the underlying "string resource"?

(Applies to other findText() calls as well.)

apps/flutter_parent/test/screens/domain_search_screen_test.dart
CalvinKern: Yeah, I think this is reasonable for writing interaction tests. We could instead verify that a route with the desired page was pushed, which just requires mocking the NavigationObserver passed into the app. Not sure which way is better, but they should both be equivalent.

apps/flutter_parent/test/screens/domain_search_screen_test.dart
CalvinKern: This is actually causing a lot of issues with the WebView for some reason, when hooking into the platform channels. I'll need to look into this a little bit.

apps/flutter_parent/test/screens/web_login_screen_test.dart
CalvinKern: So I didn't think it was possible, as normally you access it via `AppLocalizations.of(context).stringInQuestion`, and we don't have a context in our tests. However, turns out you can just use the normal constructor to get our standard english translations. So we can be using `AppLocalizations().stringInQuestion`. 

I'll go ahead and switch over my hard coded strings over to this.

apps/flutter_parent/test/screens/domain_search_screen_test.dart
CalvinKern: Okay, so there's some issues with the DomainSearchScreen meeting accessibility requirements. This will need to be addressed in a separate ticket. 

I was able to get the WebLoginScreen working with accessibility checks, and any screen that has a web view will need to pass a flag into the accessibility test method so that the platform views are correctly created.

CalvinKern:
Ticket to address missing accessibility tests: https://instructure.atlassian.net/browse/MBL-13555

apps/flutter_parent/test/utils/test_app.dart
```diff
@@ -46,13 +46,8 @@ class _TestAppState extends State<TestApp> {
...
 
-    // Probably don't want to do an async set state here, but it's better than calling ApiPrefs.init in _every_ test
-    ApiPrefs.init().then((_) {
-      SchedulerBinding.instance.addPostFrameCallback((_) {
-        // If we're not mounted, can't set state
-        if (mounted) setState(() => _locale = ApiPrefs.effectiveLocale());
-      });
-    });
+    // Init api prefs here so that each test doesn't have to
+    ApiPrefs.init();
```
CalvinKern: Doing a set state here causes the end widget to get rendered twice, which is undesirable in tests. If _locale is something needed in tests (shouldn't be) we can address a fix for it later.

joehoag: 👍 I don't think that Locale is high priority.

PR #377, opened 2019-11-08
[Flutter-Parent] Add code_coverage script with configurable exclusions

JordanMarshall:
This commit adds a `code_coverage` script which runs tests with coverage, removes excluded files from the coverage info, generates an HTML report, and opens the report in the browser. Exclusions can be configured by editing `tools/codecov_exclusions`.

The script can be run from the flutter project root:
`flutter pub run tools/code_coverage`

Running the script with the `--dont-open` flag will skip opening the generated report. Note that the `lcov` tool must be installed and available on the current path for this script to work. This script is meant to be used locally by devs, but will hopefully serve as a starting point when we eventually wire up proper code coverage reporting via CI.


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1011356&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

JordanMarshall:
@joehoag Just making sure I understand what `run_tests_with_coverage.sh` script is doing - it walks the directory tree up to two levels deep, executes tests for each dart and flutter package it encounters, and merges the the results into a single coverage report - is that correct? As far as I know we only have plans for developing the current project within a single package, so for now I believe it should be safe for us to use the output of the single test run executed by `tools/code_coverage.dart`. As such, I think we'll just need to update the `Run tests, report coverage` step in bitrise to run the new script in place of the old script and, with any luck, it should just work.

joehoag:
It's been a while since I looked at the script, and I never really took the time to deeply analyze it as it is mostly copied from some blog.  (Though I did need to make a few tweaks.)  But, yes, if you really like your new script, let's just modify the appropriate bitrise step to use your new script instead of the current script.  And then run it once to make sure that it properly generates/reports coverage data.

(And then delete the current script.)

joehoag:
And one more thing: I would be really careful about excluding things like lib/models/* and lib/api/*.  In the past, management has frowned on excluding such things.  IMO, we should be including all code that we write.

JordanMarshall:
Tested working with the updated bitrise step so I'll remove the old script.
> IMO, we should be including all code that we write.
I'm fine reporting on all non-generated files. There also seems to be a coverage calculation issue that shows static variables as not being covered, so I would prefer to exclude those as well. Let's talk with the rest of the team and settle on an initial set of exclusions.

PR #379, opened 2019-11-12
[Flutter-Parent][MBL-13490] Course List

StephenBrough:


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1011621&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/flutter_parent/lib/utils/service_locator.dart
```diff
@@ -13,13 +13,15 @@
...
 import 'package:flutter_parent/api/auth_api.dart';
+import 'package:flutter_parent/api/course_api.dart';
 import 'package:flutter_parent/screens/domain_search/domain_search_interactor.dart';
 import 'package:get_it/get_it.dart';
 
 GetIt locator = GetIt.instance;
 
 void setupLocator() {
   locator.registerLazySingleton<AuthApi>(() => AuthApi());
+  locator.registerLazySingleton<CourseApi>(() => CourseApi());
```
CalvinKern: Sorry about this, I'm removing AuthApi in another PR. The pattern we want to use here is registering interactors. Those interactors can use the static references to the api classes as those won't need to be mocked (but the interactor will be)

StephenBrough: Ah, I thought I had registered my interactor...

There's another approach where we use Mock for our apis, but it sounds like that won't work with static classes. For now, I think this approach works for us. I have a feeling our architecture will be evolving further as we go along.

CalvinKern: Yeah, there's no reason we can't put both the apis and interactors in locator. It will definitely change over time, lol. But you are for sure missing that interactor 😉

apps/flutter_parent/lib/screens/courses/courses_screen.dart
```diff
@@ -0,0 +1,155 @@
...
+        print(error);
+      });
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) => _content(context);
+
+  Widget _content(BuildContext context) {
+    if (_error) return Text("Error!"); // TODO: Show an error screen
```
Blackchai: Are we still wanting to keep this as a TODO? If so, do we have a ticket that makes sense to cover this work?

apps/flutter_parent/lib/screens/courses/courses_screen.dart
```diff
@@ -0,0 +1,155 @@
...
+  List<Course> _courses = [];
+
+  bool _loading = false;
+  bool _error = false;
+
+  CoursesInteractor _interactor = locator<CoursesInteractor>();
+
+  @override
+  void initState() {
+    _loadCourses();
```
Blackchai: So I've been given the impression that we shouldn't be making async calls form initState, but instead should be wrapping these calls with a futurebuilder and then referencing it when its needed. But I'm not familiar enough with best practices to know exactly why that is.

apps/flutter_parent/lib/api/course_api.dart
```diff
@@ -0,0 +1,93 @@
...
+        'favorites',
+        'current_grading_period_scores',
+        'course_image',
+        'sections',
+        'observed_users'
+      ],
+      'state': ['completed', 'available']
+    });
+
+    if (coursesResponse.statusCode == 200 || coursesResponse.statusCode == 201) {
```
Blackchai: Are we keeping this de-paginated just to keep it simple? Doesn't seem like an endpoint that would _need_ to be depaginated.

apps/flutter_parent/lib/screens/courses/courses_screen.dart
```diff
@@ -0,0 +1,155 @@
...
+
+    CourseGrade grade = course.getCourseGrade(widget._student?.id);
+    var format = NumberFormat();
+    format.maximumFractionDigits = 2;
+
+    // If there is no current grade, return 'No grade'
+    // Otherwise, we have a grade, so check if we have the actual grade string
+    // or a score
+    var text = grade.noCurrentGrade()
+        ? AppLocalizations.of(context).noGrade
```
Blackchai: Is this properly taking into account MGP?

apps/flutter_parent/lib/screens/courses/courses_screen.dart
StephenBrough: I don't think we have a ticket yet, but I also don't think we have designs for error screens or empty states yet either.

apps/flutter_parent/lib/screens/courses/courses_screen.dart
StephenBrough: This uses the CourseGrade class, which appears to be taking into account the grading periods.

apps/flutter_parent/lib/api/course_api.dart
StephenBrough: It looks like this was modeled after what we do in our native code. There might have been a use case for this, causing us to make the canvas-api-2 call depaginated. Might be better to be safe, in case there's some edge cases out there for this.

apps/flutter_parent/lib/screens/courses/courses_screen.dart
StephenBrough: FutureBuilder doesn't exactly fit this particular case. This initial part is like our loadData, where we're just grabbing the initial data, not returning views based on the Future results. I can't remember why exactly we did it this way, possibly to just separate out the initial data gathering a bit better, but it does have a nice familiar Mobius feel.

PR #384, opened 2019-11-15
[Flutter-Parent][MBL-13491] Alerts list

CalvinKern:
Also added a helper widget for centering something in a scroll view. This is useful for empty/error states when we want the user to still be able to swipe refresh, while displaying the content in the center of the screen.

### DIFF
--- a/apps/flutter_parent/lib/models/alert.dart
+++ b/apps/flutter_parent/lib/models/alert.dart
@@ -75,21 +75,21 @@ abstract class Alert implements Built<Alert, AlertBuilder> {
   // Utility functions
 
   bool isAlertInfo() {
-    return [
+    return const [
       AlertType.courseAnnouncement,
       AlertType.institutionAnnouncement,
     ].contains(alertType);
   }
 
   bool isAlertPositive() {
-    return [
+    return const [
       AlertType.assignmentGradeHigh,
       AlertType.courseGradeHigh,
     ].contains(alertType);
   }
 
   bool isAlertNegative() {
-    return [
+    return const [
       AlertType.assignmentMissing,
       AlertType.assignmentGradeLow,
       AlertType.courseGradeLow,
--- a/apps/flutter_parent/lib/models/alert.dart
+++ b/apps/flutter_parent/lib/models/alert.dart
@@ -1,0 +1,157 @@
+/// Copyright (C) 2019 - present Instructure, Inc.
+///
+/// This program is free software: you can redistribute it and/or modify
+/// it under the terms of the GNU General Public License as published by
+/// the Free Software Foundation, version 3 of the License.
+///
+/// This program is distributed in the hope that it will be useful,
+/// but WITHOUT ANY WARRANTY; without even the implied warranty of
+/// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+/// GNU General Public License for more details.
+///
+/// You should have received a copy of the GNU General Public License
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:built_collection/built_collection.dart';
+import 'package:built_value/built_value.dart';
+import 'package:built_value/serializer.dart';
+
+part 'alert.g.dart';
+
+/// To have this built_value be generated, run this command from the project root:
+/// flutter packages pub run build_runner build --delete-conflicting-outputs
+abstract class Alert implements Built<Alert, AlertBuilder> {
+  @BuiltValueSerializer(serializeNulls: true) // Add this line to get nulls to serialize when we convert to JSON
+  static Serializer<Alert> get serializer => _$alertSerializer;
+
+  Alert._();
+
+  factory Alert([void Function(AlertBuilder) updates]) = _$Alert;
+
+  int get id;
+
+  @BuiltValueField(wireName: "observer_alert_threshold_id")
+  int get observerAlertThresholdId;
+
+  @BuiltValueField(wireName: "context_type")
+  String get contextType;
+
+  @BuiltValueField(wireName: "context_id")
+  int get contextId;
+
+  @BuiltValueField(wireName: "alert_type")
+  AlertType get alertType;
+
+  @BuiltValueField(wireName: "workflow_state")
+  AlertWorkflowState get workflowState;
+
+  @BuiltValueField(wireName: "action_date")
+  DateTime get actionDate;
+
+  String get title;
+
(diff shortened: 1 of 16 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1012059&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/flutter_parent/lib/screens/alerts/alerts_interactor.dart
```diff
@@ -0,0 +1,33 @@
...
+        if (a.actionDate == null && b.actionDate != null) return -1;
+        if (a.actionDate != null && b.actionDate == null) return 1;
+        return b.actionDate.compareTo(a.actionDate);
+      });
+  }
+
+  Future<Alert> markAlertRead(int alertId) {
+    return AlertsApi.updateAlertWorkflow(alertId, AlertWorkflowState.read.name);
+  }
+}
```
joehoag: So is an interactor meant to be a thin wrapper around an api?  And I'm assuming it exists so that we can mock api operations in our tests?

apps/flutter_parent/test/screens/alerts_screen_test.dart
```diff
@@ -0,0 +1,256 @@
...
+    });
+
+    testWidgetsWithAccessibilityChecks('Does not show once loaded', (tester) async {
+      final interactor = _MockAlertsInteractor();
+      when(interactor.getAlertsForStudent(any)).thenAnswer((_) => Future.value());
+      _setupLocator(interactor: interactor);
+
+      await tester.pumpWidget(_testableWidget());
+      await tester.pump();
+      await tester.pump(); // One extra frame to finish loading
```
joehoag: Tricky.  How do you know how many times you have to pump?  Trial and error?

apps/flutter_parent/test/screens/alerts_screen_test.dart
```diff
@@ -0,0 +1,256 @@
...
+      await tester.tap(find.text(alerts.first.title));
+      await tester.pumpAndSettle();
+
+      verify(interactor.markAlertRead(alerts.first.id)).called(1);
+    });
+
+    testWidgetsWithAccessibilityChecks('Can tap course announcement alert to go to announcement', (tester) async {
+      await _pumpAndTapAlert(tester, AlertType.courseAnnouncement);
+
+      // TODO: Test that course announcement shows
```
joehoag: I'm assuming that these TODOs are waiting on future tickets to be completed?  Do we have a good way of remembering to do these when those future tickets are done?

apps/flutter_parent/lib/models/alert.dart
```diff
@@ -0,0 +1,157 @@
...
+    ..observerId = 0
+    ..htmlUrl = '';
+
+  // Utility functions
+
+  bool isAlertInfo() {
+    return [
+      AlertType.courseAnnouncement,
+      AlertType.institutionAnnouncement,
+    ].contains(alertType);
```
joehoag: Might it be advisable to have static arrays/lists for these types of checks, rather than creating the array/list every check?

joehoag:
I'm ready to approve this, once my questions are answered.

apps/flutter_parent/lib/screens/alerts/alerts_interactor.dart
CalvinKern: Yep. We want this to be a thin layer for mocking so we can test our views with data. These are supposed to not have logic unless necessary (which is why they're currently excluded). Seeing as I have logic, I will write a test and push it up. We'll eventually remove them from being excluded and count them towards code coverage probably.

But yes, in general these are a thin wrappers for API's.

apps/flutter_parent/test/screens/alerts_screen_test.dart
CalvinKern: Can be trial and error, in this case I know what the pumps are doing though. The first pump starts building the widget, the second pump lets the future created in the init state finish. 

It's breaking the black-box slightly to pump a specific number of times, but Google says that's preferred to the alternative of pumpAndSettle(), which just keeps pumping until the UI is idle.

apps/flutter_parent/test/screens/alerts_screen_test.dart
CalvinKern: Yep. We currently don't have any of these end detail screens created, so when we do we'll have to add tests here.
I just added notes to the tickets to add routing for this to the acceptance criteria.

apps/flutter_parent/lib/models/alert.dart
CalvinKern: Great feedback. I can just prepend const to these array declarations and it will make a static reference no matter how many times the methods are called. Will add that in the next commit with the interactor tests.

apps/flutter_parent/test/screens/alerts_screen_test.dart
joehoag: I kind of like pumpAndSettle(), specifically because it preserves the black-box.  But I don't feel strongly enough to hold up this PR.

apps/flutter_parent/test/screens/alerts_screen_test.dart
joehoag: 👍

apps/flutter_parent/test/screens/alerts_screen_test.dart
CalvinKern: I am pretty on the fence here. I agree that it could be better to preserve black-box and not know about fine details, but that's the exact reason Google argues that it should not just be pumpAndSettle().
https://api.flutter.dev/flutter/flutter_test/WidgetTester/pumpAndSettle.html
> In general, it is better practice to figure out exactly why each frame is needed, and then to pump exactly as many frames as necessary. This will help catch regressions where, for instance, an animation is being started one frame later than it should.

We can stay with doing a controlled number of pumps for now, but could switch things to pumpAndSettle() in the future if we notice that it's too much of a hassle or we feel that it's unnecessary details in our tests.

joehoag: 👍

joehoag:
Looks good.  Nice set of tests.

PR #385, opened 2019-11-18
[MBL-13487][MBL-13488][Student][Teacher] Add support for pronouns

JordanMarshall:
Adds support for user pronouns and updates the API models where necessary. Code that previously referenced a user name now calls a `Pronouns` helper function which returns a concatenation of the name and italicized pronouns.

Note that the `Pronouns` helper functions use hard-coded formatting; using translations for the format would complicate the process of building the mixed-style spannable used in most locations. Additionally, a review of our existing translations shows that similar formatting used elsewhere in the apps (i.e. string content followed by a space followed by additional string content wrapped in parentheses) is identical across locales. Examples: `fullVersion` and `submissionDetailsFileTabName`.

These changes only target the Teacher and Student apps. Pronoun support for the Parent app will be added as part of the Flutter project.

Usage in Student:
 - [x] Masquerade UI (“You are acting as…”)
 - [x] Previous user list in login landing page
 - [x] Course People list
 - [x] Group People list
 - [x] People Details
 - [x] Share file destination dialog
 - [x] Submission comments
 - [x] User Calendar (in calendar list)
 - [x] Assignment details: Graded discussion topic header author
 - [x] Navigation drawer: user name
 - [x] Discussion details: Topic header author
 - [x] Discussion details: reply authors
 - [x] Discussion details: reply author (self) immediately after sending
 - [x] Notification List: conversation item title (NOTE: Pronouns are not italicized here due to how the list is set up)
 - [x] Inbox list: Conversation participants
 - [x] Conversation: Message participants
 - [x] Compose message recipient chips
 - [x] Select Recipients list
 - [X] ~DocViewer comment author~

Usage in Teacher:
 - [X] ~Attendance list~ Needs LTI changes; Peyton says move ahead without this
 - [x] Edit graded discussion: ‘Assign to’ field for due date overrides
 - [x] Assignment/discussion/quiz details -> Due dates list: ‘For’ field
 - [x] Edit assignment -> Add assignees: selected assignees (text at top)
 - [x] Edit assignment -> Add assignees: list items
 - [x] Edit assignment:  ‘Assign To’ field for due date overrides
 - [x] Edit quiz:  ‘Assign To’ field for due date overrides
 - [x] Submission list
 - [x] SpeedGrader comments: Comment author
 - [x] SpeedGrader comments: Input hint
 - [x] SpeedGrader toolbar title
 - [x] SpeedGrader group members (tap toolbar for group submission)
 - [x] Discussion details: topic header author
 - [x] Discussion details: reply authors
 - [x] Discussion details: reply author (self) immediately after replying
 - [x] Inbox list: Conversation participants
 - [x] Conversation: Message participants
 - [x] Compose message recipient chips
 - [x] Select Recipients list
 - [x] People list items
 - [x] User details: toolbar title
 - [x] User details: user name
 - [x] Navigation drawer: user name
 - [x] User profile settings: user name
 - [X] ~DocViewer comment author~
 - [x] Rubric comment input hint
 - [x] Previous user list in login landing page
 - [x] Masquerade UI (“You are acting as…”)

### DIFF
--- a/apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
+++ b/apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
@@ -115,7 +115,7 @@ class SubmissionCommentsRenderTest: StudentRenderTest() {
         // This is dependent on a generated PendingSubmissionComment interface.
         pendingCommentItem = CommentItemState.PendingCommentItem(
                 authorName = user.shortName!!,
-                authorPronouns = user.pronouns,
+                authorPronouns = null,
                 avatarUrl = user.avatarUrl!!,
                 sortDate = Date(119, 6, 4, 13, 20),
                 pendingComment = makePendingComment(
@@ -168,6 +168,37 @@ class SubmissionCommentsRenderTest: StudentRenderTest() {
         )
         loadPageWithViewState(state)
         page.verifyPendingCommentPresent(pendingCommentItem, failed = false)
+    }
+
+    @Test
+    @TestMetaData(Priority.P2,FeatureCategory.ASSIGNMENTS,TestCategory.RENDER,secondaryFeature = FeatureCategory.COMMENTS)
+    fun testSingleCommentDisplaysAuthorPronoun() {
+        val commentItem = commentItemIsAudience.copy(authorPronouns = "Pro/Noun")
+        val state = SubmissionCommentsViewState(commentStates = listOf(commentItem))
+        loadPageWithViewState(state)
+        page.verifyDisplaysPronoun(commentItem)
+    }
+
+    @Test
+    @TestMetaData(Priority.P2,FeatureCategory.ASSIGNMENTS,TestCategory.RENDER,secondaryFeature = FeatureCategory.COMMENTS)
+    fun testSingleSubmissionDisplaysAuthorPronoun() {
+        val commentItem = submissionItem.copy(authorPronouns = "Pro/Noun")
+        val state = SubmissionCommentsViewState(commentStates = listOf(commentItem))
+        loadPageWithViewState(state)
+        page.verifyDisplaysPronoun(commentItem)
+    }
+
+    @Test
+    @TestMetaData(Priority.P2,FeatureCategory.ASSIGNMENTS,TestCategory.RENDER,secondaryFeature = FeatureCategory.COMMENTS)
+    fun testSinglePendingCommentDisplaysAuthorPronoun() {
+        // We shouldn't run this test on API 23 or lower, because we won't deal well
+        // with the ProgressBar that comes up.
+        if(Build.VERSION.SDK_INT < 24) return
+
+        val commentItem = pendingCommentItem.copy(authorPronouns = "Pro/Noun")
+        val state = SubmissionCommentsViewState(commentStates = listOf(commentItem))
+        loadPageWithViewState(state)
+        page.verifyDisplaysPronoun(commentItem)
     }
 
     @Test
--- a/apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
+++ b/apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
@@ -79,6 +79,7 @@ class SubmissionCommentsRenderTest: StudentRenderTest() {
 
         commentItemNotAudience = CommentItemState.CommentItem(
                 authorName = user.shortName!!,
+                authorPronouns = null,
                 id = 1,
                 avatarUrl = user.avatarUrl!!,
                 sortDate = Date(119,6,4,16,25),
@@ -90,6 +91,7 @@ class SubmissionCommentsRenderTest: StudentRenderTest() {
 
         commentItemIsAudience = CommentItemState.CommentItem(
                 authorName = teacher.shortName!!,
+                authorPronouns = null,
                 id = 2,
                 avatarUrl = teacher.avatarUrl!!,
                 sortDate = Date(119,6,4,17,25),
@@ -101,6 +103,7 @@ class SubmissionCommentsRenderTest: StudentRenderTest() {
 
         submissionItem = CommentItemState.SubmissionItem(
                 authorName = user.shortName!!,
+                authorPronouns = null,
                 avatarUrl = user.avatarUrl!!,
                 sortDate = Date(119,6,4,16,50),
                 dateText = "Date 3",
@@ -112,6 +115,7 @@ class SubmissionCommentsRenderTest: StudentRenderTest() {
         // This is dependent on a generated PendingSubmissionComment interface.
         pendingCommentItem = CommentItemState.PendingCommentItem(
                 authorName = user.shortName!!,
+                authorPronouns = user.pronouns,
                 avatarUrl = user.avatarUrl!!,
                 sortDate = Date(119, 6, 4, 13, 20),
                 pendingComment = makePendingComment(
--- a/apps/student/src/test/java/com/instructure/student/test/assignment/details/AssignmentDetailsPresenterTest.kt
+++ b/apps/student/src/test/java/com/instructure/student/test/assignment/details/AssignmentDetailsPresenterTest.kt
@@ -745,7 +745,7 @@ class AssignmentDetailsPresenterTest : Assert() {
 
         val model = baseModel.copy(assignmentResult = DataResult.Success(assignment))
         val state = AssignmentDetailsPresenter.present(model, context) as AssignmentDetailsViewState.Loaded
-        val expectedState = DiscussionHeaderViewState.Loaded(authorAvatarUrl, authorName, authoredDate, attachmentIconVisibility)
+        val expectedState = DiscussionHeaderViewState.Loaded(authorAvatarUrl, authorName, null, authoredDate, attachmentIconVisibility)
 
         assertEquals(expectedState, state.discussionHeaderViewState)
     }
@@ -762,7 +762,7 @@ class AssignmentDetailsPresenterTest : Assert() {
 
         val model = baseModel.copy(assignmentResult = DataResult.Success(assignment))
         val state = AssignmentDetailsPresenter.present(model, context) as AssignmentDetailsViewState.Loaded
-        val expectedState = DiscussionHeaderViewState.Loaded(authorAvatarUrl, authorName, authoredDate, attachmentIconVisibility)
+        val expectedState = DiscussionHeaderViewState.Loaded(authorAvatarUrl, authorName, null, authoredDate, attachmentIconVisibility)
 
         assertEquals(expectedState, state.discussionHeaderViewState)
     }
@@ -798,7 +798,7 @@ class AssignmentDetailsPresenterTest : Assert() {
 
         val model = baseModel.copy(assignmentResult = DataResult.Success(assignment))
         val state = AssignmentDetailsPresenter.present(model, context) as AssignmentDetailsViewState.Loaded
-        val expectedState = DiscussionHeaderViewState.Loaded(authorAvatarUrl, authorName, authoredDate, attachmentIconVisibility)
+        val expectedState = DiscussionHeaderViewState.Loaded(authorAvatarUrl, authorName, null, authoredDate, attachmentIconVisibility)
 
         assertEquals(expectedState, state.discussionHeaderViewState)
     }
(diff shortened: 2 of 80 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1012214&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
```diff
@@ -112,6 +115,7 @@ class SubmissionCommentsRenderTest: StudentRenderTest() {
         // This is dependent on a generated PendingSubmissionComment interface.
         pendingCommentItem = CommentItemState.PendingCommentItem(
                 authorName = user.shortName!!,
+                authorPronouns = user.pronouns,
```
joehoag: Why user.pronouns here, but null everywhere else?

apps/student/src/test/java/com/instructure/student/test/assignment/details/AssignmentDetailsPresenterTest.kt
```diff
@@ -798,7 +798,7 @@ class AssignmentDetailsPresenterTest : Assert() {
 
         val model = baseModel.copy(assignmentResult = DataResult.Success(assignment))
         val state = AssignmentDetailsPresenter.present(model, context) as AssignmentDetailsViewState.Loaded
-        val expectedState = DiscussionHeaderViewState.Loaded(authorAvatarUrl, authorName, authoredDate, attachmentIconVisibility)
+        val expectedState = DiscussionHeaderViewState.Loaded(authorAvatarUrl, authorName, null, authoredDate, attachmentIconVisibility)
```
joehoag: Consider adding tests specifically for pronouns here and in other places.

apps/teacher/src/main/java/com/instructure/teacher/binders/MessageBinder.kt
```diff
@@ -124,8 +127,11 @@ object MessageBinder : BaseBinder() {
             ""
         } else {
             when (users.size) {
-                in 0..2 -> users.joinToString { it.name ?: "" }
-                else -> "${users[0].name}, +${message.participatingUserIds.size - 1}"
+                in 0..2 -> users.joinTo(SpannableStringBuilder()) { Pronouns.span(it.name, it.pronouns) }
```
joehoag: Why is 0 treated as a special case in other similar logic?

apps/teacher/src/main/java/com/instructure/teacher/fragments/SpeedGraderCommentsFragment.kt
```diff
@@ -104,7 +105,12 @@ class SpeedGraderCommentsFragment : BaseListFragment<SubmissionCommentWrapper, S
                 intArrayOf(-android.R.attr.state_enabled) to requireContext().getColorCompat(R.color.defaultTextGray),
                 intArrayOf() to ThemePrefs.buttonColor
         )
-        if (!mGradeAnonymously) commentEditText.hint = getString(R.string.sendMessageToHint, mAssignee.name)
+        if (!mGradeAnonymously) commentEditText.hint = Pronouns.resource(
+            requireContext(),
+            R.string.sendMessageToHint,
+            mAssignee.pronouns,
+            Pronouns.span(mAssignee.name, mAssignee.pronouns)
```
joehoag: Why do you need both of these (mAssignee.pronouns, Pronouns.span(mAssignee.name, mAssignee.pronouns))?

apps/teacher/src/main/java/com/instructure/teacher/presenters/AssigneeListPresenter.kt
```diff
@@ -76,10 +77,10 @@ class AssigneeListPresenter(
             || selectedGroups.isNotEmpty()
 
     private fun updateSelectedAssignees() {
-        val assigneeNames = arrayListOf<String>()
+        val assigneeNames = arrayListOf<CharSequence>()
         assigneeNames += selectedSections.map { mSectionMap[it]!!.name }
         assigneeNames += selectedGroups.map { mGroupMap[it]?.name ?: "" }
-        assigneeNames += selectedStudents.map { mStudentMap[it]!!.name }
+        assigneeNames += selectedStudents.map { Pronouns.span(mStudentMap[it]?.name, mStudentMap[it]?.pronouns) }
```
joehoag: Why did we used to assume the existence of mStudentMap[it] in the old line, but not the new line? ("!!" in old logic, "?" in new.)

joehoag:
With the model changes, will you need to re-create the teacher ditto tapes?

joehoag:
It would be nice to see at least a few tests focusing on pronouns, even if they are just simple render tests.

joehoag:
Any chance that we will be able to seed data with pronouns for users?  That would allow us to write some E2E tests (or enhance existing ones) to test pronouns.

apps/teacher/src/main/java/com/instructure/teacher/fragments/SpeedGraderCommentsFragment.kt
CalvinKern: It's required by the Pronouns.resource() call. It provides the Pronouns.span(...) as an argument for the format string, then has to specify the pronouns to be replaced with the proper formatting (italicized). It's not ideal to have to specify that twice but makes sense that it has to go back through and find the pronouns to use the right formatting.

It does seem like it's possible to specify italics in the string resource, just requires wrapping the resource call in `Html.fromHtml(resourceString)`: https://developer.android.com/guide/topics/resources/string-resource#StylingWithHTML

May require a bunch of rework, but then it seems like it wouldn't require the pronouns to be listed twice.

CalvinKern: Hmm, actually, we're passing in the pronouns formatted string as an arg to the final resource string, so I'm not sure it would keep the formatting.

I think for that to work, each string that uses a user name (and pronouns) would have to get updated to support resource arguments for both of those, rather than just one name/pronouns argument. e.g., if a resource is "new message from %s" I think it would need to be "new message from %1$s (<i>%2$s</i>)".

If that is the case, I prefer repeating the pronouns here rather than updating every string resource that uses a name/pronoun.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/Pronouns.kt
```diff
@@ -0,0 +1,98 @@
...
+     * Returns a [Spanned] that contains the provided [name] and [pronouns], where [pronouns] will be italicized and
+     * wrapped in parentheses. If [pronouns] is not a valid string then the return value will only contain the [name].
+     */
+    @JvmStatic
+    fun span(name: String?, pronouns: String?) : Spanned {
+        val span = SpannableStringBuilder(name)
+        if (pronouns.isValid()) {
+            span.append(" ($pronouns)", StyleSpan(Typeface.ITALIC), Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+        }
+        return SpannedString(span)
```
CalvinKern: Creating span crashes for me when name is null (dang Java!). 
Not sure if it should just return an empty string span when name is null or if it should still allow an empty name with only pronouns. Probably just return the empty string span... ¯\_(ツ)_/¯

apps/student/src/androidTest/java/com/instructure/student/ui/renderTests/SubmissionCommentsRenderTest.kt
JordanMarshall: This should probably be null here as well. I'll fix when adding tests.

apps/student/src/test/java/com/instructure/student/test/assignment/details/AssignmentDetailsPresenterTest.kt
JordanMarshall: 👍

apps/teacher/src/main/java/com/instructure/teacher/binders/MessageBinder.kt
JordanMarshall: Dunno, also not sure why it initially checks for empty and then allows for a `0` case. I'll go ahead and clean that up while I'm making other changes.

apps/teacher/src/main/java/com/instructure/teacher/fragments/SpeedGraderCommentsFragment.kt
JordanMarshall: You can check out the docs for Pronouns.resource, but basically when dealing with user names that are part of a translated string we can't make initial assumptions about where the pronoun will appear, so we need to generate the user-name-with-pronoun string first, hand that to the string formatter (in this case context.getString), and then find the pronoun in the resulting formatted string so we can apply the appropriate style span.

apps/teacher/src/main/java/com/instructure/teacher/presenters/AssigneeListPresenter.kt
JordanMarshall: I _hope_ the old line wasn't resulting in any crashes, but I personally don't like assuming nullability when I don't have to so while adding Pronoun support I took the opportunity to make the change to safe calls.

JordanMarshall:
> With the model changes, will you need to re-create the teacher ditto tapes?  

Not yet, but they'll need to be regenerated when we add/change ditto tests to check for pronouns.

> It would be nice to see at least a few tests focusing on pronouns, even if they are just simple render tests.

I'll add a few tests to the assignment/submission features.

> Any chance that we will be able to seed data with pronouns for users? That would allow us to write some E2E tests (or enhance existing ones) to test pronouns.

I'm not sure tbh. I'd be surprised if we couldn't, but pronouns are behind two flags at the moment so we might want to wait for the feature to stabilize first.

CalvinKern:
Talked to you about teachers not seeing pronouns on these pages:
 Course People list
 Group People list
 People Details

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Assignee.kt
```diff
@@ -22,13 +22,15 @@ import kotlinx.android.parcel.Parcelize
...
     abstract val name: String
+    abstract val pronouns: String?
 }
 
 @Parcelize
 data class StudentAssignee(
         val student: User, // TODO: handle id and name when not given a student?
         override val id: Long = student.id,
-        override val name: String = student.name ?: ""
+        override val name: String = student.name,
```
CalvinKern: Student name is allowed to be null here now?

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Conversation.kt
```diff
@@ -150,41 +154,25 @@ data class Conversation(
...
-            var participants = ""
-            for (i in normalized.indices) {
-                if (participants != "") {
-                    participants += ", "
-                }
-
-                participants += normalized[i].name
-            }
-            participants
+            normalized.joinTo(SpannableStringBuilder())
```
CalvinKern: Nice job cleaning up all this logic!

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/StreamItem.kt
```diff
@@ -170,7 +170,7 @@ data class StreamItem(
         }
     }
 
-    fun getTitle(context: Context): String? {
+    fun getTitle(context: Context): CharSequence? {
```
CalvinKern: This seems like an unnecessary change, what was the reaason?

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Assignee.kt
JordanMarshall: Not sure why the evlis operator was there in the first place, but `User.name` is non-nullable and the linter was greatly offended by this line.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/StreamItem.kt
JordanMarshall: Ah, an artifact left over after backtracking on an effort to use Spanned in stream item titles. Turned out to be more effort than it was worth (would have required writing a custom gson type adapter for StreamItem) and this line must have been missed in the cleanup.

CalvinKern:
Nice work finding all those little places where we show student names!

PR #391, opened 2019-11-21
[Flutter-Parent][MBL-13552] Dashboard

StephenBrough:


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1012530&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/flutter_parent/test/screens/dashboard/dashboard_screen_test.dart
```diff
@@ -0,0 +1,334 @@
...
+//    await tester.pumpAndSettle();
+//
+//    // Click on Sign Out
+//    await tester.tap(find.text(AppLocalizations().signOut));
+//    await tester.pumpAndSettle();
+//
+//    // Test if we ended up on the Login Landing page
+//    expect(find.byType(LoginLandingScreen), findsOneWidget);
+//
+//    // TODO: Test if we called ApiPrefs#performLogout()
```
StephenBrough: We may want to discuss as a team how/if we want to test ApiPrefs

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
```diff
@@ -0,0 +1,334 @@
...
+        currentIndex: this._currentIndex,
+        onTap: (item) {
+          _handleBottomBarClick(item);
+        },
+      ),
+    );
+  }
+
+  Widget _appBarStudents(List<User> students) {
+    if (students.isEmpty) if (_studentsLoading) // Still loading, don't show anything
```
Blackchai: Can you please add some brackets to make this a bit more readable :)

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  Widget _navDrawer(User user) {
+    if (user == null) {
+      if (_selfLoading) {
+        // Still loading...
+        return Center(child: CircularProgressIndicator());
+      } else {
+        // Either we loaded a null user, or we got an error
+        return Text('Error');
```
Blackchai: Sounds like we don't have a lot of errors specs, would it make sense to add TODOs in those locations? Also, does Dave know we need error specs?

apps/flutter_parent/test/screens/dashboard/dashboard_interactor_test.dart
```diff
@@ -0,0 +1,57 @@
...
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'dart:math';
+
+import 'package:flutter_parent/models/enrollment.dart';
+import 'package:flutter_parent/models/user.dart';
+import 'package:flutter_parent/screens/dashboard/dashboard_interactor.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
```
Blackchai: Noice interactor tests!

apps/flutter_parent/lib/utils/widgets/common_widgets.dart
```diff
@@ -0,0 +1,20 @@
...
+/// but WITHOUT ANY WARRANTY; without even the implied warranty of
+/// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+/// GNU General Public License for more details.
+///
+/// You should have received a copy of the GNU General Public License
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:flutter/material.dart';
+
+class CommonWidgets {
```
Blackchai: Noice

apps/flutter_parent/test/screens/dashboard/dashboard_interactor_test.dart
```diff
@@ -0,0 +1,57 @@
...
+    var expectedSortedList = [_mockStudent('Alex'), _mockStudent('Billy'), _mockStudent('Zed')];
+
+    // Run the logic
+    var interactor = DashboardInteractor();
+    interactor.sortUsers(startingList);
+
+    expect(startingList, expectedSortedList);
+  });
+
+  test('interactor should filter out enrollments with no observee', () {
```
Blackchai: Maybe I misread, but you also handle duplicate observees, right? If so, might be good to add a test for that case.

apps/flutter_parent/test/screens/dashboard/dashboard_screen_test.dart
Blackchai: Fair. I think I liked how we handled it before, just confirming that the (mocked) function to set / access the pref was called.

apps/flutter_parent/lib/utils/widgets/user_name.dart
```diff
@@ -0,0 +1,41 @@
...
+  final String userName;
+  final String pronouns;
+  final TextStyle style;
+
+  UserName(this.userName, this.pronouns, {this.style});
+
+  UserName.fromUser(User user, {TextStyle style = null})
+      : userName = user.name,
+        pronouns = user.pronouns,
+        style = style;
```
CalvinKern: I believe you can still use {this.style} in the constructor params, then you don't have to specify a default nor initialize it along with the username and pronouns.

apps/flutter_parent/test/screens/dashboard/dashboard_screen_test.dart
StephenBrough: That would work. We'd have to see if the way we're mocking prefs handles storing actual values, and then check if those values are modified the way we expect them to. All our ApiPref methods are static, and it doesn't look like there's a way to mock static methods with mockito.

apps/flutter_parent/lib/utils/widgets/user_name.dart
```diff
@@ -0,0 +1,41 @@
...
+
+
+  @override
+  Widget build(BuildContext context) {
+    return Text.rich(
+      TextSpan(children: [
+        TextSpan(text: userName),
+        if (pronouns != null) TextSpan(text: ' (${pronouns})', style: TextStyle(fontStyle: FontStyle.italic))
```
CalvinKern: Should the pronouns parenthesis be localized? I feel like most languages have parenthesis, but maybe they use something else? Like Japan, they might use the full width parens, or even the brackets.

apps/flutter_parent/lib/utils/widgets/user_name.dart
```diff
@@ -0,0 +1,41 @@
...
+      ]),
+      style: style,
```
CalvinKern: Do we want to set the style if it's null? Will this replace the default or will it still allow the default theme to come through?

apps/flutter_parent/test/screens/dashboard/dashboard_interactor_test.dart
StephenBrough: YEAH! Good catch, I'll add one in.

apps/flutter_parent/lib/utils/widgets/default_avatar.dart
```diff
@@ -0,0 +1,32 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_parent/models/user.dart';
```
CalvinKern: I hate that AS formats imports above the copyright. Wanna fix this?

apps/flutter_parent/lib/utils/widgets/user_name.dart
StephenBrough: Ah, I was fiddling with that syntax trying to get it to work right. I'll try that out!

apps/flutter_parent/lib/utils/widgets/default_avatar.dart
StephenBrough: YES. If you get the _right_ import that AS wants at the top, everything else will format below it... but if you don't have that right import... 💣 I'll get that fixed.

apps/flutter_parent/lib/utils/widgets/default_avatar.dart
```diff
@@ -0,0 +1,32 @@
...
+      margin: EdgeInsets.only(top: 16.0),
+      decoration: BoxDecoration(
+          shape: BoxShape.circle,
+          image: DecorationImage(image: NetworkImage(user.avatarUrl), fit: BoxFit.contain)),
+      width: 48,
+      height: 48,
+      alignment: AlignmentDirectional.centerStart,
+    );
+  } else {
+    return CircleAvatar(child: Text('${user.sortableName.substring(0, 1)}'));
```
CalvinKern: Is this supposed to be first and last initials or just a single initial?

apps/flutter_parent/lib/utils/widgets/user_name.dart
StephenBrough: I think [Jordan's PR](https://github.com/instructure/canvas-android/pull/385/files#diff-67cfd9b593b994ecffa6970f3b53106c) left them hardcoded. Sounds like there were a few reasons for that, one being which similar formatting is basically the same across locales.

apps/flutter_parent/lib/utils/widgets/default_avatar.dart
```diff
@@ -0,0 +1,32 @@
...
+///
+/// This program is distributed in the hope that it will be useful,
+/// but WITHOUT ANY WARRANTY; without even the implied warranty of
+/// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+/// GNU General Public License for more details.
+///
+/// You should have received a copy of the GNU General Public License
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Widget userAvatar(User user) {
```
CalvinKern: Should we make this a class rather than a global function? Might make it more clean when looking at widget trees to see a UserAvatar class, but if there's more benefit to keeping this a function that's fine.

Also, missing tests for this class.

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
StephenBrough: Yep, TODOs would make sense here. I just told Dave about the specs, so he's at least aware :D

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
StephenBrough: YES

apps/flutter_parent/lib/utils/widgets/common_widgets.dart
```diff
@@ -0,0 +1,20 @@
...
+  static progressIndicator() => Center(child: CircularProgressIndicator());
+}
```
CalvinKern: Interesting. I'm not sure how to feel about this. the progress indicator is so simple this seems overkill, but at the same time, it's definitely going to be less typing for each screen that uses one.

Only bad thing about this is it adds another file for us to have counted towards code coverage, so things in here should be tested even though they're so simple.

apps/flutter_parent/lib/utils/design/parent_theme.dart
```diff
@@ -206,6 +206,7 @@ class _ParentThemeState extends State<ParentTheme> {
     return TextTheme(
       /// Design-provided styles
 
+      // Comments for each text style represent the nomenclature of the designs we have
```
CalvinKern: Great clarifying comment, thanks!

apps/flutter_parent/lib/utils/widgets/user_name.dart
StephenBrough: That's a good question, I'll dig into that one

apps/flutter_parent/lib/utils/widgets/default_avatar.dart
StephenBrough: I'll have to double check what we're doing in our native app. I don't know if everyone will have a first and last name, nor how that translates across locales.

apps/flutter_parent/lib/utils/widgets/default_avatar.dart
StephenBrough: I think it'd be better as a Widget class. Totally forgot about the tests, haha. I'll make those changes!

apps/flutter_parent/lib/utils/widgets/common_widgets.dart
StephenBrough: Yeah, I'm not sure I completely like it the way it is (being a static method), but it did make adding it simple. Might make more sense as a class and just have it in a `common_widgets` package.

apps/flutter_parent/lib/screens/dashboard/inbox_notifier.dart
```diff
@@ -0,0 +1,29 @@
...
+/// but WITHOUT ANY WARRANTY; without even the implied warranty of
+/// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+/// GNU General Public License for more details.
+///
+/// You should have received a copy of the GNU General Public License
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>. 
+
+import 'package:flutter/material.dart';
+
+final InboxCountNotifier inboxCountNotifier = new InboxCountNotifier._private();
```
CalvinKern: Do we want this to be a singleton or allow there to be multiple instances?

StephenBrough: Hmm... Probably a singleton. I don't think we'll need more than one. SINGLETON INCOMING.

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
```diff
@@ -0,0 +1,334 @@
...
+  @override
+  State<StatefulWidget> createState() => DashboardState();
+}
+
+class DashboardState extends State<DashboardScreen> {
+  DashboardInteractor _interactor = locator<DashboardInteractor>();
+
+  // Dashboard State
+  List<User> _students = [];
+  User _self;
```
CalvinKern: Hmm, this would be a huge change, but I almost feel like it would be better to make a splash screen where we load the user, rather than in this dashboard. If we support deep linking we're going to have to get the user wherever we deeplink, so having a splash screen where we can do that might be more clean overall.

apps/flutter_parent/lib/utils/widgets/common_widgets.dart
CalvinKern: I think I'd like that pattern better for sure. Nice idea!

apps/flutter_parent/lib/utils/widgets/user_name.dart
CalvinKern: That's fair. I think people will understand regardless, and if we do have to make a change it should just be here which will be really nice. Overall, love this class! Thanks for making it :)

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
CalvinKern: Or, maybe punt on a splash screen for now, and refactor for it once this is merged in? That way we can get this sweet dashboard in.

apps/flutter_parent/lib/screens/dashboard/dashboard_interactor.dart
```diff
@@ -0,0 +1,41 @@
...
+  Future<List<User>> getObservees() async =>
+    EnrollmentsApi.getObserveeEnrollments().then<List<User>>((enrollments) {
+      List<User> users = filterObservees(enrollments);
+      sortUsers(users);
+      return users;
+  });
+
+  Future<User> getSelf({app}) async =>
+    UserApi.getSelf().then((user) {
+      ApiPrefs.setUser(user, app: app);
```
CalvinKern: Missing a test for this get self call updating the user, though that's breaking a lotta black boxes to test that. But then if we stop excluding interactor tests this line will show up as missing...

Not sure what the right call here is... Thoughts?

apps/flutter_parent/lib/api/user_api.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'package:flutter_parent/models/serializers.dart';
+import 'package:flutter_parent/models/user.dart';
+
+class UserApi {
+  static Future<User> getSelf() async {
+    var selfResponse = await Dio().get(ApiPrefs.getApiUrl() + 'users/self/profile',
+        options: Options(headers: ApiPrefs.getHeaderMap()));
+
+    print(selfResponse.statusCode);
+    print(selfResponse.data);
```
CalvinKern: PRINTS!

apps/flutter_parent/test/screens/dashboard/dashboard_screen_test.dart
CalvinKern: ApiPrefs should still work as expected in tests, as in setting a new value should give the new value on a get. I have actually tested the entire ApiPrefs class and that could be used as a reference.

CalvinKern: On top of that, TestApp does all of the initialization needed for ApiPrefs, so you should be able to use it freely.

apps/flutter_parent/lib/utils/widgets/user_name.dart
StephenBrough: From the Text widget source, if the style is omitted, it will
> will use the style
 from the closest enclosing [DefaultTextStyle]

which is the same behavior for when style is null. So we should be good here 👍

apps/flutter_parent/lib/api/user_api.dart
StephenBrough: _Classic_

apps/flutter_parent/lib/utils/widgets/default_avatar.dart
StephenBrough: Looks like we're using the short name of the user to determine initials. If it's blank/null, we return a '?', 2 initials if there is _exactly_ two names in the short name. Only one initial otherwise. I'll copy the logic we have in the native app, I'm pretty sure it mimics the behavior that Canvas uses.

apps/flutter_parent/lib/utils/widgets/user_name.dart
CalvinKern: Awesome! That's good to know, thanks for finding that out.

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
StephenBrough: I like the idea of the splash screen grabbing the user. For now, I think it'll be better to leave the splash screen for another ticket.

apps/flutter_parent/lib/screens/dashboard/dashboard_interactor.dart
StephenBrough: It might be easier to test when this is in the splash screen, maybe it'll make sense to organize the logic a bit differently (async/await?) there.

apps/flutter_parent/test/screens/dashboard/dashboard_screen_test.dart
StephenBrough: Ok, I think I got something that works. Thanks for the heads up about the ApiPref tests!

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
```diff
@@ -0,0 +1,342 @@
...
+        children: <Widget>[
+          Container(
+            alignment: AlignmentDirectional.bottomStart,
+            child: Padding(
+              padding: const EdgeInsets.all(16),
+              child: FutureBuilder(
+                future: PackageInfo.fromPlatform(),
+                builder: (BuildContext context, AsyncSnapshot<PackageInfo> snapshot) {
+                  return Text(
+                    'v. ${snapshot.data?.version}',
```
CalvinKern: Should this be localized? We are localizing it in the normal apps.

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
```diff
@@ -0,0 +1,342 @@
...
+                    style: TextStyle(color: Colors.grey[500], fontSize: 12),
```
CalvinKern: This should tie into the theme for determining color, so that in dark/high contrast modes it's still visible. You'll probably want to do something like `Theme.of(context).textTheme.overline.color` to match the 'ash' color in the design.

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
```diff
@@ -0,0 +1,342 @@
...
+              onTap: () => _navigateToManageStudents(context)),
+          Divider(height: 0, indent: 16),
+          ListTile(
+              title: Text(AppLocalizations.of(context).help),
+              onTap: () => _navigateToHelp(context)),
+          Divider(height: 0, indent: 16),
+          ListTile(
+              title: Text(AppLocalizations.of(context).signOut),
+              onTap: () => _performSignOut(context)),
+          Divider(height: 0, indent: 16),
```
CalvinKern: Instead of adding these dividers in between each here, I believe you can do a ListView.separated and specify the divider once.

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
```diff
@@ -0,0 +1,342 @@
...
+    });
+  }
+
+  Widget _currentPage() {
+    if (_studentsLoading) {
+      // We're still loading students, just show a loading indicator for now
+      return CommonWidgets.progressIndicator();
+    }
+
+    print('Selected student: $_selectedStudent');
```
CalvinKern: PRINT!

apps/flutter_parent/lib/utils/common_widets/common_widgets.dart
```diff
@@ -0,0 +1,21 @@
...
+///
+/// You should have received a copy of the GNU General Public License
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:flutter/material.dart';
+
+
+class CommonWidgets {
+  static progressIndicator() => Center(child: CircularProgressIndicator());
+}
```
CalvinKern: Did you mean to delete this file since you moved progress indicator to it's own file?

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
StephenBrough: 💣

apps/flutter_parent/lib/utils/common_widets/common_widgets.dart
StephenBrough: Ah, yeah, that should be gone now.

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
StephenBrough: Interesting... I didn't think there'd be any translation differences, but it does look like there are in the native apps. I'll move that over 👍

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
StephenBrough: They aren't in a ListView, but I liked the ListView.separated better so it is now done.

apps/flutter_parent/lib/l10n/app_localizations.dart
```diff
@@ -146,6 +146,40 @@ class AppLocalizations {
...
+
+  String get signOut => Intl.message(
+      'Sign Out',
+      desc: 'Label text for the Sign Out nav drawer button'
+  );
+
+  String appVersion(String version) => Intl.message(
+    'v. $version',
+    args: [version],
+    desc: 'App version shown in the navigation drawer'
```
CalvinKern: Oh, I'm so sorry, but you do have to specify a name if this takes arguments. Not sure why that isn't enforced compile time, I can look into that. But would you mind adding that real quick? Everything else looks great!

StephenBrough: Ah, ok np!

PR #394, opened 2019-11-25
[Parent Flutter][MBL-13018] Inbox list and Create Conversation screens

JordanMarshall:
Notes:
 - HTTP cache is currently disabled, pending either a bug fix for response headers or research into alternative cache options
 - Attachment uploads currently do not trigger a foreground service. This will be addressed in a separate ticket if necessary


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1012954&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/flutter_parent/test/utils/network_image_response.dart
```diff
@@ -0,0 +1,62 @@
...
+
+// Sourced from https://github.com/flutter/flutter/blob/1ce4a4f36fc063ec78af56ea8df06e15e80c1dc8/dev/manual_tests/test/mock_image_http.dart
+
+import 'dart:io';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/mockito.dart';
+import 'package:transparent_image/transparent_image.dart';
+
+void mockNetworkImageResponse() {
```
CalvinKern: Nice work getting this to work with just a single method call from a test file!

apps/flutter_parent/lib/utils/widgets/avatar.dart
```diff
@@ -0,0 +1,65 @@
...
+  final Color backgroundColor;
+  final String url;
+  final double size;
+  final Widget overlay;
+
+  const Avatar(
+    this.url, {
+    Key key,
+    this.backgroundColor,
+    this.size = 20,
```
CalvinKern: `size` seems like the wrong name here, since it's getting doubled when actually being used. Maybe rename to radius or just bump up the size default to 40 and not double the value when passing the size?

apps/flutter_parent/lib/utils/widgets/avatar.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:transparent_image/transparent_image.dart';
+
+class Avatar extends StatelessWidget {
+  final Color backgroundColor;
+  final String url;
+  final double size;
+  final Widget overlay;
+
+  const Avatar(
+    this.url, {
```
CalvinKern: One thing to consider (maybe not now, but when we consolidate with `UserAvatar`) the url may be null for a user, and if so the initials should be used. Probably nothing to do now, but we should be aware of it.

apps/flutter_parent/assets/svg/camera.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="20" height="17" viewBox="0 0 20 17" fill="none" xmlns="http://www.w3.org/2000/svg">
```
CalvinKern: Do we need a license for assets?

apps/flutter_parent/lib/api/utils/fetch.dart
```diff
@@ -0,0 +1,51 @@
+/// Copyright (C) 2019 - present Instructure, Inc.
```
CalvinKern: I'd love to see some tests for this file, I know it's in the api directory so it's excluded, but it would be nice to have some peace of mind if we ever need to change something in here.

apps/flutter_parent/lib/l10n/app_localizations.dart
```diff
@@ -292,12 +336,48 @@ class AppLocalizations {
...
+      desc: 'Label for the action item that lets the user capture a photo using the device camera');
+
+  String get uploadFile =>
+      Intl.message('Upload File', desc: 'Label for the action item that lets the user upload a file from their device');
+
+  String get chooseFromGallery => Intl.message('Choose from Gallery',
+      desc: 'Label for the action item that lets the user select a photo from their device gallery');
+
+  String get attachmentPreparing =>
+      Intl.message('Preparing...', desc: 'Message shown while a file is being prepared to attach to a message');
```
CalvinKern: Do you know if we can use ... here or if we should use the special character like we do in normal xml strings?

apps/flutter_parent/lib/api/utils/dio_config.dart
```diff
@@ -0,0 +1,58 @@
...
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:dio/dio.dart';
+import 'package:dio_http_cache/dio_http_cache.dart';
+import 'package:flutter_parent/api/utils/api_prefs.dart';
+
+Dio canvasDio({
+  bool forceRefresh: false,
+  bool forceDeviceLanguage: false,
+  String token: null,
```
CalvinKern: I know I initially created the header map with the param name `token`, but I don't think that's descriptive enough. Especially here where it's a little out of context from the header map. I feel like this would be more meaningful if renamed to `overrideToken` or something like that, to signify that it's for special cases and not required to pull from ApiPrefs manually.

apps/flutter_parent/lib/api/utils/api_prefs_wrapper.dart
```diff
@@ -0,0 +1,61 @@
...
+  String getClientId() => ApiPrefs.getClientId();
+
+  String getClientSecret() => ApiPrefs.getClientSecret();
+
+  Map<String, String> getHeaderMap({
+    bool forceDeviceLanguage = false,
+    String token = null,
+    Map<String, String> extraHeaders = null,
+  }) =>
+      getHeaderMap(forceDeviceLanguage: forceDeviceLanguage, token: token, extraHeaders: extraHeaders);
```
CalvinKern: Is this calling itself? I think you're just missing the `ApiPrefs.` before the call.

apps/flutter_parent/lib/models/conversation.dart
```diff
@@ -0,0 +1,113 @@
...
+  /// The unique id for the conversation
+  int get id;
+
+  /// Message Subject
+  String get subject;
+
+  /// The workflowState of the conversation (unread, read, archived)
+  @BuiltValueField(wireName: "workflow_state")
+  @nullable
+  String get workflowState;
```
CalvinKern: Should this be made as an enum rather than a raw string? If you look at Alert, there's an example using the BuiltValueEnum annotation on an EnumClass.

apps/flutter_parent/lib/models/basic_user.dart
```diff
@@ -0,0 +1,37 @@
...
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:built_value/built_value.dart';
+import 'package:built_value/serializer.dart';
+
+part 'basic_user.g.dart';
+
+/// To have this built_value be generated, run this command from the project root:
+/// flutter pub run build_runner build
+abstract class BasicUser implements Built<BasicUser, BasicUserBuilder> {
```
CalvinKern: Do we gain any benefit from having this basic user class, rather than using jus the one user class and let all the other fields be null?

apps/flutter_parent/lib/api/inbox_api.dart
```diff
@@ -0,0 +1,71 @@
...
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:flutter_parent/api/utils/dio_config.dart';
+import 'package:flutter_parent/api/utils/fetch.dart';
+import 'package:flutter_parent/models/conversation.dart';
+import 'package:flutter_parent/models/course.dart';
+import 'package:flutter_parent/models/recipient.dart';
+import 'package:flutter_parent/models/unread_count.dart';
+
+class InboxApi {
```
CalvinKern: There's a few double quote strings in here you could change if you want to.
Overall, love how clean our Api's are turning out because of fetch/fetchList. Nice work!

apps/flutter_parent/lib/models/media_comment.dart
```diff
@@ -0,0 +1,50 @@
...
+  @nullable
+  String get displayName;
+
+  @nullable
+  String get url;
+
+  /// Can be either 'audio' or 'video'
+  @BuiltValueField(wireName: "media_type")
+  @nullable
+  String get mediaType;
```
CalvinKern: Should we make this an enum?

apps/flutter_parent/lib/api/utils/dio_config.dart
```diff
@@ -0,0 +1,58 @@
...
+/// GNU General Public License for more details.
+///
+/// You should have received a copy of the GNU General Public License
+/// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:dio/dio.dart';
+import 'package:dio_http_cache/dio_http_cache.dart';
+import 'package:flutter_parent/api/utils/api_prefs.dart';
+
+Dio canvasDio({
```
CalvinKern: Do you think a test for this would be beneficial? Since there's no actual network requests, it would just test that everything got configured correctly on the Dio object. Could help save us if we modify this in the future, but may not be super necessary.

apps/flutter_parent/lib/l10n/app_localizations.dart
JordanMarshall: I don't think there's a technical limitation either way, but the unicode character `…` does carry semantic meaning so it might be better to prefer that.

apps/flutter_parent/lib/api/utils/fetch.dart
JordanMarshall: I agree, but I vote moving it to a separate task/ticket as this PR is already oversized.

apps/flutter_parent/lib/utils/widgets/avatar.dart
JordanMarshall: True, let's add it to our action items.

apps/flutter_parent/lib/models/basic_user.dart
JordanMarshall: It more accurately matches the API payload, reducing ambiguity about what data is actually available.

apps/flutter_parent/lib/utils/widgets/avatar.dart
JordanMarshall: True. Let's add it to our action items.

apps/flutter_parent/lib/api/utils/fetch.dart
CalvinKern: Time for another action item 🤣

apps/flutter_parent/lib/models/basic_user.dart
CalvinKern: Hmmm, that's fair. I'm okay with this then.

apps/flutter_parent/lib/models/conversation.dart
```diff
@@ -0,0 +1,133 @@
...
+/// flutter pub run build_runner build
+abstract class Conversation implements Built<Conversation, ConversationBuilder> {
+  @BuiltValueSerializer(serializeNulls: false) // Add this line to get nulls to serialize when we convert to JSON
+  static Serializer<Conversation> get serializer => _$conversationSerializer;
+
+  /// The unique id for the conversation
+  int get id;
+
+  /// Message Subject
+  String get subject;
```
StephenBrough: Is this nullable?

apps/flutter_parent/lib/models/recipient.dart
```diff
@@ -0,0 +1,41 @@
...
+
+part 'recipient.g.dart';
+
+/// To have this built_value be generated, run this command from the project root:
+/// flutter pub run build_runner build
+abstract class Recipient implements Built<Recipient, RecipientBuilder> {
+  @BuiltValueSerializer(serializeNulls: true) // Add this line to get nulls to serialize when we convert to JSON
+  static Serializer<Recipient> get serializer => _$recipientSerializer;
+
+  int get id;
```
StephenBrough: It looks like in canvas-api-2 the `id` is returned as a string, rather than an int. Is that still the case?

apps/flutter_parent/lib/models/conversation.dart
JordanMarshall: Not as far as I know. If a message is created without a subject, the API returns an empty string.

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
```diff
@@ -277,24 +279,23 @@ class DashboardState extends State<DashboardScreen> {
...
-            null;
-          },
-          separatorBuilder: (_, c) => Divider(height: 0, indent: 16,),
-          itemCount: 5 // One extra item to get the divider on the bottom
+  _navDrawerItemsList() => ListView.separated(
+      itemBuilder: (context, idx) {
+        return (idx == 0)
+            ? _navDrawerInbox()
+            : (idx == 1)
+                ? _navDrawerManageStudents()
```
StephenBrough: I assume your formatter did this... feels like the formatter plays by it's own rules.

apps/flutter_parent/lib/models/recipient.dart
JordanMarshall: Not sure why it's a string, could have something to do with cross-shard shenanigans. @CalvinKern Do you know?

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
JordanMarshall: Yep, so that's how it will stay until the formatter behaves differently.

apps/flutter_parent/lib/screens/inbox/conversation_list/conversation_list_screen.dart
```diff
@@ -0,0 +1,369 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return DefaultParentTheme(
+      builder: (context) => Scaffold(
+        appBar: AppBar(
+          textTheme: Theme.of(context).textTheme,
+          iconTheme: Theme.of(context).iconTheme,
+          backgroundColor: Theme.of(context).scaffoldBackgroundColor,
+          title: Text(AppLocalizations.of(context).inbox),
```
StephenBrough: Was this going to use L10n?

apps/flutter_parent/lib/screens/inbox/conversation_list/conversation_list_screen.dart
```diff
@@ -0,0 +1,369 @@
...
+                ],
+              ),
+            ),
+          ),
+        );
+      },
+    );
+  }
+
+  Widget _loadingState(BuildContext context) => Center(child: CircularProgressIndicator());
```
StephenBrough: Super simple, but there is also a widget for this in `common_widets`. Not too hard to type out, but also kind of annoying how many times we'll end up using this.

JordanMarshall: Sounds like another action item

apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
StephenBrough: Until _my_ formatter behaves differently. I _did_ format this before I committed, I don't know what the difference is.

apps/flutter_parent/lib/api/utils/dio_config.dart
CalvinKern: Added an action item tot get back to this at a later time.

apps/flutter_parent/lib/models/recipient.dart
CalvinKern: Hmm, not sure. It's very possible for cross-shard stuff. Which, we may actually want to switch all id's to be strings for this reason, so that we don't have to worry about manually inflating a cross-shard id (given 123~123 -> 12300000123).
Maybe an action item to sweep through the app and update all id's to strings?

JordanMarshall: Sounds reasonable to me 👍

apps/flutter_parent/lib/screens/inbox/attachment_utils/attachment_handler.dart
```diff
@@ -0,0 +1,69 @@
...
+    if (stage == AttachmentUploadStage.UPLOADING || stage == AttachmentUploadStage.FINISHED) return;
+
+    // Move to Uploading stage
+    stage = AttachmentUploadStage.UPLOADING;
+    notifyListeners();
+
+    try {
+      // Upload the file and monitor progress
+      attachment = await locator<FileUploadApi>().uploadConversationFile(_file, (current, total) {
+        progress = total == -1 ? null : current.toDouble() / total;
```
CalvinKern: Why not set progress to -1 to represent indeterminate? That way we have a consistent way of referring/communicating what indeterminate means. (or change FileUploadApi to use null instead of -1)

apps/flutter_parent/lib/screens/inbox/attachment_utils/attachment_handler.dart
```diff
@@ -0,0 +1,69 @@
...
+  Function(AttachmentUploadStage) onStageChange;
+  double progress = null;
+  Attachment attachment;
+  AttachmentUploadStage _stage = AttachmentUploadStage.CREATED;
+
+  AttachmentUploadStage get stage => _stage;
+
+  set stage(AttachmentUploadStage stage) {
+    _stage = stage;
+    if (onStageChange != null) onStageChange(_stage);
```
CalvinKern: Is onStageChange necessary, couldn't you do a notifyListeners() instead?

apps/flutter_parent/lib/screens/inbox/attachment_utils/attachment_picker.dart
```diff
@@ -0,0 +1,107 @@
...
+
+import 'attachment_handler.dart';
+
+class AttachmentPicker extends StatefulWidget {
+  @override
+  _AttachmentPickerState createState() => _AttachmentPickerState();
+
+  static Future<AttachmentHandler> asBottomSheet(BuildContext context) {
+    return showModalBottomSheet(context: context, builder: (context) => AttachmentPicker());
+  }
```
CalvinKern: This seems like a really useful pattern, nice!

apps/flutter_parent/test/screens/inbox/attachment_utils/attachment_picker_test.dart
```diff
@@ -0,0 +1,243 @@
...
+    await tester.pump();
+
+    // Tap file option
+    await tester.tap(find.text(l10n.uploadFile));
+    await tester.pump();
+
+    verify(interactor.getFileFromDevice()).called(1);
+  });
+
+  testWidgetsWithAccessibilityChecks('Canceling camera option returns to picker ui', (tester) async {
```
CalvinKern: For these "returns to picker ui" tests, is that meaning the camera/gallery/file picker modal? If so, you may want to add another assert at the bottom that it's also still visible (finding the camera text again or something)

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_interactor.dart
```diff
@@ -0,0 +1,47 @@
...
+    String subject,
+    String body,
+    List<int> attachmentIds,
+  ) {
+    return locator<InboxApi>().createConversation(course, recipientIds, subject, body, attachmentIds);
+  }
+
+  Future<AttachmentHandler> addAttachment(BuildContext context) async {
+    return AttachmentPicker.asBottomSheet(context);
+  }
```
CalvinKern: Should probably do a widget test for this, just to make sure that it shows up on the screen.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
```diff
@@ -0,0 +1,676 @@
...
+        ),
+      ),
+    );
+  }
+
+  Widget _appBar(BuildContext context) {
+    return AppBar(
+      textTheme: Theme.of(context).textTheme,
+      iconTheme: Theme.of(context).iconTheme,
+      backgroundColor: Theme.of(context).scaffoldBackgroundColor,
```
CalvinKern: It would be nice if these things could be specified on the DefaultParentTheme. Is there something blocking that or is it just another action item to add?

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_interactor.dart
```diff
@@ -0,0 +1,47 @@
...
+import 'package:flutter_parent/screens/inbox/attachment_utils/attachment_handler.dart';
+import 'package:flutter_parent/utils/service_locator.dart';
+
+import '../attachment_utils/attachment_picker.dart';
+
+class CreateConversationInteractor {
+  Future<List<Recipient>> getAllRecipients(Course course) async {
+    var selfId = locator<ApiPrefsWrapper>().getUser().id;
+    var recipients = await locator<InboxApi>().getRecipients(course);
+    recipients.retainWhere((it) => it.id != selfId);
```
CalvinKern: Is this a design constraint? Looks like canvas allows sending a message to yourself, wondering why we aren't allowing that here.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
```diff
@@ -0,0 +1,676 @@
...
+            textAlign: TextAlign.center,
+            style: TextStyle(fontSize: 12),
+          ),
+        ],
+      ),
+    );
+  }
+
+  _finishedWidget(BuildContext context, AttachmentHandler handler) {
+    return PopupMenuButton<String>(
```
CalvinKern: Is there a way to get an ink well on these finished attachments?

CalvinKern: We may always want to show a border on finished widgets as well, if the uploaded image is white, it's very difficult to tell that the attachment is there (besides clicking on the white space and seeing the popup menu).

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
```diff
@@ -0,0 +1,676 @@
...
+                style: TextStyle(fontWeight: FontWeight.w500),
+              ),
+            ),
+          ),
+          avatar: Avatar(_selectedRecipients[0].avatarUrl),
+          backgroundColor: ParentTheme.of(context).nearSurfaceColor,
+        ),
+        if (_selectedRecipients.length > 1)
+          Text(
+            "+${_selectedRecipients.length - 1}",
```
CalvinKern: Does this need to be localized for RTL languages?

apps/flutter_parent/lib/screens/inbox/attachment_utils/attachment_handler.dart
JordanMarshall: Yes, as these stages affect state outside the scope of the ChangeNotifyProvider used for the attachment list. Specifically, the send button must be disabled if one or more attachments has not successfully finished uploading. Rather than extend the scope of the provider to the entire screen, which would cause rebuilds each time progress changes, discrete stage change events are used to notify and update the screen when necessary.

apps/flutter_parent/test/screens/inbox/attachment_utils/attachment_picker_test.dart
JordanMarshall: 🆗

apps/flutter_parent/lib/screens/inbox/attachment_utils/attachment_handler.dart
CalvinKern: Ah, that's fair. Seems a little strange to have two different ways of communicating back to the widget, but if it works and is efficient, then fine with me.

apps/flutter_parent/lib/screens/inbox/attachment_utils/attachment_handler.dart
JordanMarshall: 🤷‍♂ It comes down to choosing where the best place is to convert from a two `int` non-null progress to a single `double` nullable progress. imo the API should provide as much information as possible, so I want to keep the two-int progress there, and I'd rather not mix in nullability to those values as that adds an additional step to the process. I could set the single progress value to -1, but that just shifts 'indeterminate' logic to the UI.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_interactor.dart
JordanMarshall: The `Returns non-null result when successful` test uses this and interacts with an element on the page.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
JordanMarshall: Not sure if we should make that part of the DefaultParentTheme since it's not a universal style for non-student pages.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
JordanMarshall: Already discussed it with Dave and it's the behavior he wants.

apps/flutter_parent/lib/screens/inbox/attachment_utils/attachment_handler.dart
CalvinKern: Aaaah, booo flutter. I dislike that they use both -1 and null for indeterminate, especially since CircularProgressIndicator only handles [0,1] they should be able to allow -1 as an indeterminate value. This is fine then.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_interactor.dart
CalvinKern: Ah, you're right. Sorry I didn't notice that.
Nice job!

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
JordanMarshall: Also, regarding the inkwell, the PopupMenuSettings button does show visual feedback but it doesn't seem to be visible when there is a thumbnail. I'll look into it.

CalvinKern: UGH, DAVE!
Thanks for attempting at least.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
CalvinKern: Not sure I understand. Don't we always want to use our dark/HC theming when possible, even on the DefaultParentTheme?
Or rather, why do we have to specify different values here for textTheme and iconTheme? (scaffold makes more sense that it's custom white and not the electric blue)

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
JordanMarshall: I... _don't know._ I'll switch it to a translation to be on the safe side.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
JordanMarshall: If we didn't specify textTheme and iconTheme here it would display white text and icons on the white background.

CalvinKern: Ah okay. I understand, I was confusing the DefaultParentTheme with screens like inbox to be not themed for a student (but with a white app bar, even though I knew it didn't have that). 

Not sure how often we'll use that default parent theme that way, seems like all the current places are using it for a non-student themed screen (with white app bars).

Probably worth either making a new theme to represent that so we don't keep doing this theming setup, or switching default parent theme to be that.

Regardless, this would be another action item to get to later, if you also think that it would be useful to have.

JordanMarshall: Yeah, so just to clear up any confusion, the DefaultParentTheme is simply the normal app theme, but it always uses the blue student color regardless of which student color is actually selected. Per the designs, the white app bar isn't exclusive to either theme so I don't know if it would make sense to introduce new themes to accommodate the styling of a single element. We could probably add a 'whiteAppBar' flag to DefaultParentTheme's constructor to make it easier for most screens. It wouldn't help with student-themed screens since those aren't being wrapped with a theme widget, but maybe the usage there is rare enough that those can be exceptions. An alternative is to make and use something like a WhiteAppBar class.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_interactor.dart
JordanMarshall: Not sure why we had that in there. I spoke with product and it doesn't seem like there's any reason to restrict that, so I'll get that fixed.

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
CalvinKern: 👍 I added another row to our action items so we can address this later. Thanks for the description, that does make it more clear. I don't know why I was thinking the other way as I've used the default parent theme before /shrug

PR #408, opened 2019-12-03
[Flutter-Parent] Clean up api prefs to make it easier in tests

CalvinKern:
This will let anything using TestApp automatically setup any platform views (PlatformInfo, SharedPrefs, WebView).
The method TestApp uses is available for any test (including non-widget tests), so calling `setupPlatformViews()` will set up everything that is necessary for platform bridging things. This was made to be configurable so things can be turned on/off as needed, and adding more platform type things should be straightforward.


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1013853&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

CalvinKern:
Some things I wasn't sure on:
* Should all the platform config flags be default to false so things explicitly say what they use?
* Should we allow a user to be passed in as well to initialize ApiPrefs with?
* Is this even a good approach long term?

JordanMarshall:
> * Should all the platform config flags be default to false so things explicitly say what they use?

I kinda feel like the flags should be true for things we don't expect to control at runtime (like PackageInfo and DeviceInfo) and false for the other flags. However, for the sake of convenience it might make sense to automatically set `initPrefs` to true when `mockPrefs` is not empty.
> * Should we allow a user to be passed in as well to initialize ApiPrefs with?

I'd be fine with it if you think it will be a common occurrence, or if manually setting the user separately is too cumbersome.
> * Is this even a good approach long term?

¯\\\_(ツ)\_/¯ We probably won't know until we've used it for a bit, so I say we give it a try. _We can always change it later™_

CalvinKern:
>>* Should we allow a user to be passed in as well to initialize ApiPrefs with?

>I'd be fine with it if you think it will be a common occurrence, or if manually setting the user separately is too cumbersome.

I think this would probably be the most common one (besides getting a student, if we add that), but it still feels like it's uncommonly done. None of our current tests do it, but doesn't mean it won't come up. The alternative is that you just have to call `ApiPrefs.setUser(User())`. Which is easy and customizable, but does require knowledge that you can use ApiPrefs to set the user like that.

Maybe I'll just keep an eye on it, and if we end up needing a user in multiple tests we can add the functionality to the PlatformConfig to initialize a user as well.

On another note,
I do think it's worth setting initPrefs to true if mockPrefs is provided, I'll definitely add that in.

CalvinKern:
>However, for the sake of convenience it might make sense to automatically set initPrefs to true when mockPrefs is not empty.

Question, do we even need an initPrefs property then? In place of checking initPrefs, we could just check if mockPrefs is not null (instead of empty). Most cases don't care about what's mocked inside the prefs, just want it initialized.

Otherwise, I can keep initPrefs and mockPrefs separate, but then the PlatformConfig constructor won't be a const constructor, meaning we won't be able to specify it as a default in parameters. This gives more fine grain control over prefs initialization at the expense of losing control via defaults (we can still use fallbacks via `??`, but if someone truly wants a null platform config they wouldn't be able to do that).

JordanMarshall:
>  In place of checking initPrefs, we could just check if mockPrefs is not null (instead of empty)
👍 Works for me

PR #410, opened 2019-12-05
[Flutter-Parent][MBL-13574] Add user-friendly crash screen

JordanMarshall:


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1014128&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/flutter_parent/lib/utils/common_widgets/respawn.dart
```diff
@@ -0,0 +1,43 @@
...
+
+  static _RespawnState of(BuildContext context) {
+    return context.ancestorStateOfType(const TypeMatcher<_RespawnState>());
+  }
+}
+
+class _RespawnState extends State<Respawn> {
+  Key _key = GlobalKey();
+
+  kill() => setState(() => _key = UniqueKey());
```
CalvinKern: Kill sounds a little ambiguous here. Makes me think that it just kills the application and that's it. `restart` or `rebuild` or something like that might be better at describing what this function does.

apps/flutter_parent/lib/screens/crash_screen.dart
```diff
@@ -0,0 +1,184 @@
...
+  const CrashScreen(this.error, {Key key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        elevation: 0,
+        backgroundColor: Colors.transparent,
+        iconTheme: Theme.of(context).iconTheme,
+      ),
```
CalvinKern: Seems like this app bar should only be added if there is a back button to press. For example, if the CoursesScreen throws an error it will have this crash screen replace it in the dashboard pager, with an invisible white toolbar that doesn't go away with the scroll.

apps/flutter_parent/lib/screens/crash_screen.dart
```diff
@@ -0,0 +1,184 @@
...
+          padding: const EdgeInsets.all(32),
+          child: Column(
+            mainAxisAlignment: MainAxisAlignment.center,
+            children: <Widget>[
+              SvgPicture.asset('assets/svg/panda-not-supported.svg'),
+              SizedBox(height: 64),
+              Text(
+                L10n(context).crashScreenTitle,
+                textAlign: TextAlign.center,
+                style: TextStyle(fontSize: 20, fontWeight: FontWeight.w600),
```
CalvinKern: That's nice to know that the TextStyle just pulls the color info from the current theme. Makes doing custom styles like this pretty easy!

apps/flutter_parent/lib/screens/crash_screen.dart
```diff
@@ -0,0 +1,184 @@
...
+      // Intentionally left blank
+    }
+    return '$message\n\n${error.stack.toString()}';
+  }
+
+  _showDetailsDialog(BuildContext context, PackageInfo packageInfo, AndroidDeviceInfo deviceInfo) {
+    showDialog(
+      context: context,
+      builder: (context) => AlertDialog(
+        contentPadding: EdgeInsets.all(8),
```
CalvinKern: Not sure what the right behavior should be, but if you change this content padding to 0, then set the ListView's padding to 8, we won't have a padding border around the overscroll in the dialog. 

If it was intended to have padding around the overscroll that shows up, then ignore this.

apps/flutter_parent/lib/screens/crash_screen.dart
```diff
@@ -0,0 +1,184 @@
...
+    String message = '';
+    try {
+      message = error.exception.toString();
+    } catch (e) {
+      // Intentionally left blank
+    }
+    return '$message\n\n${error.stack.toString()}';
+  }
+
+  _showDetailsDialog(BuildContext context, PackageInfo packageInfo, AndroidDeviceInfo deviceInfo) {
```
CalvinKern: Is this details dialog (specifically the full error message) something we want general users to see? Seems helpful for us while debugging, but I don't think this will be helpful to normal people. 

Maybe for now just putting a todo to move this behind a debug build flag.

apps/flutter_parent/lib/utils/common_widgets/respawn.dart
```diff
@@ -0,0 +1,43 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:flutter/cupertino.dart';
```
CalvinKern: `cupertino`
👎

apps/flutter_parent/lib/utils/common_widgets/respawn.dart
JordanMarshall: Sure

apps/flutter_parent/lib/screens/crash_screen.dart
JordanMarshall: I'm not sure how we would know whether or not to show a back button here.

apps/flutter_parent/lib/screens/crash_screen.dart
JordanMarshall: I'll add the padding to the listview instead, but a contentPadding still needs to specified here otherwise the AlertDialog will use a default padding of 24.

CalvinKern: Right, if you specify 0 for the dialog then that makes it look the same as what you had (when moving the padding to the listview)

apps/flutter_parent/lib/screens/crash_screen.dart
JordanMarshall: I don't think it hurts anything for users to be able to view it, and David didn't seem to care either. Regardless, we'll probably want to figure out how to set up set up dev/prod flags.

apps/flutter_parent/lib/screens/crash_screen.dart
CalvinKern: Looks like it should be able to be determined via:
`ModalRoute.of(context)?.canPop ?? false`

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
```diff
@@ -153,6 +153,7 @@ class _CreateConversationScreenState extends State<CreateConversationScreen> {
 
   @override
   Widget build(BuildContext context) {
+    if (context != null) throw 'tralala';
```
CalvinKern: 👻

apps/flutter_parent/lib/screens/crash_screen.dart
CalvinKern: Agreed about the flags. And that's fine then to have this in production if Dave didn't care. Maybe users will find it interesting /shrug

apps/flutter_parent/lib/utils/crash_utils.dart
```diff
@@ -0,0 +1,29 @@
...
+class CrashUtils {
+  static init() {
+    // Set up custom crash screen
+    ErrorWidget.builder = (error) => CrashScreen(error);
+
+    // Set up error handling
+    FlutterError.onError = (error) {
+      // TODO: Report error to firebase/crashlytics
+      FlutterError.dumpErrorToConsole(error);
+    };
```
CalvinKern: Looks like this only catches Flutter errors, usually dealing with platform channels. We may want to add a general place where we handle errors from both Flutter and from Dart (which is more common), like how this cookbook shows: https://flutter.dev/docs/cookbook/maintenance/error-reporting#5-catch-and-report-dart-errors

Otherwise, we'll have two places where we'll need to be logging firebase/crashlytics/new-relic errors.

JordanMarshall: Cool, so whenever we get that 'general place' set up we can just have this call through to that.

PR #417, opened 2019-12-09
[Student][Interaction][MBL-13247]: Interaction tests for user file upload

joehoag:
Populated three p1 tests, eliminated an extraneous stubbed P1 test, moved the audio upload test from P1 to P2 since it really doesn't add anything new in terms of how we upload files.

I also fixed an a11y issue.

For uploading-from-device, uploading-from-camera and uploading-from-gallery, I mocked the handling of the intents that are expected to be launched from those actions.

BTW, I relied on the UploadFilesDialog.onFromXXX() methods as my "source of truth" for these tests.


instructure-cx:
Scan submitted to Checkmarx

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/UserFilesInteractionTest.kt
```diff
@@ -15,50 +15,208 @@
...
+        fileUploadPage.clickUpload()
+        // Should be on file list page now
+        fileListPage.refresh()
+        fileListPage.assertItemDisplayed("sample.jpg")
     }
 
     @Stub
     @Test
-    @TestMetaData(Priority.P1, FeatureCategory.FILES, TestCategory.INTERACTION, true)
+    @TestMetaData(Priority.P2, FeatureCategory.FILES, TestCategory.INTERACTION, true)
```
joehoag: I don't think that this needs to be P1.  We would upload it the same way that we upload other device files.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/UserFilesInteractionTest.kt
```diff
@@ -15,50 +15,208 @@
...
     fun testUpload_audioFile() {
         // Should be able to upload an audio file
     }
 
-    @Stub
-    @Test
-    @TestMetaData(Priority.P1, FeatureCategory.FILES, TestCategory.INTERACTION, true)
-    fun testUpload_file() {
-        // Should be able to upload a file
-    }
```
joehoag: This test made no sense to me -- aren't all of the other tests testing file upload?  So I deleted it.

libs/pandautils/src/main/res/layout/adapter_file_uploads.xml
```diff
@@ -84,12 +84,12 @@
 
         <ImageView
             android:id="@+id/removeFile"
-            android:layout_width="36dp"
-            android:layout_height="36dp"
+            android:layout_width="48dp"
+            android:layout_height="48dp"
```
joehoag: A11y issue fix.

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1014447&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/mockCanvas/MockCanvas.kt
```diff
@@ -181,6 +181,9 @@ class MockCanvas {
     /** Map of quiz submission id to quiz submission questions */
     val quizSubmissionQuestions = mutableMapOf<Long, MutableList<QuizSubmissionQuestion>>()
 
+    /** Map of user ids to user root folders */
+    val userRootFolders = mutableMapOf<Long, FileFolder>()
```
CalvinKern: Do we need to keep user folders separate from folderFiles?

CalvinKern: Ah I see, this represents the users folder, which would be hard to mimic with folderFiles.
Nevermind then, ignore this :)

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/UserFilesInteractionTest.kt
CalvinKern: Hmm, yeah. I don't think testUpload_audioFile is necessary at all. Looks like users have the option of camera, gallery, and device; so there's no need to test mock interactions with an audio file. This method can probably be deleted just like the tetstUpload_file method was.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/UserFilesInteractionTest.kt
```diff
@@ -15,50 +15,208 @@
...
+import androidx.test.rule.GrantPermissionRule
+import com.instructure.canvasapi2.utils.FileUtils
+import org.hamcrest.core.AllOf.allOf
+import org.junit.Rule
+import java.io.FileOutputStream
+import java.io.InputStream
+import java.io.OutputStream
+
 
 class UserFilesInteractionTest : StudentTest() {
```
CalvinKern: Looks like we don't have a test for creating a folder, is that something we want to do now or should we create a new ticket to track that?

CalvinKern: Hmm, it actually seems like there's quite a bit on this page that still needs interaction testing. Probably all lower than P1 though, which is why they weren't stubbed out. Maybe just punt on this then.

automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/mockCanvas/MockCanvas.kt
joehoag: (You probably mean "fileFolders", which is a map from folderId to all folders.  The naming is admittedly confusing, but in my defense, having a class named "FileFolder" that is both a file and a folder is probably the root of the confusion.)

We need to be able to retrieve a user's root folder, and that would be tough if we didn't store them separately from our generic fileFolders.  Though I *suppose* that we could search the generic folder list for context type ("user") and context id (user id) to be able to pick out the user's root folder.  

That might be more realistic, but less efficient.  What do you think?

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/UserFilesInteractionTest.kt
joehoag: It wasn't part of the original P0/P1 tests.  But it seems like a useful test.  What if I add it as a P2 test and stub it out?

automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/mockCanvas/MockCanvas.kt
joehoag: I'm making an attempt to get rid of userRootFolders...

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/UserFilesInteractionTest.kt
joehoag: ✔️ See latest commit.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/UserFilesInteractionTest.kt
joehoag: ✔️ See latest commit.

PR #436, opened 2019-12-12
Modified manage students screen to be testable for failures

CalvinKern:
This just moves the handling of errors into the widget itself, rather than relying on FutureBuilder to handle it. This was the only way I could see to test the screen based on testing Future.error (stupid flutter uncaught exceptions for that).


CalvinKern:
Note: the builds are failing since it's building against a newer version of flutter. The base branch also needs to merge in master to get the latest flutter changes.

CalvinKern:
Closing in favor of Jordan's slightly more minimal approach (where he doesn't need a flag).

PR #438, opened 2019-12-13
[Parent][UITests][MBL-13615]: Fixed flutter_parent UI tests so that they work again

joehoag:


apps/flutter_parent/android/gradle.properties
```diff
@@ -1,3 +1,4 @@
 org.gradle.jvmargs=-Xmx1536M
 android.useAndroidX=true
 android.enableJetifier=true
+android.enableR8=true
```
joehoag: This just magically appeared when I ran the UI tests.

JordanMarshall: I'm seeing the same thing on my other branch. I'm guessing this latest version of Flutter is adding it, so we probably want to keep it.

joehoag:
Thanks for the review!

PR #439, opened 2019-12-13
[Flutter-Parent][MBL-13675] Use fetch/fetchList for api calls

JordanMarshall:
Also finishes fleshing out DioConfig which helps to set up and configure Dio instances for different use cases while retaining core functinality like logging, cache options, per_page query param, etc. For example, a configuration for typical Canvas API usage can be obtained with `DioConfig.canvas()`, and a Dio instance can be obtained by calling `.dio` on that config.

For convenience, the `canvasDio()` function has been kept and now just calls through to `DioConfig.canvas().dio`

### DIFF
--- a/apps/flutter_parent/lib/api/alert_api.dart
+++ b/apps/flutter_parent/lib/api/alert_api.dart
@@ -12,47 +12,18 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import 'package:dio/dio.dart';
-import 'package:flutter_parent/api/utils/api_prefs.dart';
-import 'package:flutter_parent/api/utils/paged_list.dart';
+import 'package:flutter_parent/api/utils/dio_config.dart';
+import 'package:flutter_parent/api/utils/fetch.dart';
 import 'package:flutter_parent/models/alert.dart';
-import 'package:flutter_parent/models/serializers.dart';
 
 class AlertsApi {
   /// Alerts were depaginated in the original parent app, then sorted by date. Depaginating here to follow suite.
   Future<List<Alert>> getAlertsDepaginated(int studentId) async {
-    var response = await Dio().get(ApiPrefs.getApiUrl(path: 'users/self/observer_alerts/$studentId'),
-        options: Options(headers: ApiPrefs.getHeaderMap()));
-
-    if (response.statusCode == 200 || response.statusCode == 201) {
-      final list = PagedList<Alert>(response);
-      return (list.nextUrl == null) ? list.data : _getAlertsDepaginated(list);
-    } else {
-      return Future.error(response.statusMessage);
-    }
-  }
-
-  Future<List<Alert>> _getAlertsDepaginated(PagedList<Alert> prevList) async {
-    var response = await Dio().get(prevList.nextUrl, options: Options(headers: ApiPrefs.getHeaderMap()));
-
-    if (response.statusCode == 200 || response.statusCode == 201) {
-      prevList.updateWithResponse(response);
-      return (prevList.nextUrl == null) ? prevList.data : _getAlertsDepaginated(prevList);
-    } else {
-      return Future.error(response.statusMessage);
-    }
+    var dio = canvasDio();
+    return fetchList(canvasDio().get('users/self/observer_alerts/$studentId'), depaginateWith: dio);
   }
 
   Future<Alert> updateAlertWorkflow(int alertId, String workflowState) async {
-    var response = await Dio().put(
-      ApiPrefs.getApiUrl(path: "users/self/observer_alerts/$alertId/$workflowState"),
-      options: Options(headers: ApiPrefs.getHeaderMap()),
-    );
-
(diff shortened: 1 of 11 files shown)


apps/flutter_parent/lib/api/assignment_api.dart
```diff
@@ -12,130 +12,59 @@
...
-    } else {
-      return Future.error(assignmentResponse.statusMessage);
-    }
+    var params = {
+      'include': ['all_dates', 'overrides', 'rubric_assessment', 'submission'],
+      'order_by': 'due_at',
+      'override_assignment_dates': 'true',
+      'needs_grading_count_by_section': 'true',
+    };
+    return fetchFirstPage(canvasDio().get('courses/$courseId/assignments', queryParameters: params));
```
CalvinKern: How are subsequent pages fetched with this?

JordanMarshall: Subsequent pages can be fetched by calling `fetchNextPage` and passing the nextUrl. It's generic enough that it didn't make sense to just put it here so it went into `fetch.dart`. I didn't think too much of it since nothing is actually using it yet.

CalvinKern: Ah, I see the fetchNextPage, so we'd just need to add another method to this class to get subsequent pages. I don't think it's needed right now, so don't worry about adding it. Was mostly just curious how that flow was (then I kept reading and got that answered)

apps/flutter_parent/lib/api/utils/fetch.dart
```diff
@@ -38,6 +60,7 @@ Future<List<T>> fetchList<T>(
     var response = await request;
     if (depaginateWith == null) return deserializeList(response.data);
     depaginateWith.options.baseUrl = null;
+    depaginateWith.options.queryParameters?.remove('per_page');
```
CalvinKern: What is optimized when removing the per_page param? Not that this needs to change, just curious how this helps.

apps/flutter_parent/test/api/dio_config_test.dart
```diff
@@ -27,7 +27,15 @@ void main() {
     expect(canvasDio(), isA<Dio>());
   });
 
-  group('options', () {
+  test('DioDOnfig.canvas returns a config object', () async {
```
CalvinKern: DioD0nfig 🤣

apps/flutter_parent/test/api/dio_config_test.dart
```diff
@@ -78,4 +127,66 @@ void main() {
...
+        pageSize: PageSize(123),
+      );
+
+      expect(copy.baseUrl, '123');
+      expect(copy.baseHeaders, {'123': '123'});
+      expect(copy.cacheMaxAge, Duration(minutes: 123));
+      expect(copy.forceRefresh, false);
+      expect(copy.pageSize, PageSize(123));
+    });
+  });
```
CalvinKern: Lovely tests! ❤️

apps/flutter_parent/lib/api/alert_api.dart
```diff
@@ -12,47 +12,18 @@
...
-    var response = await Dio().get(prevList.nextUrl, options: Options(headers: ApiPrefs.getHeaderMap()));
-
-    if (response.statusCode == 200 || response.statusCode == 201) {
-      prevList.updateWithResponse(response);
-      return (prevList.nextUrl == null) ? prevList.data : _getAlertsDepaginated(prevList);
-    } else {
-      return Future.error(response.statusMessage);
-    }
+    var dio = canvasDio();
+    return fetchList(canvasDio().get('users/self/observer_alerts/$studentId'), depaginateWith: dio);
```
CalvinKern: fetchList could use the `dio` reference for the get request.

apps/flutter_parent/lib/api/auth_api.dart
```diff
@@ -14,66 +14,47 @@
...
-      },
-      options: Options(headers: {'User-Agent': userAgent}),
+    Dio dio = DioConfig.core().dio;
+    String userAgent = ApiPrefs.getUserAgent();
+    return fetch(
+      dio.get(
+        'mobile_verify.json',
+        queryParameters: {
+          'domain': domain,
+          'user_agent': userAgent,
```
CalvinKern: That's nice that these don't need to be encoded. Nice work simplifying the code!

apps/flutter_parent/lib/api/utils/fetch.dart
JordanMarshall: The nextUrl already has the par_page param so we were accidentally duplicating it before

apps/flutter_parent/test/api/dio_config_test.dart
JordanMarshall: 🤣

apps/flutter_parent/lib/api/alert_api.dart
JordanMarshall: Good catch

PR #440, opened 2019-12-13
[Flutter-Parent][MBL-13685] Launcher config

CalvinKern:


JordanMarshall:
Just a small polish item, but could you update `launch_background.xml` to use the blue parent color instead of white?

CalvinKern:
>Just a small polish item, but could you update launch_background.xml to use the blue parent color instead of white?

The background color is what gets displayed behind the actual canvas logo (which is blue). So changing the background to blue would require us to also change the logo to be white. I'm would be okay with us making this change, kinda sleek, but that does invert the way that we have the parent app icon right now.

CalvinKern:
Added `settings_aar.gradle`, as it keeps being generated when I run the app. Part of the flutter update possibly?

PR #466, opened 2020-01-07
[Student][MBL-13741] Added check before showing error dialog in Sign In

Blackchai:
https://console.firebase.google.com/u/1/project/automatic-rite-88319/crashlytics/app/android:com.instructure.candroid/issues/5c941d7df8b88c2963044293?time=last-seven-days&sessionId=5E136697036A00017E1BDF1CE8860107_DNE_0_v2

### DIFF
--- a/libs/login-api-2/src/main/java/com/instructure/loginapi/login/activities/BaseLoginSignInActivity.java
+++ b/libs/login-api-2/src/main/java/com/instructure/loginapi/login/activities/BaseLoginSignInActivity.java
@@ -393,12 +393,12 @@ public abstract class BaseLoginSignInActivity extends AppCompatActivity implemen
                     errorId = R.string.mobileVerifyUnknownError;
                 }
 
-                AlertDialog.Builder builder = new AlertDialog.Builder(BaseLoginSignInActivity.this);
-                builder.setTitle(R.string.errorOccurred);
-                builder.setMessage(errorId);
-                builder.setCancelable(true);
-                AlertDialog dialog = builder.create();
                 if(!(BaseLoginSignInActivity.this).isFinishing()) {
+                    AlertDialog.Builder builder = new AlertDialog.Builder(BaseLoginSignInActivity.this);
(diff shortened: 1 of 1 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1017594&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

libs/login-api-2/src/main/java/com/instructure/loginapi/login/activities/BaseLoginSignInActivity.java
```diff
@@ -398,7 +398,9 @@ public void onResponse(@NonNull Response<DomainVerificationResult> response, @No
                 builder.setMessage(errorId);
                 builder.setCancelable(true);
                 AlertDialog dialog = builder.create();
-                dialog.show();
+                if(!(BaseLoginSignInActivity.this).isFinishing()) {
+                    dialog.show();
+                }
```
joehoag: Maybe put this entire builder/dialog-setup block inside the if-clause?   

Also, is there something more general than isFinishing() that we could use?  Might it be useful to check whether the activity has been destroyed with isDestroyed(), or something like that, in addition to the isFinishing() check?

JordanMarshall: After looking at Activity's source, it looks like `isFinishing()` should be safe to use here as it will return true even after the activity is destroyed and the 'finishing' process is complete. You can thank Android's docs for not being clear on that.

Blackchai: Thanks :+1:

PR #467, opened 2020-01-07
[Student][Teacher][MBL-13725] Don’t filter notifications

CalvinKern:
Filtering notifications is problematic, as the domain users log in with may be different that the account’s root domain. This makes it hard to store notifications based on user info. Instead, we will just always show notifications.

Instead of keeping old communication channels alive, we now delete them when logging out or switching users. This prevents the currently logged in account from seeing notifications meant for an account which they don’t have access to.

To test: (student and teacher)
Log into an account, verify you can get notifications.
Switch users, verify notifications disappear from the status bar.
Log out, verify you no don't get any notifications.
*vanity domain* would be nice to test, don't think we have an instance of this (should be able to log in via the non vanity domain and receive notifications).

### DIFF
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/CommunicationChannelsManager.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/CommunicationChannelsManager.kt
@@ -48,8 +48,6 @@ object CommunicationChannelsManager {
     }
 
     fun deletePushCommunicationChannelSynchronous(registrationId: String) {
-        val adapter = RestBuilder()
-        val params = RestParams()
-        return CommunicationChannelsAPI.deletePushCommunicationChannelSynchronous(registrationId, adapter, params)
+        return CommunicationChannelsAPI.deletePushCommunicationChannelSynchronous(registrationId)
     }
 }
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/CommunicationChannelsManager.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/CommunicationChannelsManager.kt
@@ -20,6 +20,7 @@ import com.instructure.canvasapi2.apis.CommunicationChannelsAPI
 import com.instructure.canvasapi2.builders.RestBuilder
 import com.instructure.canvasapi2.builders.RestParams
 import com.instructure.canvasapi2.models.CommunicationChannel
+import com.instructure.canvasapi2.utils.weave.apiAsync
 import okhttp3.ResponseBody
 import retrofit2.Response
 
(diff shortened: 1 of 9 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1017600&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/CommunicationChannelsManager.kt
```diff
@@ -45,4 +46,10 @@ object CommunicationChannelsManager {
         val params = RestParams(isForceReadFromNetwork = true)
         return CommunicationChannelsAPI.addNewPushCommunicationChannelSynchronous(registrationId, adapter, params)
     }
+
+    fun deletePushCommunicationChannelSynchronous(registrationId: String) {
+        val adapter = RestBuilder()
```
Blackchai: Suuuuuper minor, and totally not a blocker, but since we aren't doing anything with params from the manager we could move the adapter and the params into the API implementation.

libs/pandautils/src/main/java/com/instructure/pandautils/receivers/PushExternalReceiver.kt
```diff
@@ -145,7 +109,7 @@ abstract class PushExternalReceiver : FirebaseMessagingService() {
 
             createNotificationChannel(context, CHANNEL_PUSH_GENERAL, loginId, nm)
 
-            nm.notify(555443, notification)
+            nm.notify(PushNotification.NOTIFY_ID, notification)
```
Blackchai: Good change. :+1:

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/CommunicationChannelsManager.kt
CalvinKern: Oh, if that's a pattern we use I'm totally down to do that. Felt weird declaring those here since there's nothing special happening, but that's how it seemed managers were done so I just kept marching along.

Blackchai: I'm really not sure I have a strong preference. We pretty much always leave them in the manager, regardless of whether we need to modify them or not. I guess the benefit is that it keeps the actual API class looking consistent and super brief. Let's leave this as is for now, it so minor either way imo.

instructure-cx:
Scan submitted to Checkmarx

PR #476, opened 2020-01-13
[Student][Test][MBL-13793]: Speculative fixes to stabilize Grades E2E test

joehoag:
This was a little frustrating to chase, because all 4 failures/retries showed slightly different causes and crash locations.  And also because the test intermittently passes in Bitrise and runs flawlessly locally.  

I don't know for sure what is going wrong, but what I added has a chance to address all flavors of the failures that we were seeing.

### DIFF
--- a/apps/student/src/androidTest/java/com/instructure/student/ui/pages/CourseGradesPage.kt
+++ b/apps/student/src/androidTest/java/com/instructure/student/ui/pages/CourseGradesPage.kt
@@ -16,6 +16,7 @@
  */
 package com.instructure.student.ui.pages
 
+import android.os.SystemClock.sleep
 import android.view.View
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.action.ViewActions.click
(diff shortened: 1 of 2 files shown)


apps/student/src/androidTest/java/com/instructure/student/ui/pages/CourseGradesPage.kt
```diff
@@ -71,6 +72,7 @@ class CourseGradesPage : BasePage(R.id.courseGradesPage) {
     fun refresh() {
         onView(allOf(withId(R.id.swipeRefreshLayout), isDisplayed()))
                 .perform(withCustomConstraints(swipeDown(), isDisplayingAtLeast(5)))
+        sleep(1000) // Allow some time to react to the update.
```
CalvinKern: May be good to document that this is speculative, so that we know we can remove it if it's still failing. That way we don't keep these test times inflated.

joehoag: I'll keep the ticket open for a few days and see how we do.  (We run the E2E tests every night.)  If it seems like we've fixed the issue, I'll leave the code as is.  If not, I'll make additional changes, possibly undoing these changes.

instructure-cx:
Scan submitted to Checkmarx

PR #478, opened 2020-01-14
[Flutter Parent][MBL-13743] Refresh Indicator design check

StephenBrough:
Looks like we are following our loading/refresh indicator pattern we've established.

This is just something I noticed while poking around.


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
Scan submitted to Checkmarx

PR #479, opened 2020-01-14
[All][Pact][MBL-13408]: Client-side Pact logic for courses + some others

joehoag:
Keep in mind that this is only the client-side logic.  We'll need to do another round where we work with the server-side folks to actually get these published to the Pact broker and make sure that they are compatible with existing endpoints.  That will probably require some tweaks to these tests, and those will be reviewed in a separate PR.

My primary goal with this ticket was to write tests for the Courses API, but as part of "kicking the tires" on my design I also wrote tests for the Enrollments and Users APIs.  Note that I am very purposely not writing a test for every endpoint, just trying to pick a few representative endpoints for each API.

When reviewing this, please take a 10,000-foot approach to the review.  Don't focus so much on every line of code.  Instead, please review:
(1) The general pattern of having generic populateXXXFields() extension methods for generating field specifications for an object, assertXXXPopulated() methods for checking that an object in a response has been populated correctly and, in most cases, and PactXXXFieldConfig classes/objects that govern which fields should be populated in various scenarios.
(2) The endpoints being tested in each API (Users, Courses, Enrollments).  Is this a good representative set of endpoints to test, or are there others that should be added?  Or are some of the tests so similar as to be redundant?
(3) The strategy of putting all Pact tests for an API into a single code module.
(4) The location of the pact tests and accompanying logic.


libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/apis/CourseAPI.kt
```diff
@@ -29,7 +29,7 @@ import java.io.IOException
 
 object CourseAPI {
 
-    internal interface CoursesInterface {
+    interface CoursesInterface {
```
joehoag: I believe that this is the only piece of production code that I changed.  (This and the build.gradle change above.)

instructure-cx:
Scan submitted to Checkmarx

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/apis/CourseAPI.kt
CalvinKern: Unfortunate that we'll have to change the interface visibility on all of these api classes, though it doesn't seem like it's an issue to make them more visible (the retrofit example has them as public).

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1018603&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/CoursesApiPactTests.kt
```diff
@@ -0,0 +1,303 @@
...
+        }
+    }.build()
+
+    @Pact(consumer = "mobile")
+    fun getFavoriteCoursesPact(builder: PactDslWithProvider) : RequestResponsePact {
+        return builder
+                .given(MAIN_PROVIDER_STATE)
+
+                .uponReceiving("A request for favorite courses")
+                .path("/api/v1/users/self/favorites/courses")
```
CalvinKern: It would be nice if we didn't have to duplicate these paths from our retrofit interfaces. Seems like maybe it can be referenced via something like this: `createService().favoriteCourses.request().url().encodedPath()`.  If there's not a good way to get the path from our interface, we may want to create another ticket to extract those strings out. Maybe just having constants in that interface or api class that have the paths that can be used by both the interface annotations and this pact builder.

More of a thought than anything, no action is necessarily needed in this PR.

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/CoursesApiPactTests.kt
```diff
@@ -0,0 +1,303 @@
...
+        val getCourseResult = getCourseCall.execute()
+
+        assertNotNull("Expected non-null response body", getCourseResult.body())
+        val course = getCourseResult.body()!!
+
+        assertCoursePopulated(description = "returned course", course = course, fieldConfig = courseWithGradeFieldInfo)
+    }
+    //endregion
+
+    // TODO: Grab course with syllabus?
```
CalvinKern: This would definitely be nice to have, but does feel a little redundant with the course and course_grade tests. Would it be reasonable to combine some things like this that just have different query params? As in, one exhaustive call and one slim call (if there's a benefit to having a slim call)

joehoag: I don't think that we have a "super-call" in our production code that grabs both grades and syllabus.  But it is an interesting idea to test just one call per API, with that call just basically grabbing every available piece of data.  A few potential drawbacks to that:
(1) Such a call usually doesn't exist -- we would have to add it to the existing API(s) and just use it during testing.
(2) We wouldn't be testing the existing calls -- we would just be testing the "super-call".  So it could be the case that one of the calls that we actually use gets broken, but the super-call still passes.

But "super-call" would be an interesting compromise approach, if we decide against exhaustive testing.  (And I don't think that exhaustive testing is a realistic goal.)  We would just need to write one test for each of ~50 APIs.

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/CoursesApiPactTests.kt
joehoag: Good idea, if it works.  I'll give it a try.

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/CoursesApiPactTests.kt
joehoag: And regarding redundant tests: I don't think that it's *too* expensive in terms of execution time to have a redundant or mostly-redundant test.  But it would require coding/maintenance on our part that would be undesirable if not a lot of new stuff was being tested.  

Of course, defining the difference between "mostly-redundant" and "sufficiently non-redundant" is an inexact science.

CalvinKern: Ah, that's a fair point that we don't have a call like that in our existing APIs. I kind of like the idea of a "super-call", but that would increase the maintenance of our Pact tests since they wouldn't be able to tie into a used API call (so we'd need to make sure we update that "super-call" if we need to make modifications to a call we're using in production).

I do also think that exhaustive testing would be very tedious, so not ideal, but it does align more with what our production code is using.

¯\_(ツ)_/¯

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/apis/CourseAPI.kt
joehoag: Odd that I didn't have to remove the "internal" from the UsersInterface.  I'll check once more to make sure that this is necessary.

CalvinKern:
>I personally dislike having the XXX part be in the middle of names.
>> ...if we have some other standard/guideline/unwritten rule...

Nope, we don't have anything about naming like this. So if you prefer keeping it like you have I think it's fine. The names are plenty descriptive too. I just like being able to type it easier by having the same words together rather than split up. Not a big deal either way, so feel free to keep it as is.

instructure-cx:
Scan submitted to Checkmarx

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/apis/CourseAPI.kt
joehoag: See the latest commit -- I've undone this change.

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/CoursesApiPactTests.kt
joehoag: Working on one more path-related commit...

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1018800&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

instructure-cx:
Scan submitted to Checkmarx

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/CoursesApiPactTests.kt
joehoag: OK, here's what I did (see latest commit): I kept the paths and query params hard-coded, because I think that the contract should be rigid, and not changeable just by tweaking API paths/query params in the code.  But I added checks in each test to verify that the path/query params we were expecting were the same as the ones that the call actually generated.

That info *might* have already been partially or fully checked already by Pact.  But this extra logic will make sure.

CalvinKern: Oh, I like that. It's nice to guarantee that we'll be testing the same endpoints as the app. Nice work!

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1018802&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/CoursesApiPactTests.kt
Blackchai: Yeah, this is a tough one. I think my vote is to aim for the most highly used apis, and stick with the api calls we make with the apps. As nice as a super call sounds, idk if I like the idea of our pact tests using an api call that our apps don't "actually" use.

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/EnrollmentsApiPactTests.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.instructure.canvasapi2.apis.EnrollmentAPI
+import com.instructure.canvasapi2.pact.canvas.logic.PactEnrollmentFieldConfig
+import com.instructure.canvasapi2.pact.canvas.logic.assertEnrollmentPopulated
+import com.instructure.canvasapi2.pact.canvas.logic.populateEnrollmentFields
+import io.pactfoundation.consumer.dsl.LambdaDsl
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Test
+
+class EnrollmentsApiPactTests : ApiPactTestBase() {
```
Blackchai: I know we are testing enrollments directly, but it might also make sense to hit a course endpoint with include enrollments.

Blackchai: Especially since we rely on the enrollment object for the course grades.

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/CoursesApiPactTests.kt
Blackchai: Yeah that's awesome! :+1: Best of both worlds.

joehoag:
Thanks for the reviews!  Hopefully future client-side Pact stuff will be a little simpler now that we've decided on some code-factoring and "API surface area compromise" strategies.

PR #483, opened 2020-01-15
[Student][Tests][MBL-13802] : Stabilized PickerSubmissionUploadInteractionTest.testSubmit

joehoag:
These changes seemed to make a difference locally (where I was always failing the first attempt but passing the retry.  ???)  And I suspect that they will address the other failure modes I was seeing on FTL as well.

(And, to follow up on a Slack thread, I don't think that notifications are interfering with the test.)

### DIFF
--- a/automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/CanvasTest.kt
+++ b/automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/CanvasTest.kt
@@ -63,6 +63,7 @@ abstract class CanvasTest : InstructureTest(BuildConfig.GLOBAL_DITTO_MODE) {
         // Let's set ourselves up to log information about our retries.
         ScreenshotTestRule.registerRetryHandler( handler = { error, testMethod, testClass ->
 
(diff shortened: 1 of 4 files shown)


automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/CanvasTest.kt
```diff
@@ -63,6 +63,7 @@ abstract class CanvasTest : InstructureTest(BuildConfig.GLOBAL_DITTO_MODE) {
         // Let's set ourselves up to log information about our retries.
         ScreenshotTestRule.registerRetryHandler( handler = { error, testMethod, testClass ->
 
+            Log.d("TEST RETRY", "testMethod: $testMethod, error=$error, stacktrace=${error.stackTrace.joinToString("\n")} cause=${error.cause}")
```
joehoag: I figured that I would just leave this in, as it does provide valuable info (your test retried, and here's why).

CalvinKern: This is actually a really good idea. Nice!

instructure-cx:
Scan submitted to Checkmarx

PR #511, opened 2020-01-28
[Student] A small change to displaying total grades for MGP

Blackchai:
Okay, so going to add some context around this small change.

It was noticed that we were hiding final grades for all periods when outside of the grading period start/end date range. This was happening specifically as a result of our function call that checks for "hasActiveGradingPeriods()". When we combine the boolean result of hasActiveGradingPeriods() w/ isTotalsForAllGradingPeriodsEnabled to determine if the course grade is locked, we run into problems. The problem here is that we (really I) were using hasActiveGradingPeriods() to determine if we should be viewing the "All Grading Periods" period, and forgot about the case of a user attempting to view individual grading periods after their term was over. In order to correct this within the current function for getting a course grade, we would need to make an extra api call to get all grading periods and check if we are outside the date range of all grading periods. This seems unreasonable, especially for such a specific case. As such, I opted to make a new function specific to this situation, that should only be used when viewing a specific grading period.

While this is not an idea solution, we have made a request to the folks working on the graphql endpoints for grades to simplify this for us. Hopefully, some day, we'll be able to have a "display grade" string that we can just display, rather than duplicate all of this logic on the client side.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/adapter/GradesListRecyclerAdapter.kt
+++ b/apps/student/src/main/java/com/instructure/student/adapter/GradesListRecyclerAdapter.kt
@@ -246,7 +246,7 @@ open class GradesListRecyclerAdapter(
                     // Fetch the enrollments associated with the selected gradingPeriodID, these will contain the
                     // correct grade for the period
                     val enrollments = awaitApi<List<Enrollment>>{CourseManager.getUserEnrollmentsForGradingPeriod(canvasContext!!.id, ApiPrefs.user!!.id, gradingPeriodID, it, true)}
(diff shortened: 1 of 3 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1020090&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

PR #539, opened 2020-02-07
[Flutter-Parent] Fix crash when switching users

CalvinKern:
No need to obtain the selected student notifier via locator, since it’s always available through provider. This was causing a crash as provider was disposing of the notifier, and the locator was trying to reuse the same one.
Also reset the selected student to 0 so that the color theme resets on logouts. We'll probably want to persist this later in the login object.

To test:
Currently on master, switching users will show a crash screen once the new user is loaded (after the splash screen). This should no longer happen.


instructure-cx:
Scan submitted to Checkmarx

PR #540, opened 2020-02-07
[Flutter Parent] Clear nav backstack when switching users

JordanMarshall:
Previously the nav backstack was not being cleared after switching logins, so the user could press the back button and it would return to the login landing page instead of closing the app.


instructure-cx:
Scan submitted to Checkmarx

PR #541, opened 2020-02-07
[Flutter Parent] Find School or District -> Find School

StephenBrough:
Dave requested we change 'Find School or District' to just 'Find School'


instructure-cx:
Scan submitted to Checkmarx

PR #542, opened 2020-02-07
[Flutter Parent] Domain search screen design tweaks

StephenBrough:


instructure-cx:
Scan submitted to Checkmarx

apps/flutter_parent/lib/screens/domain_search/domain_search_screen.dart
```diff
@@ -103,10 +104,21 @@ class _DomainSearchScreenState extends State<DomainSearchScreen> {
           ),
           bottom: ParentTheme.of(context).appBarDivider(shadowInLightMode: false),
           actions: <Widget>[
-            FlatButton(
-              child: Text(L10n(context).next),
-              textColor: Theme.of(context).accentColor,
-              onPressed: _query.isEmpty ? null : () => _next(context),
+            SizedBox(
+              width: 70,
```
JordanMarshall: Due to i18n concerns we probably can't use a fixed size here. For example, here is the word 'Next' in Maltese:
Screen Shot 2020-02-07 at 11 18 51 AM

StephenBrough: Whoa... I wonder if that fits even without the sized box...

Dave wanted that button near the edge, this seemed like the easiest way to do it. I'll have to dig a bit more, maybe there's a way to get it to expand to the left.

CalvinKern: What about something like wrapping it in a column with the cross axis alignment to end?

StephenBrough: That might help. The problem is the flat button seems to add its own padding on the ends. Setting the padding to 0 doesn't do anything, so there's a somewhat large space on the ends. If it's going to be where Dave wants it, I might have to find something other than a MaterialButton.

CalvinKern: Oof. Yeah, may just need to wrap it in an inkwell then.

StephenBrough: Looks like ButtonTheme+RaisedButton might do the trick, gonna give it a go and push something up here soon.

instructure-cx:
Scan submitted to Checkmarx

apps/flutter_parent/lib/screens/domain_search/domain_search_screen.dart
StephenBrough: Went with `MaterialButton` instead. EZ.

instructure-cx:
Scan submitted to Checkmarx

PR #544, opened 2020-02-07
[Flutter-Parent][MBL-13638] Fix talkback on manage children switches

CalvinKern:
Was reading the wrong value on toggle.


instructure-cx:
Scan submitted to Checkmarx

PR #546, opened 2020-02-07
[Flutter Parent] Set up release signing config

JordanMarshall:


instructure-cx:
Scan submitted to Checkmarx

CalvinKern:
There's also an 'Android Sign' step we could add instead (just have to unfilter Flutter when searching in our workflow). This may be easier to maintain long term though.

PR #548, opened 2020-02-10
[Flutter Parent] Bump drawer text size to be consistent with screens

CalvinKern:
Trying to be consistent with sizes between our normal screens and the nav drawer. By default, flutter makes drawer content a smaller size.


instructure-cx:
Scan submitted to Checkmarx

PR #553, opened 2020-02-10
[Flutter-Parent] Moar grade fixes

CalvinKern:
Now shows current grading period by default when going into course details.
Will default to showing the course enrollments current_grade for All Grading Periods if we can’t get the enrollment from the individual enrollment endpoint. i.e., for legacy parents (didn’t pair via pairing code) we can’t get grades by the enrollment endpoint, we can only see the course enrollment which has current_grade (all grading periods) and the current_period_computed_current_grade (current period).

### DIFF
--- a/apps/flutter_parent/.gitignore
+++ b/apps/flutter_parent/.gitignore
@@ -1,4 +1,5 @@
 # Private data
+key.properties
 fabric.properties
 google-services.json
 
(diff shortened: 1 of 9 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
Scan submitted to Checkmarx

PR #560, opened 2020-02-12
[Flutter-Parent][MBL-13509] Help Dialog

CalvinKern:
Also adds a static convenience method to TestApp to enable testing widgets that need a build context (like dialogs).

Main pain point here, I had to include a second intent library to support emails. The url_launcher can start an email activity, but only the default one and it can’t handle a multiline body. There’s also the android_intent library we are using for showing files from inbox, this unfortunately has a bug where we emails aren’t getting set up properly on the actual intent, therefore the email app can’t read the ‘to’ email we’re trying to send to.

Unfortunately, the intent library I found handles emails correctly, but does not support setting both data and a type, so it does not work for the original intent inclusion of viewing a file from a url (just passing the url will usually only prompt for a browser to open, rather than a media viewing app). This means that both intent libraries are required as of right now. I plan on creating a pull request for the originally included android_intent library to fix email intents. That plugin seems better for long term use as it’s a first party library maintained by google, so ideally we will be able to only use that one.

### DIFF
--- a/apps/flutter_parent/lib/l10n/app_localizations.dart
+++ b/apps/flutter_parent/lib/l10n/app_localizations.dart
@@ -160,9 +160,6 @@ class AppLocalizations {
 
   String get help => Intl.message('Help', desc: 'Label text for the help nav drawer button');
 
-  String get shareFeedback =>
-      Intl.message('Send us feedback', desc: 'Label text for the nav drawer button to share feedback');
-
   String get logOut => Intl.message('Log Out', desc: 'Label text for the Log Out nav drawer button');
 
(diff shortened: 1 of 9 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1021976&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

PR #569, opened 2020-02-14
[Flutter-Parent] Add more logs and info for crashes to help debugging

CalvinKern:
You can test by throwing an exception somewhere in the code, running in release mode, and making the crash happen.

Or you can view what it looks like when this is done: https://console.firebase.google.com/project/automatic-rite-88319/crashlytics/app/android:com.instructure.parentapp/issues/c3c4934e105770e3ee0ee85bcfd20db0

### DIFF
--- a/apps/flutter_parent/lib/utils/crash_utils.dart
+++ b/apps/flutter_parent/lib/utils/crash_utils.dart
@@ -46,12 +46,14 @@ class CrashUtils {
     debugPrintStack(stackTrace: stacktrace);
 
     if (kReleaseMode) {
-      // Report to crashlytics
+      // Set any user info that will help to debug the issue
       await Future.wait([
         FlutterCrashlytics().setInfo('domain', ApiPrefs.getDomain()),
         FlutterCrashlytics().setInfo('user_id', ApiPrefs.getUser()?.id),
-        FlutterCrashlytics().reportCrash(exception, stacktrace, forceCrash: false),
       ]);
+
+      // Report to crashlytics
+      await FlutterCrashlytics().reportCrash(exception, stacktrace, forceCrash: false);
     }
   }
 }
--- a/apps/flutter_parent/lib/utils/crash_utils.dart
+++ b/apps/flutter_parent/lib/utils/crash_utils.dart
@@ -15,6 +15,7 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_crashlytics/flutter_crashlytics.dart';
+import 'package:flutter_parent/network/utils/api_prefs.dart';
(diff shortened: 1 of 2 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1022287&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

apps/flutter_parent/lib/utils/crash_utils.dart
```diff
@@ -46,7 +47,11 @@ class CrashUtils {
 
     if (kReleaseMode) {
       // Report to crashlytics
-      await FlutterCrashlytics().reportCrash(exception, stacktrace, forceCrash: false);
+      await Future.wait([
+        FlutterCrashlytics().setInfo('domain', ApiPrefs.getDomain()),
+        FlutterCrashlytics().setInfo('user_id', ApiPrefs.getUser()?.id),
+        FlutterCrashlytics().reportCrash(exception, stacktrace, forceCrash: false),
```
JordanMarshall: Do you think there's any risk of a race condition here, where the report payload is constructed before the user or domain are set?

CalvinKern: Hmm, maybe? Even worse is that these pieces of data won't be set on people that haven't been able to complete login.
The only race condition I think could cause real issues here is if ApiPrefs hasn't been initialized yet, but that's the first thing we do in main (besides running zoned which catches errors and also ensuring flutter binding which shouldn't error out).

We could optionally set these pieces of info only if they aren't null. Do you think that would be enough to mitigate any race conditions that could occur?

JordanMarshall: My understanding is that Future.wait does not guarantee sequential execution of the futures passed to it, so I'm just curious if some some edge condition would allow `reportCrash` to complete before `setInfo` completes.

CalvinKern: ooooooooooooh, I see. That would be bad. I'll move those out so they finish first, nice catch!

instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1022315&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

PR #579, opened 2020-02-19
[Student][Pact][MBL-13982]: Tweaks to allow publishing our pact contract(s) from Bitrise

joehoag:
I also changed our consumer name from "mobile" to "android" to allow for separate ios contracts.


instructure-cx:
Scan submitted to Checkmarx

PR #584, opened 2020-02-19
[Flutter-Parent][all] Add flutter strings to export/import flow

CalvinKern:
This is a little complicated to incorporate into our existing flows.

1. The sync we do for the sources deletes all files on S3 that don’t exist locally. This means we have to upload flutter and native translation sources at the same time. Otherwise, we can’t delete on sync, which could result in stale translation files (for no longer used projects/modules).

2. In order to minimize the impact on the original translation workflow, this first requires the ‘string-extract’ step in bitrise on our flutter-parent app. This way the s3 export jobs don’t need to have flutter installed to be able to do the extraction. The PR generated by that workflow must be merged in before the automated export workflow will use any of the new strings.

3. Since we import manually only, and not very often, the import workflow will update to install flutter so the final linking between translator arb files and the generated messages for flutter is done in one go. Otherwise, another step would be required to then link the translator files into flutter via another bitrise task which would also have another PR. This will increase the overall time of importing translations, but should be low impact since it’s done infrequently. The major downside to this is we now have to manage flutter version in two bitrise projects, alternatively we'd have to do 2 pull requests every translation import.


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1022842&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

PR #618, opened 2020-03-10
[Flutter-Parent][MBL-14091] Domain search support small screens

CalvinKern:
To test:
See ticket for what small screens look like. Should test with a large screen as well to make sure it works for the majority of devices still.

### DIFF
--- a/apps/flutter_parent/lib/screens/domain_search/domain_search_screen.dart
+++ b/apps/flutter_parent/lib/screens/domain_search/domain_search_screen.dart
@@ -97,6 +97,7 @@ class _DomainSearchScreenState extends State<DomainSearchScreen> {
   Widget build(BuildContext context) {
     return DefaultParentTheme(
       builder: (context) => Scaffold(
+        resizeToAvoidBottomInset: false,
         appBar: AppBar(
           title: Text(
             L10n(context).findSchool,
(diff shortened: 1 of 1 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1025204&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

JordanMarshall:
Hmmm, tricky. One of the original design constraints was that the 'How do I find my school or district?' button always be visible, i.e. not in a scrollable container. Unfortunately this change seems to break that (as the original author this is partially my bad; I should have written a test for that). Not sure what our options are here. The experience already works well for the vast majority of devices and only seems be an issue on small screens in landscape with the keyboard open.

CalvinKern:
> button always be visible, i.e. not in a scrollable container

I do remember that being a constraint, now that you mention it. I'll revisit this and see if I can at least make that happen when we have multiple institutions. Definitely won't be able to force it to be shown at the same time as the keyboard being open in landscape on small devices, but I'll see if there's anything else we can do.

JordanMarshall:
I guess this brings up the question about how much time we should spend supporting landscape-with-keyboard scenarios. Part of the problem is that flutter doesn't support fullscreen IME mode like we see in student and teacher, which avoids this problem entirely:


Unfortunately it looks like there are currently no plans to support it either (see [this issue](https://github.com/flutter/engine/pull/3131)). That being the case, since the users are able to dismiss the keyboard themselves, imo we should only need to ensure that things work correctly in landscape _without_ the keyboard showing, and if the keyboard _is_ showing we just need to ensure that the text input is visible.

CalvinKern:
That's a very fair point. Users can dismiss the keyboard to see the entire screen (we may just have to account for that in tests as well). I'll talk with Joe to make sure that will unblock this. 

I'll focus my attention on getting that 'cant find school' button to always be at the bottom of the screen (and not scroll) while removing any widget errors from overflowing like the ticket states (which probably can be supported, but we'll see)

instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1025538&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

CalvinKern:
Closing as this isn't currently failing any of our widget tests. Can look into this again if it becomes a problem.

PR #631, opened 2020-03-16
[Flutter-Parent][MBL-14102] Simplify debug flag

CalvinKern:
All release/debug checks now key off of `ApiPrefs.isDebug`. This is to simplify logic for debug related features to one place, and making it easy to enable debug features in a release build.

Feel free to request any other changes/features with this. Kept is very simple so we could modify as needed, but would be nice to have a way to change this at runtime (without leaking into actual release builds)

### DIFF
--- a/apps/flutter_parent/lib/network/utils/api_prefs.dart
+++ b/apps/flutter_parent/lib/network/utils/api_prefs.dart
@@ -15,6 +15,7 @@
 import 'dart:convert';
 import 'dart:ui' as ui;
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_parent/models/login.dart';
 import 'package:flutter_parent/models/serializers.dart';
@@ -38,6 +39,10 @@ class ApiPrefs {
   static SharedPreferences _prefs;
   static PackageInfo _packageInfo;
   static Login _currentLogin;
+
+  // A static bool rather than a persisted pref, change this as desired during release testing
+  static bool _debugMode = kDebugMode;
+  static bool get isDebug => !kReleaseMode || _debugMode;
 
   static Future<void> init() async {
     if (_prefs == null) _prefs = await SharedPreferences.getInstance();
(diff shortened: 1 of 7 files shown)


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1025871&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

JordanMarshall:
Two thoughts: 
 - Do you think we should move the flag to a location where we can't accidentally commit it to version control? Like the local.properties file, or maybe with a setup similar to snickerdoodles? It's probably not a huge issue since we're pretty good about catching stuff like this in code reviews.
 - Should we have this flag be mutually exclusive from the compile mode? In other words, should we have a isDev/isProd flag that is unrelated to the kDebugMode/kProfileMode/kReleaseMode flags?

CalvinKern:
> Two thoughts:

I really like the idea of putting it into a non version control file, probably more like we have snickerdoodles, that way it's already set up and shouldn't have further changes tracked. One less thing to worry about in code reviews.

I also think we can make it exclusive from the compile mode, but it really depends on the usage for this flag. If it's for showing things that aid in debugging (currently only the settings relies on debug flag for when to show the theme viewer, this can extend to snickers with this change though) then we may still want to have the default for this untracked file be set to kDebugMode. This could further expand later to have dev menus visible or extra logging (which we may just want to have multiple types of flags? for example, the manual debug flag in dio config could live in this file now but should probably be different from the base debug flag).

For now, I'll remove it from referencing kReleaseMode, and we'll keep things that need release behind it (currently just crashlytics). Then I'll setup this file and default the base debug flag to kDebugMode.

instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1026352&projectid=20)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

PR #642, opened 2020-03-19
[Student][MBL-14124] Show active conferences on dashboard

JordanMarshall:
~Waiting on [MBL-14123](https://instructure.atlassian.net/browse/MBL-14123) (new conferences [endpoint](https://gerrit.instructure.com/c/canvas-lms/+/230542)) to hit prod on April 13 before merging, but this can be tested right now on beta.~ Now available on prod

To test, create and start one or more conferences via the web. In the app, these live conferences should display on the dashboard above the user's courses. Tapping a conference should attempt to join it in a chrome custom tab. Tapping the dismiss (X) button should hide the conference (this preference is local and will be reset on logout / user change). Note: it is intended that the API will only return live conferences started in the past day; older conferences will not be shown on the dashboard.


StephenBrough:
Beautiful.

PR #643, opened 2020-03-19
[Student][Teacher][MBL-14122] Support opening LTI links that are intents

CalvinKern:
To test:
Use my sandbox and go to course 1.
* Google Meet -> Opens our LTIWebView momentarily, then goes straight to the play store page (or opens the app if already downloaded)
* Google Meet (new tab) -> Opens our LTIWebView, once you click "open in new tab" it behaves like "Google Meet"
* Goog intent -> Opens our LTIWebView, once you click "open in new tab" it loads googles search page as a fallback for a non-supported intent.


CalvinKern:
@enov This is the pull request that fixes the issue you reported. We'll be pushing an update out in the next week or two most likely, so watch for Student version 6.6.7!

enov:
Awesome! Thank you!!

PR #652, opened 2020-03-23
[Parent][MBL-14007] Add analytics

CalvinKern:
Also merged Logger class into analytics to be more concise. A lot of test consolidation too.

To test:
You can set your device as a debug device in firebase so you can see analytics coming in faster. It took me a couple tries to get it, but I followed the steps here: https://firebase.google.com/docs/analytics/debugview
You should see screen view events (the firebase_screen param is where our route name is shown)
Also should contain the login event, logout/switch users, adding student, help on the login screen, and refresh (not expiring yet, so won't happen).


apps/flutter_parent/lib/network/utils/analytics.dart
```diff
@@ -11,12 +11,86 @@
...
+import 'package:flutter/foundation.dart';
+import 'package:flutter_crashlytics/flutter_crashlytics.dart';
+import 'package:flutter_parent/utils/debug_flags.dart';
+
+/// Event names
+/// The naming scheme for the majority of these is found in a google doc so that we can be consistent
+/// across the platforms.
 class AnalyticsEventConstants {
-  static String get REFRESH_TOKEN => 'refresh_token';
-  static String get FOREVER_TOKEN => 'forever_token';
```
joehoag: What happened to these guys?  Did they go somewhere else?

apps/flutter_parent/lib/network/utils/analytics.dart
```diff
@@ -11,12 +11,86 @@
...
+  /// Logs a message to crashlytics to help when looking over crash logs (only in release mode).
+  /// If isDebug, it will also print to the console
+  void logMessage(String message) {
+    if (kReleaseMode) {
+      FlutterCrashlytics().log(message);
+    }
+
+    if (DebugFlags.isDebug) {
+      print(message);
+    }
```
joehoag: Does kReleaseMode = !DebugFlags.isDebug?

apps/flutter_parent/lib/network/utils/analytics.dart
CalvinKern: I took them out as it isn't valuable for us in Parent app. These analytics in the other apps are to distinguish between people logging in with a refresh_token vs a forever_token. In flutter parent, everyone will be on a refresh_token so no need to analyze.

apps/flutter_parent/lib/network/utils/analytics.dart
CalvinKern: Not always. DebugFlags can be set independent of release mode. By default, it is the same as kDebugMode, but force setting it to true (without committing the change) will allow these debug prints to happen even in a release mode. This will let us have more information while debugging issues that happen in release.

CalvinKern:
> taking advantage of tearDown() and tearDownAll()

I think the tests kind of need an overhaul to clean up. We have a few different patterns that we've played with and some things are definitely a little old and crufty. I wanted to make a ticket to come back and clean up some of those after we get this app released, so we could take a look at adding some of these utilities like tearDown to keep our tests more robust for order (which I also want us to find a way to get a randomized order when running tests eventually too).

PR #658, opened 2020-03-24
[Flutter-Parent] Separate out generating flutter translation hooks

CalvinKern:
Moved the flutter generation out to it's own script and workflow in bitrise. There were also a few different problems with localization that have now been fixed.

Also fixes:
* custom language packs
* rebuilding the app when a user has a different locale in canvas, both when opening the app and when logging out
* app locale no longer resets to the device locale if the app gets backgrounded
* killing the illegally formatted translation files

To test:
* Note: There are currently no successful languages imported into the app right now. (besides Canadian english and the dummy ukhe I am including, see me for test creds on the ukhe stuff, though if you run the local-import and local-flutter-generate-hooks you'll have all languages)
* Dates do show up translated into their language, so arabic, or japanese or something like that will have dates show as those locales.

Should switch languages when users have a different locale in canvas.
Should switch languages when users have a different locale on the device (even while the app is open in the backgrounded) (only if "system language" is specified in their canvas profile)

Lastly, run `ruby translations/local-import.rb` from root and verify there are no problems for the native project, while using names that are acceptable for Flutter (without a + or a prepended 'b'). After that you can run `ruby translations/local-flutter-generate-hooks.rb` which generates the dart hooks for the translated arb files from the previous step. Then the parent app should run and support all the languages.


apps/flutter_parent/lib/network/utils/api_prefs.dart
```diff
@@ -178,7 +180,9 @@ class ApiPrefs {
     } else {
       return Locale.fromSubtags(
         languageCode: localeParts.first,
-        scriptCode: userLocale[1],
+        scriptCode: userLocale[1].length < 5
```
Blackchai: I feel like a comment here with an example locale string could be helpful :)

PR #659, opened 2020-03-25
[Student][6.7 hotfix] Fix for conferences group crash

Blackchai:
Duplicating all of these tests for the group case felt a little overkill, open to suggestions. Only the render case is relevant, since the group bug was in:

val canvasContext by ParcelableArg<CanvasContext>(key = Const.CANVAS_CONTEXT).

However, adding group cases to the future E2E test should cap this off.


JordanMarshall:
> Duplicating all of these tests for the group case felt a little overkill, open to suggestions

Probably overkill since these are render tests and there is no behavioral or visual difference between groups and courses, so maybe just have one test per file that confirms we don't crash when using a group instead of a course. (Plus, I don't think anyone wants to maintain the dupes)

Blackchai:
> > Duplicating all of these tests for the group case felt a little overkill, open to suggestions
> 
> Probably overkill since these are render tests and there is no behavioral or visual difference between groups and courses, so maybe just have one test per file that confirms we don't crash when using a group instead of a course. (Plus, I don't think anyone wants to maintain the dupes)

Yeah that's a _way_ saner approach. I'll make that change.

PR #660, opened 2020-03-25
[Student] login QR fix for tests

Blackchai:


libs/login-api-2/src/main/java/com/instructure/loginapi/login/activities/BaseLoginLandingPageActivity.kt
```diff
@@ -39,6 +39,7 @@ import com.google.gson.Gson
 import com.instructure.canvasapi2.apis.ErrorReportAPI
 import com.instructure.canvasapi2.models.ErrorReportPreFill
 import com.instructure.canvasapi2.utils.*
+import com.instructure.loginapi.login.BuildConfig
```
joehoag: Unnecessary for now?

apps/student/src/androidTest/java/com/instructure/student/ui/pages/LoginLandingPage.kt
```diff
@@ -17,6 +17,8 @@
 package com.instructure.student.ui.pages
```
joehoag: The changes in this file seem unnecessary for now.

Blackchai: yup yup! Just missed that assert = true

libs/login-api-2/src/main/java/com/instructure/loginapi/login/activities/BaseLoginLandingPageActivity.kt
Blackchai: Roger!

PR #664, opened 2020-03-27
[Student][Teacher][MBL-14145]: Now emitting build_type and os_version user properties

joehoag:
From student and teacher.  We'll need another ticket for parent.

I also (1) Added a "remote config" menu under options in the Teacher app, (2) Cleaned up the RemoteConfigParamsFragment to work correctly.

### DIFF
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
@@ -17,6 +17,10 @@
 package com.instructure.pandautils.fragments
 
 import android.os.Bundle
+import android.text.Editable
+import android.text.TextWatcher
+import android.util.Log
+import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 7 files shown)


libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
```diff
@@ -61,9 +65,15 @@ private class RemoteConfigParamAdapter : RecyclerView.Adapter<RecyclerView.ViewH
...
-                RemoteConfigPrefs.putString(param.rc_name, newText)
+            valueEditText.setText(RemoteConfigPrefs.getString(param.rc_name, param.safeValueAsString))
+            valueEditText.setOnKeyListener() {view, id, event ->
+                if(event.action == KeyEvent.ACTION_UP) {
+                    Log.d("RemoteConfigPrefs", "onKeyListener ${param.rc_name} content ${(view as EditText).text}")
+                    RemoteConfigPrefs.putString(param.rc_name, (view as EditText).text.toString())
+                }
+
+                // Returning false allows, for example, the back-button to be handled by other logic.
+                return@setOnKeyListener false
```
joehoag: The former logic exhibited really bizarre behavior when the soft keyboard was enabled and disabled.  Bogus "new" values would get stored, stuff like that.

PR #666, opened 2020-03-30
[Flutter-Parent][MBL-13924] Limit web access

CalvinKern:
To test:
Change the account settings at the institution level to limit web access to parents using the mobile app.
Open the app and go to an assignment with some links to places we don't handle in the app (my sandbox, course1 -> link assignment). Things like modules, files, groups, pages.
When the checkbox is not enabled, the links should open out to the browser (or show the action sheet to choose between apps, always pick browser).
When the checkbox is enabled (to limit web access), then links should open to our simple web view that doesn't allow any further page loads, also displaying an alert that only disappears when clicked.

### DIFF
--- a/apps/flutter_parent/lib/network/api/user_api.dart
+++ b/apps/flutter_parent/lib/network/api/user_api.dart
@@ -18,4 +18,6 @@ import 'package:flutter_parent/network/utils/fetch.dart';
 
 class UserApi {
   Future<User> getSelf() => fetch(canvasDio(forceDeviceLanguage: true).get('users/self/profile'));
+
+  Future<User> getSelfPermissions() => fetch(canvasDio(forceRefresh: true).get('users/self'));
 }
--- a/apps/flutter_parent/lib/network/api/user_api.dart
+++ b/apps/flutter_parent/lib/network/api/user_api.dart
@@ -19,5 +19,6 @@ import 'package:flutter_parent/network/utils/fetch.dart';
 class UserApi {
   Future<User> getSelf() => fetch(canvasDio(forceDeviceLanguage: true).get('users/self/profile'));
 
-  Future<User> getSelfPermissions() => fetch(canvasDio(forceRefresh: true).get('users/self'));
+  Future<UserPermission> getSelfPermissions() =>
+      fetch(canvasDio(forceRefresh: true).get<User>('users/self')).then((user) => user.permissions);
 }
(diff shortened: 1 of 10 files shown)


apps/flutter_parent/lib/router/panda_router.dart
```diff
@@ -268,7 +269,8 @@ class PandaRouter {
 
   static Handler _simpleWebViewHandler = Handler(handlerFunc: (BuildContext context, Map<String, List<String>> params) {
     final url = params[_RouterKeys.url][0];
-    return SimpleWebViewScreen(url, url);
+    final infoText = params[_RouterKeys.infoText]?.elementAt(0);
```
JordanMarshall: The implementation of `simpleWebViewRoute` leads me to believe that future uses of that method which pass in a null `infoText` param would result in `infoText` here being the String `'null'`. Is that something we need to handle?

apps/flutter_parent/lib/router/panda_router.dart
```diff
@@ -371,6 +371,7 @@ class _RouterKeys {
   static final eventId = 'eventId';
   static final quizId = 'quizId';
   static final topicId = 'topicId';
+  static final infoText = 'infoText';
```
JordanMarshall: Are we declaring these alphabetically?

apps/flutter_parent/lib/network/api/user_api.dart
```diff
@@ -18,4 +18,6 @@ import 'package:flutter_parent/network/utils/fetch.dart';
 
 class UserApi {
   Future<User> getSelf() => fetch(canvasDio(forceDeviceLanguage: true).get('users/self/profile'));
+
+  Future<User> getSelfPermissions() => fetch(canvasDio(forceRefresh: true).get('users/self'));
```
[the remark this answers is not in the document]
CalvinKern: I think that would be a reasonable change, makes it more clear it's just for permissions too, rather than used for any user info.

apps/flutter_parent/lib/router/panda_router.dart
CalvinKern: Yup... whoops. Good catch!

apps/flutter_parent/lib/router/panda_router.dart
CalvinKern: We didn't initially, then Stephen went in and fixed everything, then I rebased and forgot to update placement. Thanks for catching that!

PR #668, opened 2020-03-30
[Teacher][MBL-14149] QR Login

Blackchai:
Identical behavior to student QR login. Only going to work on beta. The flag is currently on, so no need to modify anything to test.

### DIFF
--- a/apps/teacher/src/main/java/com/instructure/teacher/activities/LoginActivity.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/activities/LoginActivity.kt
@@ -46,8 +46,8 @@ class LoginActivity : BaseLoginInitActivity() {
         super.finish()
     }
 
-    override fun onCreate(savedInstanceState: Bundle?, persistentState: PersistableBundle?) {
-        super.onCreate(savedInstanceState, persistentState)
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
         // Check to see if we are switching users from an external QR scan
         if(QRLogin.verifySSOLoginUri(intent.data)) {
             startActivity(RouteValidatorActivity.createIntent(this, intent.data!!))
--- a/apps/teacher/src/main/java/com/instructure/teacher/activities/LoginActivity.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/activities/LoginActivity.kt
@@ -48,7 +48,7 @@ class LoginActivity : BaseLoginInitActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         // Check to see if we are switching users from an external QR scan
-        if(QRLogin.verifySSOLoginUri(intent.data)) {
+        if(QRLogin.verifySSOLoginUri(intent.data, true)) {
             startActivity(RouteValidatorActivity.createIntent(this, intent.data!!))
             finish()
         }
--- a/apps/teacher/src/main/java/com/instructure/teacher/activities/LoginActivity.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/activities/LoginActivity.kt
@@ -19,10 +19,13 @@ package com.instructure.teacher.activities
 
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
 import android.os.Bundle
+import android.os.PersistableBundle
 import com.instructure.canvasapi2.models.User
 import com.instructure.canvasapi2.utils.ApiPrefs
 import com.instructure.loginapi.login.activities.BaseLoginInitActivity
+import com.instructure.loginapi.login.util.QRLogin
 import com.instructure.pandautils.services.PushNotificationRegistrationService
 import com.instructure.pandautils.utils.ColorKeeper
 import com.instructure.pandautils.utils.ThemePrefs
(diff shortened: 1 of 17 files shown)


apps/teacher/src/main/java/com/instructure/teacher/activities/RouteValidatorActivity.kt
```diff
@@ -49,45 +59,113 @@ class RouteValidatorActivity : FragmentActivity() {
...
+
+                    val tokenResponse = QRLogin.performSSOLogin(data, this@RouteValidatorActivity, true)
+
+                    // If we have a real user, this is a QR code from a masquerading web user
+                    val intent = if (tokenResponse.realUser != null && tokenResponse.user != null) {
+                        // We need to set the masquerade request to the user (masqueradee), the real user it the admin user currently masquerading
+                        ApiPrefs.isMasqueradingFromQRCode = true
+                        val extras = Bundle()
+                        extras.putLong(Const.QR_CODE_MASQUERADE_ID, tokenResponse.user!!.id)
+                        LoginActivity.createLaunchApplicationMainActivityIntent(this@RouteValidatorActivity, extras)
```
joehoag: Do we want to log a QR event for this case?

libs/login-api-2/src/main/java/com/instructure/loginapi/login/activities/BaseLoginLandingPageActivity.kt
```diff
@@ -127,8 +127,13 @@ abstract class BaseLoginLandingPageActivity : AppCompatActivity(), ErrorReportDi
 
         helpButton.onClickPopupMenu(getString(R.string.requestLoginHelp) to { requestLoginHelp() })
 
+        val remoteConfigParam = when {
+            packageName.contains("teacher") -> RemoteConfigParam.QR_LOGIN_ENABLED_TEACHER
+            else -> RemoteConfigParam.QR_LOGIN_ENABLED
+        }
```
joehoag: I've thought about the logic above.  I think that we should have an explicit case for student, and the default/else case should just throw.  You never know if this class is going to be used for some other app (Polls?  Canvas kids?) and it would be better not to use Student values for that other app.

apps/teacher/src/main/java/com/instructure/teacher/activities/LoginActivity.kt
```diff
@@ -43,6 +45,15 @@ class LoginActivity : BaseLoginInitActivity() {
         super.finish()
     }
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        // Check to see if we are switching users from an external QR scan
+        if(QRLogin.verifySSOLoginUri(intent.data)) {
```
joehoag: Should we set the isTeacher param to true for this call?

libs/login-api-2/src/main/java/com/instructure/loginapi/login/activities/BaseLoginLandingPageActivity.kt
Blackchai: Hmm so if it just throws this would break our current master build of parent app and polls.

Blackchai: But how this works is its paired with the implementation of the abstract BaseLoginLandingPageActivity, so there is the remote config param and the app specific boolean value. So for teacher and parent we'll use a specific remote config flag + their locally implemented flag. All of the other apps have that flag set to false by default, see the loginWithQREnabled() function in this file.

apps/teacher/src/main/java/com/instructure/teacher/activities/LoginActivity.kt
Blackchai: Yup, great catch.

apps/teacher/src/main/java/com/instructure/teacher/activities/RouteValidatorActivity.kt
Blackchai: I'm not currently logging the masquerading case for student either. I suppose we could, but since its not really for our users I don't think it makes sense.

joehoag: 👍

libs/login-api-2/src/main/java/com/instructure/loginapi/login/activities/BaseLoginLandingPageActivity.kt
joehoag: I can see how it ultimately will work for the other apps, because they'll never actually use/reference qrLoginEnabled with the current logic.  It's still a bit messy, IMO, logically speaking; it's not hard to see how future tweaks might break free from your rules and cause some confusion.

Maybe you could only compute/check the remote config param if loginWithQRCodeEnabled() is true:

if(loginWithQRCodeEnabled()) {
  remoteConfigParam = <case statement with throw for default>
  qrLoginEnabled = ...
  if(qrLoginEnabled) {
    qrLogin.setVisible()
    ...
  }
}

Though admittedly the if/else there could get a little messy.

I'll go ahead and approve.  Up to you if you want to clean up the default logic here or not.

joehoag: (Wow, github really butchered my code above.  :-( )

joehoag:
It's unclear to me as to whether your changes might have caused the teacher test failure, or if that's just a flaky test.  Failing test is LoginLandingPageTest. opensCanvasNetworksSignInPage.

libs/login-api-2/src/main/java/com/instructure/loginapi/login/activities/BaseLoginLandingPageActivity.kt
Blackchai: Haha yeah github did. I'll reconsider the area, see if I can improve it a bit, I see your point.

PR #673, opened 2020-03-31
[Flutter Parent] Delete token on logout

JordanMarshall:


apps/flutter_parent/lib/network/utils/api_prefs.dart
```diff
@@ -130,8 +131,13 @@ class ApiPrefs {
...
     _checkInit();
     var logins = getLogins();
-    logins.retainWhere((it) => it.uuid != uuid);
-    await saveLogins(logins);
+    Login login = logins.firstWhere((it) => it.uuid == uuid, orElse: () => null);
+    if (login != null) {
+      // Delete token (fire and forget - no need to await)
+      locator<AuthApi>().deleteToken(login.domain, login.accessToken);
+      logins.retainWhere((it) => it.uuid != uuid);
+      await saveLogins(logins);
```
CalvinKern: Do we want to delete the old accessToken in our authentication interceptor as well when we get a new access token?

JordanMarshall: I would assume that deleting the token via the API also deletes the refresh token, so it's probably not something we want/need to do during a refresh.

CalvinKern: Ah, that sounds reasonable. 👍

PR #679, opened 2020-04-02
[Flutter-Parent][MBL-13598] Login improvements

CalvinKern:
To test:
* Try entering in the domain search www.<sandbox> and make sure it still works (strips out www.)
* Try skipping mobile verify and using the expiring refresh token key in mobile dev (manually input client id/secret from the key)
* Special test: Pull down the code, change the user-agent header in AuthApi to ‘candroid’ -> Try loading canvas.sfu.ca and verify you get redirected (doesn’t allow ‘canvasParent’ normally)

### DIFF
--- a/apps/flutter_parent/lib/screens/web_login/web_login_screen.dart
+++ b/apps/flutter_parent/lib/screens/web_login/web_login_screen.dart
@@ -14,12 +14,15 @@
 
 import 'dart:async';
 
+import 'package:device_info/device_info.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_parent/l10n/app_localizations.dart';
 import 'package:flutter_parent/models/mobile_verify_result.dart';
 import 'package:flutter_parent/network/utils/analytics.dart';
+import 'package:flutter_parent/network/utils/api_prefs.dart';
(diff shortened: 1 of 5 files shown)


apps/flutter_parent/lib/screens/web_login/web_login_screen.dart
```diff
@@ -94,109 +121,143 @@ class WebLoginScreen extends StatelessWidget {
...
+    });
   }
 
   NavigationDecision _navigate(BuildContext context, NavigationRequest request, MobileVerifyResult result) {
-    // TODO: Handle parent specific urls too
-    /*
-     *   private const val PARENT_SUCCESS_URL = "/oauthSuccess"
-     *   private const val PARENT_CANCEL_URL = "/oauth2/deny"
-     *   private const val PARENT_ERROR_URL = "/oauthFailure"
-     *   private const val PARENT_TOKEN_URL = "/canvas/tokenReady"
```
CalvinKern, line 138: None of these are supported now. All of these are related to the old, deprecated AirWolf project that is no longer being used. The `/oauth2/deny` can actually come back if a user 'cancels' an authorization request, though in native we just don't process (not sure why it was a special case then).

PR #680, opened 2020-04-02
[Flutter Parent][MBL-14107] Filter instructure apps from url intents

JordanMarshall:


apps/flutter_parent/lib/utils/url_launcher.dart
```diff
@@ -0,0 +1,43 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:flutter/services.dart';
+
+class UrlLauncher {
+  static const channelName = 'com.instructure.parentapp/url_launcher';
+  static const launchMethod = 'launch';
+  static const canLaunchMethod = 'canLaunch';
```
CalvinKern, line 20: These two aren't being used in this file. Wanna update the method declarations to use these to be consistent?

I also wouldn't be opposed if you did this with the parameter names too (url, excludeInstructure). Then we'd be able to use those in the test like the method names.

apps/flutter_parent/test/router/panda_router_test.dart
```diff
@@ -403,29 +402,13 @@ void main() {
...
-      when(_mockLauncher.launch(
-        url,
-        useSafariVC: anyNamed('useSafariVC'),
-        useWebView: anyNamed('useWebView'),
-        enableJavaScript: anyNamed('enableJavaScript'),
-        enableDomStorage: anyNamed('enableDomStorage'),
-        universalLinksOnly: anyNamed('universalLinksOnly'),
-        headers: anyNamed('headers'),
-      )).thenAnswer((_) => Future.value(true));
+      when(_mockLauncher.launch(url)).thenAnswer((_) => Future.value(true));
```
CalvinKern, line 405: These look so nice now!

apps/flutter_parent/test/screens/help/help_screen_test.dart
```diff
@@ -38,11 +37,13 @@ import '../../utils/test_app.dart';
 void main() {
   final l10n = AppLocalizations();
   HelpScreenInteractor interactor = _MockHelpScreenInteractor();
+  _MockUrlLauncher launcher = _MockUrlLauncher();
```
CalvinKern, line 40: Since launcher and interactor are shared between multiple tests, we should be calling reset with them in set up to avoid any test issues in the future (like when we get randomized test ordering).

The pattern done in other tests is
```
setUp(() {
  reset(launcher);
  reset(interactor);
});
```

PR #682, opened 2020-04-02
[Student][E2E][MBL-14172]: Fix RemoteConfigSettings E2E test issue

joehoag:
The problem was that "qr_login_enabled" is a prefix for "qr_login_enabled_teacher", so when I looked for the former (as a partial string) Espresso got confused.  I fixed the matcher to look for rc_name+':'.

I also consolidated some common logic into a single method to get rid of some code duplication.


joehoag:
> LGTM but can we swap to using kotlin string interpolation? so "${param.rc_name}:" instead

Done.

PR #683, opened 2020-04-03
[Flutter Parent][MBL-13796] Debounce domain picker

StephenBrough:


apps/flutter_parent/lib/utils/debouncer.dart
```diff
@@ -0,0 +1,29 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'dart:async';
+
+class Debouncer {
```
joehoag, line 17: A little documentation would be helpful.  What does the debouncer do, and why do we need it?

apps/flutter_parent/test/utils/debouncer_test.dart
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    // Second one should cancel out the first
+    await Future.delayed(const Duration(milliseconds: 2), () {});
+    debouncer.debounce(() {
+      called = true;
+    });
+
+    // Should be at the threshold for the first debounce after this call
+    await Future.delayed(const Duration(milliseconds: 3), () {});
```
joehoag, line 64: The delays you are using are pretty short -- a system hiccup could cause your test to fail.  You could multiply all of these delays by 10 and still have a pretty fast test.

apps/flutter_parent/lib/screens/domain_search/domain_search_screen.dart
```diff
@@ -162,7 +165,7 @@ class _DomainSearchScreenState extends State<DomainSearchScreen> {
                         },
                       ),
               ),
-              onChanged: (query) => _searchDomains(query),
+              onChanged: (query) => _debouncer.debounce(() => _searchDomains(query)),
```
joehoag, line 168: So the purpose here is to wait for 500ms of inactivity before calling searchDomains()?

apps/flutter_parent/test/utils/debouncer_test.dart
StephenBrough: Probably a good idea. I was afraid of adding too much time to the tests, but it's probably better to give it some leeway so we don't have weird intermittent failures.

apps/flutter_parent/lib/screens/domain_search/domain_search_screen.dart
StephenBrough: Yeah, instead of making a call after every keystroke we debounce it. Not sure if it's a huge hit on the servers, but it'll keep the potential network calls down if a user decides to type in the fully qualified name of their institution.

PR #687, opened 2020-04-03
[Teacher][RC 1.10] Fix for backgrounding after qr login

Blackchai:


apps/teacher/src/main/java/com/instructure/teacher/activities/LoginActivity.kt
```diff
@@ -80,9 +81,7 @@ class LoginActivity : BaseLoginInitActivity() {
         return BuildConfig.IS_TESTING
     }
 
-    override fun userAgent(): String {
-        return "androidTeacher"
-    }
+    override fun userAgent(): String = Const.TEACHER_USER_AGENT
```
CalvinKern, line 84: 💯

Blackchai: :D

PR #698, opened 2020-04-07
[Flutter Parent] Support dark mode in WebViews where possible

JordanMarshall:
This change updates the androidx webkit dependency to add support for dark mode in WebViews. This requires a recent version of the native WebView to test, so it may not work on emulators without Google Play. Tested working on an Android 7 emulator w/ Google Play and an up-to-date physical device running Lollipop. 

Note that the background of WebViews in dark mode is a very dark gray, which doesn't fully match the app's completely black background. Cursory research on this issue did not reveal an easy fix, so we may just have to live with it.


JordanMarshall:
Good point, I could see that being an issue. I'll have to run it by product and see what we should do for those cases. Dunno how I feel about it being a feature flag though since we wouldn't want to degrade the dark mode experience for non-affected users.

PR #703, opened 2020-04-08
[Flutter-Parent] Improve code coverage from tests

CalvinKern:
Takes us form 93.8% to 96.3%

Added some tests for missing line count, excluded more things that we can’t/won’t test.

### DIFF
--- a/apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
+++ b/apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
@@ -36,18 +36,18 @@ import 'create_conversation_interactor.dart';
 
 class CreateConversationScreen extends StatefulWidget {
   CreateConversationScreen(
-    this._courseId,
-    this._studentId,
-    this._subjectTemplate,
-    this._postscript,
-  )   : assert(_courseId != null),
-        assert(_studentId != null),
-        assert(_subjectTemplate != null);
+    this.courseId,
+    this.studentId,
+    this.subjectTemplate,
+    this.postscript,
+  )   : assert(courseId != null),
+        assert(studentId != null),
+        assert(subjectTemplate != null);
 
-  final String _courseId;
-  final String _studentId;
-  final String _subjectTemplate;
-  final String _postscript;
+  final String courseId;
+  final String studentId;
+  final String subjectTemplate;
+  final String postscript;
 
   static final sendKey = Key('sendButton');
   static final attachmentKey = Key('attachmentButton');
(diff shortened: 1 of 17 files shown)


apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
```diff
@@ -186,6 +186,7 @@ class _CreateConversationScreenState extends State<CreateConversationScreen> wit
 
   @override
   Widget build(BuildContext context) {
+    print('build');
```
joehoag, line 189: Is this still necessary?

apps/flutter_parent/test/l10n/app_localizations_delegate_test.dart
```diff
@@ -15,110 +15,131 @@
...
+      expect(AppLocalizations.delegate.isSupported(locale), false);
+    });
   });
 
   /// resolution tests
 
-  test('resolution callback returns locale when supported', () {
-    final fallback = Locale('en', '');
-    final locale = Locale('es', '');
+  group('resoultion callback', () {
```
joehoag, line 78: Misspelled "resolution"

apps/flutter_parent/test/screens/events/event_details_screen_test.dart
```diff
@@ -366,8 +355,7 @@ void main() {
 
       when(interactor.loadReminder(any)).thenAnswer((_) async => null);
 
-      await tester
-          .pumpWidget(_testableWidget(EventDetailsScreen.withEvent(event: event), PlatformConfig(mockPrefs: null)));
```
joehoag, line 370: Was "PlatformConfig(mockPrefs: null)" just not necessary in all of these tests before?

apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_screen.dart
CalvinKern: Nope, good catch

apps/flutter_parent/test/l10n/app_localizations_delegate_test.dart
CalvinKern: Whoops... Thanks!

apps/flutter_parent/test/screens/events/event_details_screen_test.dart
CalvinKern: Since all of these tests were specifying `mockPrefs: null`, I moved it as a default value in the method to clean up all of these calls.

I can't remember why, but it was required to set the current student before the test app was built, so that's done in setup. We specify mockPrefs as null here so that the original mocked prefs from setup don't get overwritten.

PR #704, opened 2020-04-08
[Flutter-Parent] More analytics

CalvinKern:
Added events for toggling dark/hc modes, as well as for creating reminders. Creating different events for each as we only get 25 parameters across all events, so saving those for more important things.

Also created a new `testNonWidgetsWithContext` test wrapper for testing NON widgets that need a BuildContext (there’s interactors that have contexts as parameters, using this test method and the widget tester you can achieve testing those functions).

### DIFF
--- a/apps/flutter_parent/lib/network/utils/analytics.dart
+++ b/apps/flutter_parent/lib/network/utils/analytics.dart
@@ -32,6 +32,14 @@ class AnalyticsEventConstants {
   static String get HELP_DOMAIN_SEARCH => 'help_domain_search';
   static String get ADD_STUDENT_MANAGE_STUDENTS => 'add_student_manage_students';
   static String get ADD_STUDENT_DASHBOARD => 'add_student_dashboard';
+  static String get REMINDER_ASSIGNMENT_CREATE => 'reminder_assignment';
+  static String get REMINDER_EVENT_CREATE => 'reminder_event';
+  static String get DARK_MODE_ON => 'dark_mode_on';
+  static String get DARK_MODE_OFF => 'dark_mode_off';
+  static String get HC_MODE_ON => 'hc_mode_on';
+  static String get HC_MODE_OFF => 'hc_mode_off';
+  static String get DARK_WEB_MODE_ON => 'dark_web_mode_on';
+  static String get DARK_WEB_MODE_OFF => 'dark_web_mode_off';
 }
 
 /// (Copied from canvas-api-2, make sure to stay in sync)
--- a/apps/flutter_parent/lib/network/utils/analytics.dart
+++ b/apps/flutter_parent/lib/network/utils/analytics.dart
@@ -20,26 +20,26 @@ import 'package:flutter_parent/utils/debug_flags.dart';
 /// The naming scheme for the majority of these is found in a google doc so that we can be consistent
 /// across the platforms.
 class AnalyticsEventConstants {
-  static String get LOGIN_FAILURE => 'login_failure';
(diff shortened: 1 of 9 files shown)


apps/flutter_parent/test/utils/accessibility_utils.dart
```diff
@@ -54,7 +54,18 @@ extension A11yExclusionExtension on A11yExclusion {
   }
 }
 
-// A testWidgets() wrapper that runs accessibility checks
+/// A testWidgets() wrapper that does not run accessibility checks. This should not be used by things that are actually
+/// rendering widgets for a meaningful purpose. This should only be used by non widget tests that need access to a
+/// context and a WidgetTester. (example, interactors that need a context as a parameter)
+@isTest
+void testNonWidgetsWithContext(
```
JordanMarshall, line 61: What advantage does this provide over directly using `testWidgets`? Just a different name?

apps/flutter_parent/lib/network/utils/analytics.dart
```diff
@@ -32,6 +32,14 @@ class AnalyticsEventConstants {
   static String get HELP_DOMAIN_SEARCH => 'help_domain_search';
   static String get ADD_STUDENT_MANAGE_STUDENTS => 'add_student_manage_students';
   static String get ADD_STUDENT_DASHBOARD => 'add_student_dashboard';
+  static String get REMINDER_ASSIGNMENT_CREATE => 'reminder_assignment';
+  static String get REMINDER_EVENT_CREATE => 'reminder_event';
```
JordanMarshall, line 36: I used a getter in my recent PR to follow the existing pattern here, but I'm curious why we're using getters for these instead of consts.

apps/flutter_parent/test/utils/notification_util_test.dart
```diff
@@ -141,6 +144,39 @@ void main() {
     expect(details.android.channelId, NotificationUtil.notificationChannelReminders);
     expect(details.android.channelName, AppLocalizations().remindersNotificationChannelName);
     expect(details.android.channelDescription, AppLocalizations().remindersNotificationChannelDescription);
+
+    verify(analytics.logEvent(AnalyticsEventConstants.REMINDER_EVENT_CREATE));
+  });
+
+  test('scheduleReminder calls plugin with expected parameters', () async {
```
JordanMarshall, line 151: Copypasta test name?

apps/flutter_parent/test/utils/accessibility_utils.dart
CalvinKern: Yep! Just wanted a different name so we don't pollute our tests with `testWidgets`. That way we can keep better track and make sure our actual widget tests use the ally tester, and that it doesn't get tempting to use a normal `testWidgets` because "it was used in that other file, etc...".

There's a few places where our non-widgets require a context, so using this should theoretically make testing those possible/easier.

apps/flutter_parent/lib/network/utils/analytics.dart
CalvinKern: Um.... no reason.... I think I just didn't think hard enough about this...

That's actually a really good point. All of these should be const. I'll make that change. Good call!

apps/flutter_parent/test/utils/notification_util_test.dart
CalvinKern: Yep... (sigh)

Dangit, nice catch

PR #705, opened 2020-04-08
[Parent][MBL-14150] Parent QR Login

Blackchai:
Y’all know the drill, same test plan as all others. This time, its flutter time.


apps/flutter_parent/test/screens/login/login_landing_screen_test.dart
```diff
@@ -58,9 +61,13 @@ void main() {
     locator.registerFactory<DomainSearchInteractor>(() => null);
   });
 
-  setUp(() {
+  setUp(() async {
     reset(analytics);
     reset(authApi);
+    RemoteConfigUtils.clean();
```
joehoag, line 67: Given your change to TestApp, I think that you should call RemoteConfigUtils.clean() in tearDown() instead of setUp() in all of these tests that use RemoteConfigUtils.  Otherwise ensuing tests might not be able to set their mockPrefs correctly in TestApp().

apps/flutter_parent/test/utils/test_app.dart
```diff
@@ -175,7 +176,7 @@ Future<void> setupPlatformChannels({PlatformConfig config = const PlatformConfig
   if (config.initDeviceInfo) _initPlatformDeviceInfo();
 
   Future<void> apiPrefsInitFuture;
-  if (config.mockPrefs != null) {
+  if (config.mockPrefs != null && RemoteConfigUtils.isInitialized() == false) {
```
joehoag, line 179: Another potential way to fix this issue would be to add an enableQrTesting boolean field (default to false) to TestApp and the TestApp constructor.  And if it's true, we could manually add in the 'rc_qr_login_enabled_parent' entry right here.

apps/flutter_parent/test/router/panda_router_test.dart
```diff
@@ -373,6 +388,14 @@ void main() {
 
       expect(widget, isA<RouterErrorScreen>());
     });
+
+    test('returns Splash screen for qr login', () {
```
joehoag, line 392: Is this any different from 'qrLogin returns splash screen' above?

apps/flutter_parent/test/screens/login/login_landing_screen_test.dart
Blackchai: setup() gets called before every test runs, so this is effectively the same thing. Only setupAll() gets called before all tests run, rather than every test. I figured it would just be a little simpler to keep it all in one function.

apps/flutter_parent/test/router/panda_router_test.dart
Blackchai: Yeah, the first one directly tests the qr_login handler, this one uses _rootWithUrl() to test the external url handler. Basically internal QR scan vs external QR scan.

apps/flutter_parent/test/utils/test_app.dart
Blackchai: Hmm yeah I was thinking about this more. I wonder if another good option would be to just have an option to pass in the entire remoteConfig to TestApp, and have it set things up. I'll play around with this a bit and then see what you think.

apps/flutter_parent/test/screens/login/login_landing_screen_test.dart
joehoag: But setup() won't be called before the next test module runs.

apps/flutter_parent/test/router/panda_router_test.dart
joehoag: Ah -- ok.

apps/flutter_parent/test/screens/login/login_landing_screen_test.dart
Blackchai: Ahh I see, okay cool, that makes sense. I'm hoping we won't need this at all with the changes I have in mind tho, we'll see 🤞 .

apps/flutter_parent/test/utils/test_app.dart
joehoag: Or you could just pass a field called keepCurrentPrefs which keeps the current prefs values.  Default to false.

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
```diff
@@ -0,0 +1,105 @@
...
+                automaticallyImplyLeading: true,
+                actions: <Widget>[_nextButton(context)],
+              ),
+              body: _body(context),
+            ));
+  }
+
+  Widget _nextButton(BuildContext context) {
+    return MaterialButton(
+      minWidth: 20,
```
CalvinKern, line 50: Shouldn't min width be 48?

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
```diff
@@ -0,0 +1,105 @@
...
+      highlightColor: Colors.transparent,
+      splashColor: Theme.of(context).accentColor.withAlpha(100),
+      textColor: Theme.of(context).accentColor,
+      shape: CircleBorder(side: BorderSide(color: Colors.transparent)),
+      onPressed: () async {
+        var barcodeResult = await locator<QRLoginTutorialScreenInteractor>().scan();
+        if (barcodeResult.isSuccess) {
```
CalvinKern, line 57: Do we want any analytics logged here on success/fail?

apps/flutter_parent/lib/screens/splash/splash_screen.dart
```diff
@@ -69,8 +75,13 @@ class _SplashScreenState extends State<SplashScreen> with SingleTickerProviderSt
...
             } else if (snapshot.hasError) {
-              // On error, proceed without pre-fetched student list
-              _navigate(PandaRouter.dashboard());
+              if(snapshot.error is QRLoginError) {
+                Scaffold.of(context).showSnackBar(SnackBar(content: Text(L10n(context).loginWithQRCodeError)));
+                Navigator.pop(context);
+              } else {
+                // On error, proceed without pre-fetched student list
+                _navigate(PandaRouter.dashboard());
+              }
```
CalvinKern, line 84: Analytics here as well?

apps/flutter_parent/test/utils/test_app.dart
```diff
@@ -175,7 +176,7 @@ Future<void> setupPlatformChannels({PlatformConfig config = const PlatformConfig
   if (config.initDeviceInfo) _initPlatformDeviceInfo();
 
   Future<void> apiPrefsInitFuture;
-  if (config.mockPrefs != null) {
+  if (config.mockPrefs != null && config.clearPrefs) {
```
joehoag, line 179: You *could* make the TestApp() call/constructor more concise by passing the clearPrefs bool as part of the TestApp constructor instead of part of the PlatformConfig constructor.  But it's also fine as as.

apps/flutter_parent/lib/screens/splash/splash_screen_interactor.dart
```diff
@@ -40,6 +57,39 @@ class SplashScreenInteractor {
...
+    Login login = Login((b) => b
+      ..accessToken = tokenResponse.accessToken
+      ..refreshToken = tokenResponse.refreshToken
+      ..domain = mobileVerifyResult.baseUrl
+      ..clientId = mobileVerifyResult.clientId
+      ..clientSecret = mobileVerifyResult.clientSecret
+      ..masqueradeUser = isMasquerading ? tokenResponse.user.toBuilder() : null
+      ..masqueradeDomain = isMasquerading ? mobileVerifyResult.baseUrl : null
+      ..isMasqueradingFromQRCode = isMasquerading
+      ..canMasquerade = isMasquerading
```
CalvinKern, line 84: For `canMasquerade`, I think you would be better off doing `isMasquerading? true : null`
That way, when getData continues, it will perform the proper can masquerade checks for the domain/permissions for that user.

apps/flutter_parent/test/utils/test_app.dart
Blackchai: I think I prefer keeping it with the other platform config values, since its only relevant for the setupPlatformChannels call.

apps/flutter_parent/lib/utils/veneers/barcode_scan_veneer.dart
```diff
@@ -0,0 +1,23 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import 'package:barcode_scan/barcode_scan.dart';
+
+class BarcodeScanVeneer {
+  Future<String> scanBarcode() async {
+    return await BarcodeScanner.scan();
+  }
```
CalvinKern, line 22: This is technically not hurting anything as is, but you don't need tot make this method async and await the result. You can just `return BarcodeScanner.scan();`

apps/flutter_parent/test/utils/test_app.dart
joehoag: 👍

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
```diff
@@ -0,0 +1,105 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return DefaultParentTheme(
+        builder: (context) => Scaffold(
+              key: _scaffoldKey,
+              appBar: AppBar(
+                elevation: 4,
+                title: Text(L10n(context).locateQRCode),
+                automaticallyImplyLeading: true,
+                actions: <Widget>[_nextButton(context)],
```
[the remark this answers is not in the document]
Blackchai: roger that!

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
Blackchai: Tbh, I copied this button from the other login screen that has a NEXT button. I'm not sure.

apps/flutter_parent/lib/utils/veneers/barcode_scan_veneer.dart
Blackchai: oh cool, thanks!

apps/flutter_parent/lib/screens/splash/splash_screen.dart
Blackchai: Damnit, forgot analytics 🤦

apps/flutter_parent/lib/screens/splash/splash_screen_interactor.dart
Blackchai: :+1:

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
CalvinKern: Ah, okay. Not sure but it's passing ally checks so it seems okay.

So if you copied from the other one, you are also aware that this button has a circular shape as opposed to the other one with a square shape? Should we make those both be the same shape? (don't care which one, just care that they're consistent)

apps/flutter_parent/test/screens/dashboard/dashboard_screen_test.dart
```diff
@@ -88,8 +90,15 @@ void main() {
     _locator.registerLazySingleton<Analytics>(() => analyticsMock);
   }
 
-  setUp(() {
+  setUp(() async {
     reset(analyticsMock);
+    await setupPlatformChannels();
+    final mockRemoteConfig = setupMockRemoteConfig(valueSettings: {'qr_login_enabled_parent': 'true'});
+    await RemoteConfigUtils.initializeExplicit(mockRemoteConfig);
```
CalvinKern, line 97: Is it necessary to setup remote config utils on the dashboard; why does it care about qr_login_enabled_parent?

apps/flutter_parent/test/screens/login/login_landing_screen_test.dart
```diff
@@ -276,6 +286,16 @@ void main() {
     // TODO: Remove this back press once DomainSearchScreen is passing accessibility checks
     await tester.pageBack();
   });
+
+  testWidgetsWithAccessibilityChecks('Tapping QR login shows QR Login Tutorial screen', (tester) async {
+    await tester.pumpWidget(TestApp(LoginLandingScreen(), platformConfig: PlatformConfig(clearPrefs: false),));
```
CalvinKern, line 291: Rather than creating a new flag for `clearPrefs` you can just pass in a null for `mockPrefs` inside the PlatformConfig.

I realize that's not super clear. maybe having a flag is better for that to be more explicit?

CalvinKern: I also think it may be worth it to provide a mocked remoteconfig to test app to have it initialize so that it's more built into the flow and we don't have to worry about providing a 'null' mock prefs or a false clear prefs flag. Then TestApp can initialize them both without clearing one or the other out.

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
Blackchai: Yeah, I'll update the other one.

apps/flutter_parent/test/screens/login/qr_login_tutorial_screen_interactor_test.dart
```diff
@@ -0,0 +1,101 @@
...
+import 'package:mockito/mockito.dart';
+
+void main() {
+  final mockScanner = _MockScanner();
+
+  setUp(() {
+    final _locator = GetIt.instance;
+    _locator.reset();
+    _locator.registerLazySingleton<BarcodeScanVeneer>(() => mockScanner);
+  });
```
CalvinKern, line 35: Nope. Reject.

Use `setupTestLocator` instead. You just have to provide a callback to register things to a locator, rather than having to get the instance and call reset manually.

apps/flutter_parent/test/screens/login/qr_login_tutorial_screen_test.dart
```diff
@@ -0,0 +1,98 @@
...
+
+import '../../utils/accessibility_utils.dart';
+import '../../utils/test_app.dart';
+
+void main() {
+  final barcodeResultUrl =
+      'https://${QRUtils.QR_HOST}/canvas/login?${QRUtils.QR_AUTH_CODE}=1234'
+      '&${QRUtils.QR_DOMAIN}=mobiledev.instructure.com';
+  final interactor = _MockInteractor();
+  final mockNav = _MockNav();
```
CalvinKern, line 34: Need a `setUp()` to reset both interactor and mockNav

apps/flutter_parent/test/screens/login/qr_login_tutorial_screen_interactor_test.dart
Blackchai: Roger that.

apps/flutter_parent/test/screens/splash/splash_screen_interactor_test.dart
```diff
@@ -27,14 +31,33 @@ import '../../utils/test_app.dart';
...
+    ..accessToken = '123'
+    ..refreshToken = '123'
+    ..user = CanvasModelTestUtils.mockUser().toBuilder()
+    ..realUser = null
+  );
+
   setupTestLocator((locator) {
     locator.registerFactory<DashboardInteractor>(() => dashboardInteractor);
     locator.registerLazySingleton<AccountsApi>(() => accountsApi);
+    locator.registerLazySingleton<AuthApi>(() => authApi);
```
CalvinKern, line 60: Need to reset AuthApi in set up

apps/flutter_parent/test/utils/remote_config_utils_test.dart
```diff
@@ -100,4 +100,4 @@ _MockRemoteConfig _setupMockRemoteConfig({Map<String, String> valueSettings = nu
   return mockRemoteConfig;
 }
 
-class _MockRemoteConfig extends Mock implements RemoteConfig {}
+class MockRemoteConfig extends Mock implements RemoteConfig {}
```
CalvinKern, line 103: This should stay private so that it's only referenced in the one file. The pattern we've been doing is just redeclaring these at the bottom of any file that needs them.

Something I've been thinking about is just having a test file with all of these mock declarations so that any test file can use them (but we're not depending on a specific file, like depending on remote_config_utils_test for MockRemoteConfig). You could start doing that if you want to... Or just make this one private and add one to your files for now.

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
Blackchai: I'm leaning towards only logged it at the site of the actual SSO call to reduce false negatives.

apps/flutter_parent/test/screens/login/login_landing_screen_test.dart
Blackchai: We chatted about this over DMs, decided to keep it as is due to encrypted prefs changing things up.

joehoag: So TestApp would construct a mocked RemoteConfig if none was provided, or else use the passed-in RemoteConfig?  That could work.  Then you wouldn't have to worry about initializing RemoteConfig in tests that didn't care about its flag values.

PR #707, opened 2020-04-09
[Flutter Parent][MBL-14213] Display message regarding old reminders

JordanMarshall:
Also moves `OldAppMigrations` to the service locator and adds missing tests.

To test:
 - Build/install the native parent app, log in, and set up some reminders
 - Build/install the flutter app using the same signing key so that the app is updated instead of uninstalled
 - If no reminders were set up in the native app, or the date for all reminders are prior to today's date, then _no_ dialog should display on the dashboard.
 - If the date for one or more reminders is today or in the future, a dialog _should_ display on the dashboard.
 - After dismissing the dialog, it should never display again.


apps/flutter_parent/test/utils/old_app_migrations_test.dart
```diff
@@ -0,0 +1,60 @@
...
+
+import 'test_app.dart';
+
+void main() {
+  setUp(() async {
+    await setupPlatformChannels();
+  });
+
+  group('performMigrationIfNecessary', () {
+    // TODO
```
JordanMarshall, line 28: These tests will be added in a separate PR

apps/flutter_parent/test/utils/old_app_migrations_test.dart
```diff
@@ -0,0 +1,107 @@
...
+import 'package:flutter_parent/utils/old_app_migration.dart';
+import 'package:test/test.dart';
+
+import 'canvas_model_utils.dart';
+import 'test_app.dart';
+
+void main() {
+  setUp(() async {
+    await setupPlatformChannels();
+  });
```
CalvinKern, line 30: Minor nitpick, this could be a `setUpAll` instead, so that we don't setup the platform channels multiple times for each test (no shared data, just setting up platform related things that don't even need to be reset).

JordanMarshall: This is being used to reset the ApiPrefs before each test.

> don't setup the platform channels multiple times for each test

As far as I can tell it's only invoked once per test.

CalvinKern: Ah, sorry, my wording was a little weird on that. It currently will run setupPlatformChannels before each test is run, which does reset ApiPrefs like you want. Alternatively, you can be more efficient by only calling to `setupPlatformChannels()` once for the whole file (rather than each test) in `setUpAll`, then if you just need ApiPrefs reset you can still have a `setUp` or `tearDown` that does `ApiPrefs.clean()`.

The way it currently is, we're waiting for the initialization again on every test and it makes the entire test suite take longer (probably only by a few milliseconds, but if this is done in enough tests that's a lot of extra time for nothing).

That's a good call though to note that you need ApiPrefs reset between tests, and we should probably be doing this `ApiPrefs.clean()` in more tests that use ApiPrefs.

JordanMarshall: A quick benchmark (100k iterations) shows an initialization cost of ~0.1ms. So If we initialized this for every single test in the entire suite it will cost us an additional ~0.13 seconds. I'm not worried about it.

CalvinKern: Oh that's very interesting. I'm surprised it's so small, but that's good to know.

Actually, after thinking about this more, any time we are creating a test app it is also initializing platform channels each time test app is inflated. So we're already doing this kind of pattern (setupPlatformChannels() on each test).

This may actually be a safer way to just have all the platform stuff initialized.

Thanks for benchmarking that, I'm actually now leaning towards always putting this `setupPlatformChannels()` in our `setUp()` functions.

PR #711, opened 2020-04-10
[Flutter Parent][MBL-14236] Wire up notifications to router

JordanMarshall:
To test:
 - Set up a reminder for an assignment or calendar event (~1 minute in the future to avoid waiting too long)
 - Wait for the notification to appear
 - Tap the notification and the app should route to the correct assignment/event.
 - Routing should succeed regardless of whether the the app is in the foreground, background, or not started.

### DIFF
--- a/apps/flutter_parent/lib/utils/notification_util.dart
+++ b/apps/flutter_parent/lib/utils/notification_util.dart
@@ -15,28 +15,17 @@
 import 'dart:convert';
 
 import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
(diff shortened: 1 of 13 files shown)


apps/flutter_parent/lib/utils/notification_util.dart
```diff
@@ -88,8 +77,19 @@ class NotificationUtil {
     // Delete reminder from db
     await locator<ReminderDb>().deleteById(reminder.id);
 
-    // Route to event or assignment
-    // TODO: MBL-13819
+    // Create route
+    String route;
+    switch (reminder.type) {
+      case Reminder.TYPE_ASSIGNMENT:
+        route = PandaRouter.assignmentDetails(reminder.courseId, reminder.itemId);
```
Blackchai, line 84: Noice!

PR #715, opened 2020-04-13
[Flutter-Parent] A few more tests and cleanup

CalvinKern:
Coverage goes by .4% to 96.8%

Also migrated a large portion of mock dependencies for utility/non-screen tests to mock_helpers.dart. Plus a bunch of dart analysis clean up.

### DIFF
--- a/apps/flutter_parent/lib/utils/common_widgets/web_view/simple_web_view_screen.dart
+++ b/apps/flutter_parent/lib/utils/common_widgets/web_view/simple_web_view_screen.dart
@@ -12,7 +12,6 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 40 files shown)


apps/flutter_parent/lib/utils/common_widgets/web_view/simple_web_view_screen.dart
```diff
@@ -102,7 +101,7 @@ class _SimpleWebViewScreenState extends State<SimpleWebViewScreen> {
   }
 
   Future<bool> _handleBackPress(BuildContext context) async {
-    if (await _controller?.canGoBack()) {
+    if (await _controller?.canGoBack() == true) {
```
Blackchai, line 104: When looking at the code, this doesn't throw warnings/errors etc. Is this necessary for the null case or something? Just curious

CalvinKern: It actually did throw an error when I tested this case. I'm not sure why there's no warning for this, maybe because of the await?

Blackchai: Interesting, that's good to know! Yeah I'm not sure... good find tho

PR #718, opened 2020-04-14
[Parent][MBL-14237] Course routing for syllabus/frontpage

Blackchai:
To Test:
Click links (internally/externally) to either the course syllabus, or the course front page. This should open the new course detail view with just that specific element in it.


apps/flutter_parent/lib/screens/courses/routing_shell/course_routing_shell_interactor.dart
```diff
@@ -0,0 +1,50 @@
...
+import 'package:flutter_parent/utils/service_locator.dart';
+
+class CourseRoutingShellInteractor {
+  Future<CourseShellData> loadCourseShell(CourseShellType type, String courseId, {bool forceRefresh = false}) async {
+    var course = await _loadCourse(courseId);
+    var frontPage = null;
+
+    if (type == CourseShellType.frontPage) {
+      frontPage = await _loadHomePage(courseId, forceRefresh: forceRefresh);
+    }
```
JordanMarshall, line 31: Would it be worthwhile to add a null check for `course.syllabusBody` if the type is `CourseShellType.syllabus`? It's definitely a corner case but I imagine the syllabus body could be null when routing from a link to a syllabus that has since been removed, resulting in a widget crash down the line.

apps/flutter_parent/lib/screens/courses/routing_shell/course_routing_shell_screen.dart
```diff
@@ -0,0 +1,114 @@
...
+            : CanvasHtml(data.course.syllabusBody));
+  }
+
+  Widget _appBarTitle(String title, String subtitle) {
+    return Column(
+      mainAxisAlignment: MainAxisAlignment.center,
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: <Widget>[
+        Text(title, style: TextStyle(fontSize: 18.0)),
+        Text(subtitle, style: TextStyle(fontSize: 12.0)),
```
JordanMarshall, line 104: The text style for the title is applied by AppBar so you don't need to specify a style here. For the subtitle I think we've been using `Theme.of(context).primaryTextTheme.caption`.

apps/flutter_parent/lib/screens/courses/routing_shell/course_routing_shell_interactor.dart
Blackchai: oh hmm that's a good point, I'll do some testing and handle that case.

apps/flutter_parent/lib/screens/courses/routing_shell/course_routing_shell_screen.dart
Blackchai: :+1:

PR #722, opened 2020-04-14
[Parent] notification init change for routing

Blackchai:
To Test:
Set up a notification reminder, then kill the app.

### DIFF
--- a/apps/flutter_parent/lib/parent_app.dart
+++ b/apps/flutter_parent/lib/parent_app.dart
@@ -12,6 +12,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import 'dart:async';
 import 'dart:ui';
 
 import 'package:flutter/material.dart';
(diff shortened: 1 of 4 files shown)


apps/flutter_parent/lib/parent_app.dart
```diff
@@ -45,6 +46,10 @@ class _ParentAppState extends State<ParentApp> {
   @override
   void initState() {
     super.initState();
+
+    // Init notifications here to ensure it happens after routing/etc
+    NotificationUtil.init();
```
CalvinKern, line 51: Hmm, you really shouldn't do asynchronous things inside of an init state. This could introduce a race condition where the parent app starts building something that requires notificationUtil before notificationUtil has a chance to finish initialization. I don't think there's anything that time sensitive right now in the app, but it could cause us a very bad headache later on (if we expand on notifications or if we continue this pattern in other places).

CalvinKern: In fact, reviewing further, I think the issue here is that NotificationUtil needs ParentApp built before it can safely finish initializing. NotificationUtil will try to push a named route once it sees it has a notification, which relies on ParentApp getting built with our route generator.

Not sure the right way around this...

CalvinKern: Thought of an idea around this, and you could keep the initialization in main. Then, pass a completer into ParentApp when calling run. That might be a way to let the notifications only trigger once the app has been built. So the line in NotificationUtil that handles this would turn from this:
```
WidgetsBinding.instance?.handlePushRoute(route);
```
to this:
```
_appCompleter.future.then((_) => WidgetsBinding.instance?.handlePushRoute(route));
```

We might even be able to have the completer be a context so that we can make use of QuickNav instead of using the (protected) handlePushRoute method.

CalvinKern: Lol, or it might be easier and you just need to move this init call to further down the build tree, once MaterialApp is done building (and routes are hooked in).

Blackchai: lots of DMs to get to this solution, thanks yo!

PR #725, opened 2020-04-15
[Parent][E2E][MBL-14260]: Speculative fix for submission-seeding issue

joehoag:
This root cause of this issue was very hard to pin down, as the issue itself was unpredictably intermittent.  But I think that this delay added between creating an assignment and creating a submission for that assignment will address suspected delayed-task issues to a degree.


apps/flutter_parent/test_driver/flutter_driver_extensions.dart
```diff
@@ -17,11 +17,15 @@ import 'package:flutter_driver/flutter_driver.dart';
 // Extensions on FlutterDriver that allow us to try pull-to-refresh if the initial attempt fails
```
joehoag, line 17: These changes allow the test to refresh the screen if the expected text does not match the actual text.

PR #727, opened 2020-04-15
[Flutter-Parent][MBL-13956] Encrypted prefs

CalvinKern:
Everything under the `plugin` directory is copied from the SharedPreferences flutter plugin. There was a small tweak to use EncryptedSharedPreferences instead, but 99% of the code is the same as the initial plugin.

To test:
Run the app and login (or just run the app if you had previous logins).
Open an instance of android studio, and navigate to the device file explorer
Navigate to `data/data/com.instructure.parentapp/shared_prefs/`
* Verify there is a FlutterSharedPreferences.xml that doesn't contain any user/login data
* Verify there is a FlutterEncrypedSharedPreferences.xml that contains user data that is not readable (it's encrypted)


apps/flutter_parent/test/utils/remote_config_utils_test.dart
```diff
@@ -31,29 +29,24 @@ void main() {
...
-    // Start up with no cached values
-    await setupPlatformChannels();
-
     // Create a mocked RemoteConfig object that will fetch a 'test_string' value
     final mockRemoteConfig = setupMockRemoteConfig(valueSettings: {'test_string': 'fetched value'});
-    await RemoteConfigUtils.initializeExplicit(mockRemoteConfig);
+    await setupPlatformChannels(config: PlatformConfig(initRemoteConfig: mockRemoteConfig));
 
     expect(RemoteConfigUtils.getStringValue(RemoteConfigParams.TEST_STRING), 'fetched value');
   });
```
joehoag, line 52: Did you forget to adapt the two tests below?

CalvinKern: Oops, good catch!

apps/flutter_parent/lib/network/utils/api_prefs.dart
```diff
@@ -34,17 +35,45 @@ import 'dio_config.dart';
...
+    oldPrefs.remove(KEY_HAS_MIGRATED);
+
+    _prefs.setBool(KEY_HAS_CHECKED_OLD_REMINDERS, oldPrefs.getBool(KEY_HAS_CHECKED_OLD_REMINDERS));
+    oldPrefs.remove(KEY_HAS_CHECKED_OLD_REMINDERS);
+
+    _prefs.setString(KEY_CURRENT_LOGIN_UUID, oldPrefs.getString(KEY_CURRENT_LOGIN_UUID));
+    oldPrefs.remove(KEY_CURRENT_LOGIN_UUID);
+
+    _prefs.setString(KEY_CURRENT_STUDENT, oldPrefs.getString(KEY_CURRENT_STUDENT));
+    oldPrefs.remove(KEY_CURRENT_STUDENT);
```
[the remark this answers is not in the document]
CalvinKern: Hmmm, that's a fair question. I'm not sure it buys us anything, but I like the idea of this ending with us in sync with the platform.

I'll push up a quick change.

PR #730, opened 2020-04-16
[Flutter Parent] Allow only instructor recipients if send_messages course permission is not granted

JordanMarshall:
If a user is not granted the `send_messages` permission for a course, Canvas only allows them to send messages to instructors (see this [relevant Canvas commit](https://github.com/instructure/canvas-lms/commit/bcdbfb58e15a35f32dd7c4ff4c3f55c643aabd19#diff-aba01162ec4412dd54a8edde3b05afb4R384)). This change adds that permission check and allows only instructors in the recipient list if the permission is not granted.


apps/flutter_parent/lib/screens/inbox/create_conversation/create_conversation_interactor.dart
```diff
@@ -27,12 +27,21 @@ import '../attachment_utils/attachment_picker.dart';
 class CreateConversationInteractor {
   Future<CreateConversationData> loadData(String courseId, String studentId) async {
     final courseFuture = locator<CourseApi>().getCourse(courseId);
-    final recipients = await locator<InboxApi>().getRecipients(courseId);
```
joehoag, line 30: I don't get it -- what does converting 

recipients = await locator...

to

recipientsFuture = locator...
recipients = await recipientsFuture;

buy you?

JordanMarshall: It allows us to fetch the course permissions and recipients simultaneously, rather than waiting on one before we can fetch the other.

joehoag: 👍

PR #742, opened 2020-04-20
[Flutter-Parent] Fix edge case for grading period header

CalvinKern:
If a course has no assignments with a current grading period, filtering to ‘All Grading Periods’ removes the period header and requires users to back out and in to filter by period again.

To test:
Use Wilsons sandbox and view the course `Better Chat`, ensure you can change the grading period to "All Grading Periods" and then change again without having to reload the screen.


apps/flutter_parent/test/screens/courses/course_grades_screen_test.dart
```diff
@@ -534,7 +535,7 @@ void main() {
     });
   });
 
-  testWidgetsWithAccessibilityChecks('grading period not shown when noot multiple grading periods', (tester) async {
```
Blackchai, line 537: noot doot

CalvinKern: Noot what I meant to type 😜

PR #748, opened 2020-04-21
[Parent] Barcode Scanner 3.0

Blackchai:
Behavior unchanged, just had to modify some of the error handling logic and get those tests passing.

### DIFF
--- a/apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen_interactor.dart
+++ b/apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen_interactor.dart
@@ -48,6 +48,9 @@ class QRLoginTutorialScreenInteractor {
         // Unknown error while scanning
         result = BarcodeScanResult(false, errorType: QRError.invalidQR);
       }
+    } catch (e) {
+      // Just in case
+      result = BarcodeScanResult(false, errorType: QRError.invalidQR);
     }
 
     return result;
--- a/apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen_interactor.dart
+++ b/apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen_interactor.dart
@@ -24,24 +24,30 @@ class QRLoginTutorialScreenInteractor {
   Future<BarcodeScanResult> scan() async {
     BarcodeScanResult result;
     try {
-      String barcodeResult = await locator<BarcodeScanVeneer>().scanBarcode();
-      if (QRUtils.verifySSOLogin(barcodeResult) != null) {
-        result = BarcodeScanResult(true, result: barcodeResult);
-      } else {
-        result = BarcodeScanResult(false, errorType: QRError.invalidQR);
+      ScanResult scanResult = await locator<BarcodeScanVeneer>().scanBarcode();
+      String barcodeResult = scanResult.rawContent;
+      switch (scanResult.type) {
+        case ResultType.Barcode:
+          if (QRUtils.verifySSOLogin(barcodeResult) != null) {
+            result = BarcodeScanResult(true, result: barcodeResult);
+          } else {
+            result = BarcodeScanResult(false, errorType: QRError.invalidQR);
+          }
+          break;
+        case ResultType.Error:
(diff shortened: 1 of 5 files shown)


apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen_interactor.dart
```diff
@@ -24,24 +24,30 @@ class QRLoginTutorialScreenInteractor {
...
+        case ResultType.Barcode:
+          if (QRUtils.verifySSOLogin(barcodeResult) != null) {
+            result = BarcodeScanResult(true, result: barcodeResult);
+          } else {
+            result = BarcodeScanResult(false, errorType: QRError.invalidQR);
+          }
+          break;
+        case ResultType.Error:
+          result = BarcodeScanResult(false, errorType: QRError.invalidQR);
+          break;
```
joehoag, line 39: Should the user be able to tell the difference between failed login and ResultType.Error?  Maybe the latter is due to a mechanical failure, lack of camera permissions, blurry scan, etc...?  Should we have different error codes / error messages between the two?

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen_interactor.dart
```diff
@@ -24,24 +24,30 @@ class QRLoginTutorialScreenInteractor {
...
-      if (e.code == BarcodeScanner.CameraAccessDenied) {
+      if (e.code == BarcodeScanner.cameraAccessDenied) {
         result = BarcodeScanResult(false, errorType: QRError.cameraError);
       } else {
         // Unknown error while scanning
         result = BarcodeScanResult(false, errorType: QRError.invalidQR);
       }
-    } on FormatException {
-      // User returned, do nothing
-    } catch (e) {
```
joehoag, line 42: Do we no longer need a general error catcher?

apps/flutter_parent/pubspec.lock
```diff
@@ -696,6 +696,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "3.0.12"
+  protobuf:
+    dependency: transitive
+    description:
+      name: protobuf
+      url: "https://pub.dartlang.org"
+    source: hosted
+    version: "1.0.1"
```
joehoag, line 705: Whoa -- protobufs.  Any indication as to whether this addition beefed up the size of our app?

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen_interactor.dart
Blackchai: I don't believe we do. The example code from the library changed with the update. I did quite a bit of testing but I can try some more. I wasn't able to trigger this catch.

apps/flutter_parent/pubspec.lock
Blackchai: Not sure, but it was a change introduced by the library, not us. I'm not super stoked on the changes for 3.0 of this library though..

joehoag: Splunk shows the size of your apk as 64.741 MB, just marginally higher than the normal 64.2 MB.  So it doesn't look like we bloated that much.

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen_interactor.dart
joehoag: 🤷‍♂️ 👍

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen_interactor.dart
Blackchai: This error isn't for a failed login, but only for barcode scanning errors. The login errors are handled further in, at the SplashScreen.

apps/flutter_parent/pubspec.lock
Blackchai: kk thanks!

PR #753, opened 2020-04-22
[Flutter-Parent][RC-3.2.0] Fix migration for users not on refresh tokens

CalvinKern:
Also some misc null fixes

### DIFF
--- a/apps/flutter_parent/android/app/src/main/kotlin/com/instructure/parentapp/plugins/OldAppMigrations.kt
+++ b/apps/flutter_parent/android/app/src/main/kotlin/com/instructure/parentapp/plugins/OldAppMigrations.kt
@@ -67,7 +67,7 @@ object OldAppMigrations {
 
                 // Some users may not be on refresh tokens, so pull their old token as the 'accessToken'
                 val token = it.optString("token")
-                if (token != null) it.put("accessToken", token)
+                if (token != null && token.isNotEmpty()) it.put("accessToken", token)
 
                 // Add client id/secret if this is the current user
                 if (refreshToken == it.optString("refreshToken")) {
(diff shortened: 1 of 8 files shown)


apps/flutter_parent/android/app/src/main/kotlin/com/instructure/parentapp/plugins/OldAppMigrations.kt
```diff
@@ -65,6 +65,10 @@ object OldAppMigrations {
                 it.put("domain", "${it.optString("protocol") ?: "https"}://${it.optString("domain")}")
                 it.remove("protocol")
 
+                // Some users may not be on refresh tokens, so pull their old token as the 'accessToken'
+                val token = it.optString("token")
+                if (token != null) it.put("accessToken", token)
```
Blackchai, line 70: Hmm so I think token will always be non-null. We set it to an empty string instead of actually nulling it out:

`ApiPrefs.setToken(""); // TODO: Remove when we're 100% using refresh tokens`

Blackchai: So probably just non-null && non-empty

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
```diff
@@ -54,12 +55,13 @@ class _QRLoginTutorialScreenState extends State<QRLoginTutorialScreen> {
       shape: CircleBorder(side: BorderSide(color: Colors.transparent)),
       onPressed: () async {
         var barcodeResult = await locator<QRLoginTutorialScreenInteractor>().scan();
-        if (barcodeResult.isSuccess) {
+        if (barcodeResult?.isSuccess == true) {
```
Blackchai, line 58: :+1:

apps/flutter_parent/android/app/src/main/kotlin/com/instructure/parentapp/plugins/OldAppMigrations.kt
CalvinKern: Oh, good call out

PR #754, opened 2020-04-22
[Student][MBL-14248] Show submission when observer

StephenBrough:
Submission details page would error out because we were using the observer's ID instead of the observee's ID when retrieving submissions.

If the user is an observer, we use the ID of the first observee we retrieve that is in that course.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsEffectHandler.kt
+++ b/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsEffectHandler.kt
@@ -75,8 +75,8 @@ class SubmissionDetailsEffectHandler : EffectHandler<SubmissionDetailsView, Subm
         launch {
             // If the user is an observer, get the id of the first observee that comes back, otherwise use the user's id
             val enrollmentsResult = EnrollmentManager.getObserveeEnrollmentsAsync(true).await()
-            val observee = enrollmentsResult.dataOrNull?.firstOrNull { it.isObserver && it.courseId == effect.courseId }
-            val userId = observee?.associatedUserId ?: ApiPrefs.user!!.id
+            val observeeId = enrollmentsResult.dataOrNull?.firstOrNull { it.isObserver && it.courseId == effect.courseId }?.associatedUserId
+            val userId = observeeId ?: ApiPrefs.user!!.id
 
             val submissionResult = SubmissionManager.getSingleSubmissionAsync(effect.courseId, effect.assignmentId, userId, true).await()
             val assignmentResult = AssignmentManager.getAssignmentAsync(effect.assignmentId, effect.courseId, true).await()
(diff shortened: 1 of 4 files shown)


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsEffectHandler.kt
```diff
@@ -71,7 +73,12 @@ class SubmissionDetailsEffectHandler : EffectHandler<SubmissionDetailsView, Subm
 
     private fun loadData(effect: SubmissionDetailsEffect.LoadData) {
         launch {
-            val submissionResult = SubmissionManager.getSingleSubmissionAsync(effect.courseId, effect.assignmentId, ApiPrefs.user!!.id, true).await()
+            // If the user is an observer, get the id of the first observee that comes back, otherwise use the user's id
+            val enrollmentsResult = EnrollmentManager.getObserveeEnrollmentsAsync(true).await()
+            val observee = enrollmentsResult.dataOrNull?.firstOrNull { it.isObserver && it.courseId == effect.courseId }
```
joehoag, line 78: This seems non-intuitive: Assign an enrollment with isObserver==true to a variable called observee?

apps/student/src/test/java/com/instructure/student/test/assignment/details/submissionDetails/SubmissionDetailsEffectHandlerTest.kt
```diff
@@ -220,7 +281,7 @@ class SubmissionDetailsEffectHandlerTest : Assert() {
         every { ApiPrefs.user } returns user
         every { ApiPrefs.fullDomain} returns "https://www.instructure.com"
 
-        connection.accept(SubmissionDetailsEffect.LoadData(assignment.id, courseId))
+        connection.accept(SubmissionDetailsEffect.LoadData(courseId, assignment.id))
```
joehoag, line 284: LOL.  How was this working before?

StephenBrough: I think we got lucky and the assignment id and course id happened to be the same, lol

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsEffectHandler.kt
StephenBrough: Yeah, it's kind of weird. I named it that way because that's how we're using the Enrollment object, as the observee (to get their ID). When the observer requests enrollments, that's how we know which enrollments are their observees, by which ones have the observer role/type. I suppose I could also grab the ID right then and just rename that variable to `observeeID` to make it more clear.

apps/student/src/test/java/com/instructure/student/test/assignment/details/submissionDetails/SubmissionDetailsEffectHandlerTest.kt
StephenBrough: Which might be something we want to change, just so stuff like this doesn't happen.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/SubmissionDetailsEffectHandler.kt
joehoag: 👍 Up to you.  I'll go ahead and approve.

PR #758, opened 2020-04-23
[Parent] Cameras optional

Blackchai:
To Test:
I don't have a real device without a camera, but using an emulator with 0 cameras did the trick. The splash screen does the check/awaiting, and the login landing page should show/hide the qr code button. We can just re-use the ApiPrefs value for QR scanning pairing codes in the future.

### DIFF
--- a/apps/flutter_parent/lib/screens/splash/splash_screen.dart
+++ b/apps/flutter_parent/lib/screens/splash/splash_screen.dart
@@ -26,7 +26,6 @@ import 'package:flutter_parent/utils/quick_nav.dart';
 import 'package:flutter_parent/utils/service_locator.dart';
 
 class SplashScreen extends StatefulWidget {
-
   final String qrLoginUrl;
 
   SplashScreen({this.qrLoginUrl, Key key}) : super(key: key);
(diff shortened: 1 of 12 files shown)


apps/flutter_parent/lib/screens/splash/splash_screen.dart
```diff
@@ -55,8 +55,20 @@ class _SplashScreenState extends State<SplashScreen> with SingleTickerProviderSt
...
-      _navigate(PandaRouter.login());
-      return _defaultBody(context);
+      if (_cameraFuture == null) {
+        _cameraFuture = locator<SplashScreenInteractor>().getCameraCount();
+      }
+
+      return FutureBuilder(
+          future: _cameraFuture,
+          builder: (BuildContext context, AsyncSnapshot<int> snapshot) {
+            if (snapshot.hasData || snapshot.hasError) {
```
joehoag, line 65: What are the other possibilities besides hasData and hasError?  (Sorry, I'm not very familiar with AsyncSnapshot.)

Blackchai: The other thing we often observe is the connection state, which is sometimes used to determine if we are still loading. There are a few different ways of using this, but by returning the loading container after checking if we have data is another way of showing loading while waiting for either data or an error to be given to the future builder.

PR #760, opened 2020-04-24
[Flutter Parent][MBL-14285] Add Today button to calendar

StephenBrough:


apps/flutter_parent/lib/screens/dashboard/dashboard_screen.dart
```diff
@@ -186,6 +193,30 @@ class DashboardState extends State<DashboardScreen> {
             appBar: PreferredSize(
               preferredSize: Size.fromHeight(107.0),
               child: AppBar(
+                // Today button is only for the calendar and the notifier value is set in the calendar screen
+                actions: [
+                  ChangeNotifierProvider<CalendarTodayNotifier>(
```
CalvinKern, line 198: I wonder if this might be more clean on the dashboard if we combine all these providers into one multiprovider?

StephenBrough: Oooh YEAH! I looked at that but it wasn't immediately obvious how it was to be used, but I typed in some stuff and IT WORKS! EZ

apps/flutter_parent/lib/screens/calendar/calendar_widget/calendar_widget.dart
```diff
@@ -473,6 +477,9 @@ class CalendarWidgetState extends State<CalendarWidget> with TickerProviderState
 
     selectedDay = day;
 
+    // Enable/disable the today button
+    locator<CalendarTodayNotifier>().value = !DateTime.now().isSameDayAs(selectedDay);
```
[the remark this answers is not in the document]
StephenBrough: Oddly enough, removing the double notifying action surfaced a bug. BUT IT DIDN'T LIVE LONG.

apps/flutter_parent/lib/screens/calendar/calendar_widget/calendar_widget.dart
```diff
@@ -190,6 +193,8 @@ class CalendarWidgetState extends State<CalendarWidget> with TickerProviderState
 
   @override
   void initState() {
+    locator<CalendarTodayClickNotifier>().addListener(() => _todayClicked());
```
CalvinKern, line 196: This isn't quite right, it's creating an anonymous callback each time here, instead of passing the reference to _todayClicked to be able to remove it later correctly.

Should be more like:
```
locator<CalendarTodayClickNotifier>().addListener(_todayClicked);
```

apps/flutter_parent/lib/screens/calendar/calendar_widget/calendar_widget.dart
```diff
@@ -226,6 +231,7 @@ class CalendarWidgetState extends State<CalendarWidget> with TickerProviderState
 
     if (widget.startingDate != null) {
       WidgetsBinding.instance.addPostFrameCallback((_) {
+        print('Select day from addPostFrameCallback');
```
CalvinKern, line 234: Agh, so many prints added to this class!

apps/flutter_parent/lib/screens/calendar/calendar_today_click_notifier.dart
```diff
@@ -0,0 +1,21 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import 'package:flutter/material.dart';
+
+class CalendarTodayClickNotifier extends ValueNotifier<bool> {
+  CalendarTodayClickNotifier() : super(false);
+
+  update() => notifyListeners();
```
CalvinKern, line 20: Should delete this update, it's not doing anything (and update dashboard screen onTap to toggle the value)

apps/flutter_parent/lib/screens/calendar/calendar_widget/calendar_widget.dart
StephenBrough: Lol :smh:

apps/flutter_parent/lib/screens/calendar/calendar_today_click_notifier.dart
StephenBrough: How do you trigger the update that way? It looks like I'd have to have a bool in the dashboard screen and set the value of the notifier to !thatValue.

CalvinKern: Yeah, it'd look a little gross on the dashboard. So you could keep this update function, and just have it toggle for you, but you don't need the `notifyListeners` then. (and update this `update` function to set `value = !value`)

StephenBrough: Wouldn't it be easier to just not deal with a value? I don't really need one, I just need essentially a bus event that lets us know when the button was clicked. Doesn't matter what the value is.

StephenBrough: Maybe it would make more sense if I named the function `trigger()` or something, instead of `update()`.

CalvinKern: Hmm, that is a fair point that you don't need the value. The ValueNotifier is a little bit heavier than a simple event bus. It still works, just does a little more than you need it to.

I'm not sure of a better approach to this either. I think your rename `trigger()` would be good though. Makes it seem less like the value is getting used and we just want the notify.

StephenBrough: Yeah, the design of this feature didn't play nicely with the architecture we used. 🤷‍♂️

CalvinKern:
LGTM! Thanks for addressing that feedback!

PR #761, opened 2020-04-24
[Parent] Beef up QR error handling

Blackchai:
This is in response to this [crash](https://console.firebase.google.com/project/automatic-rite-88319/crashlytics/app/android:com.instructure.parentapp/issues/cb52c6344beab447e5e6b1c5c354592a?time=1587513600000:1587772799000&sessionId=5EA24134006600014EC6F3E1B437B260_DNE_0_v2):

How to repro:
Login with qr code, log out immediately, attempt to login again with the same qr code.

Basically, anytime something went wrong inside of the sign on portion of QR login, the error handling wasn't working. This PR fixes those issues and fixes properly handling an error message and pop without resulting in an app crash.

### DIFF
--- a/apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
+++ b/apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
@@ -55,7 +55,13 @@ class _QRLoginTutorialScreenState extends State<QRLoginTutorialScreen> {
       onPressed: () async {
         var barcodeResult = await locator<QRLoginTutorialScreenInteractor>().scan();
         if (barcodeResult.isSuccess) {
-          locator<QuickNav>().pushRoute(context, PandaRouter.qrLogin(barcodeResult.result));
+          final result = await locator<QuickNav>().pushRoute(context, PandaRouter.qrLogin(barcodeResult.result));
+
+          // Await this result so we can show an error message if the splash screen has to pop after a login issue
+          // (This is typically in the case of the same QR code being scanned twice)
+          if (result != null) {
+            _showSnackBarError(context, result);
+          }
         } else {
           _showSnackBarError(
               context,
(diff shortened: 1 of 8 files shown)


apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
```diff
@@ -55,8 +55,14 @@ class _QRLoginTutorialScreenState extends State<QRLoginTutorialScreen> {
       shape: CircleBorder(side: BorderSide(color: Colors.transparent)),
       onPressed: () async {
         var barcodeResult = await locator<QRLoginTutorialScreenInteractor>().scan();
-        if (barcodeResult?.isSuccess == true) {
-          locator<QuickNav>().pushRoute(context, PandaRouter.qrLogin(barcodeResult.result));
+        if (barcodeResult.isSuccess) {
```
joehoag, line 58: We no longer need a null-check for barcodeResult?

apps/flutter_parent/test/screens/login/qr_login_tutorial_screen_test.dart
```diff
@@ -93,6 +93,18 @@ void main() {
     await tester.pump();
     expect(find.text(AppLocalizations().invalidQRCodeError), findsOneWidget);
   });
+
+  testWidgetsWithAccessibilityChecks(
+      'Clicking next scans, returns valid result, but then displays qr login error on login failure', (tester) async {
```
joehoag, line 98: So this tests a successful scan followed by an failed login?

Blackchai: yeah!

apps/flutter_parent/lib/screens/qr_login/qr_login_tutorial_screen.dart
Blackchai: This was a miss actually, I just messed up the merge conflict. However, it reminded me that I wanted this to treat result as non-null. So I make some changes to the interactor to make sure this is always non-null, so we no longer need the null check.

joehoag: 👍

apps/flutter_parent/test/screens/login/qr_login_tutorial_screen_test.dart
joehoag: 👍

PR #767, opened 2020-04-28
[Student][MBL-13745] Generate pairing code

CalvinKern:
To test:
The permission check only works on beta right now (USERS-473)
* Should be able to use the pairing code generated to pair a student in the parent app.
* Should be able to scan the qr code and see the pairing code as a query param using a third party qr scanner (or it will open the parent app to the dashboard, just depends on the qr scanner you use)
* Testing without network connection should prevent navigating to this screen from settings
* Testing as a teacher/observer using the student app should show an error on this screen (they can't generate pairing codes, even though the permission isn't restricted)


libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/logic/PactAssignmentLogic.kt
```diff
@@ -240,7 +240,6 @@ fun assertAssignmentPopulated(description: String, assignment: Assignment, field
     assertNotNull("$description + unlockAt", assignment.unlockAt)
     assertNotNull("$description + freeFormCriterionComments", assignment.freeFormCriterionComments)
     assertNotNull("$description + published", assignment.published)
-    assertNotNull("$description + muted", assignment.muted)
```
joehoag, line 243: The actual Pact contract, defined in the populateAssignmentFields() method above this, still expects a "muted" field to be populated in an Assignment object.  Does that need to change?

CalvinKern: Hmm, probably? I removed the line locally that is still populating that field and it's still passing the task that this line failed (and the AssignmentApiPactTests). Seems like we don't need it anymore so I'll push up that removal as well. Nice catch!

apps/polling/src/main/java/com/instructure/androidpolling/app/asynctasks/PollsLogoutTask.kt
```diff
@@ -38,6 +39,10 @@ class PollsLogoutTask(type: Type) : LogoutTask(type) {
         return InitLoginActivity.createIntent(ContextKeeper.appContext)
     }
 
+    override fun createQRLoginIntent(context: Context, uri: Uri): Intent? {
+        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
+    }
```
CalvinKern, line 44: This was an error by Android Studio every time I started up. This class was just missing the implementation for this method, but since we don't use polls anymore I just left it stubbed (but at least AS will stop complaining now)

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/logic/PactAssignmentLogic.kt
joehoag: If we push up a change to the actual contract, we'll need to re-publish the contract as part of the work.  (Publishing is not done automatically, just manually whenever there is a change to a contract.)  I don't *think* that publishing the new contract will break anything on the server side, because we're just relaxing the contract.

apps/polling/src/main/java/com/instructure/androidpolling/app/asynctasks/PollsLogoutTask.kt
Blackchai: Hah, nice fix. That's funny, I remember seeing that and thought I had fixed it. Guess not.

automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/mockCanvas/endpoints/UserEndpoints.kt
```diff
@@ -286,3 +284,14 @@ object UserSettingsEndpoint : Endpoint(response = {
         request.successResponse(settings)
     }
 })
+
+var pairingCodeCount = 0
```
Blackchai, line 288: Total nit pick, but we have been using the MockCanvas.newItemId() fun for this sorta thing. Maybe if there are other code changes after I test we can update this. Otherwise.. /shrug.

libs/login-api-2/build.gradle
```diff
@@ -106,8 +106,8 @@ dependencies {
     implementation Libs.ANDROIDX_RECYCLERVIEW
     implementation Libs.ANDROIDX_CONSTRAINT_LAYOUT
 
-    implementation('com.journeyapps:zxing-android-embedded:4.1.0') { transitive = false }
-    implementation 'com.google.zxing:core:3.3.0'
+    implementation(Libs.JOURNEY_ZXING) { transitive = false }
```
Blackchai, line 109: noice

automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/mockCanvas/endpoints/UserEndpoints.kt
CalvinKern: hmm, we can use newItemId, but this is to test against a specifically known value. I'm not sure we'd be able to do that reliably in the test using the newItemId. It would probably need an entirely different solution where we specify the pairing codes then. I can add that if you'd like.

Blackchai: Ah, yeah I see. Just looked at the interaction test. You could achieve a very similar goal by just asserting that the new pairing code is not equal to the old and is greater than (newItemId also just increments in value). That being said, this is _super_ minor. NewItemId() is mostly just used for IDs, so this doesn't even make that much sense. If I don't see anything in the manual testing I say we leave this as is.

PR #768, opened 2020-04-29
[Flutter-Parent] Rename Page to CanvasPage

CalvinKern:
This is just getting ahead of some changes to Navigator in the beta channel where we have a conflicting name for Page.


JordanMarshall:
I'm curious why this necessitates renaming classes on our end. Can we get away with simply hiding the offending class from the import in affected files?

CalvinKern:
Hmm, I didn't know about hiding in imports. We can do that. Any screen that imports page and material will have to hide Page on the material import. That would require less changes, and we probably won't use Page from the material navigator directly.

I still prefer a rename just so we aren't shutting off using Navigator's Page in the future (without using import prefixes, in which case why not just do a simple rename). Also, this allows us to play around with less local changes while switching flutter channels. Which, I guess beta will change next week so that may not be as important now.

Though, if you think it's better to just hide the offending class in the import, I will just close this PR and we can address the hiding when it arises (shows as a warning when hiding an import that doesn't have the specified class, which stable doesn't have the Navigator's Page class).

JordanMarshall:
idk, on one hand it seems more pertinent to use existing language features to deal with a problem as common naming collisions, but on the other hand I recognize that `Page` is somewhat generic and tbh I'm not opposed to renaming it. Although I would prefer `CanvasPage` rather than `CoursePage` since pages aren't exclusive to courses.

CalvinKern:
Yeah, I think because `Page` is pretty generic I don't want to overshadow it. Great feedback about `CanvasPage`, I do like that better. I'll push up that change.

PR #772, opened 2020-04-30
[Flutter-Parent] Fix small, miscellaneous problems

CalvinKern:
When an empty url is passed to an image loader (FadeInImage/NetworkImage/etc) it can throw a non-fatal exception. It won’t crash the app but does create a non-fatal on Firebase.

Encoding domain provided by users so navigation can happen. If the url is invalid, the web login page may show an empty screen, but that’s what our native apps do in the same scenario (and it is better than not navigating to the screen with a non url safe character)

Updated native splash screen to use same accessible color as flutter

### DIFF
--- a/apps/flutter_parent/lib/router/panda_router.dart
+++ b/apps/flutter_parent/lib/router/panda_router.dart
@@ -107,7 +107,7 @@ class PandaRouter {
     String authenticationProvider = null,
     LoginFlow loginFlow = LoginFlow.normal,
   }) =>
-      '$_loginWeb?${_RouterKeys.domain}=$domain&${_RouterKeys.authenticationProvider}=$authenticationProvider&${_RouterKeys.loginFlow}=${loginFlow.toString()}';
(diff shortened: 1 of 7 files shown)


apps/flutter_parent/lib/router/panda_router.dart
```diff
@@ -131,7 +131,7 @@ class PandaRouter {
   static final String _simpleWebView = '/internal';
 
   static String simpleWebViewRoute(String url, String infoText) =>
-      '/internal?${_RouterKeys.url}=${Uri.encodeQueryComponent(url)}&${_RouterKeys.infoText}=$infoText';
+      '/internal?${_RouterKeys.url}=${Uri.encodeQueryComponent(url)}&${_RouterKeys.infoText}=${Uri.encodeQueryComponent(infoText)}';
```
CalvinKern, line 134: Encoding info text in case another language uses a character that is not allowed in a uri.

apps/flutter_parent/test/router/panda_router_test.dart
```diff
@@ -137,6 +137,16 @@ void main() {
       expect(widget.domain, domain);
     });
 
+    test('loginWeb returns web login screen with non-uri safe charachters', () {
```
StephenBrough, line 140: charachters -> characters 

:D

CalvinKern: lol, oops!

PR #775, opened 2020-05-01
[Parent][Stability][MBL-14341]: Fail gracefully from a remote config fetch failure

joehoag:


joehoag:
I was able to verify that a release build run while airplane mode was on would launch to black screen before these fixes and do a successful launch after these fixes.

PR #777, opened 2020-05-01
[Flutter-Parent][MBL-14286] Rating dialog

CalvinKern:
Added a rating dialog, only shows after 4 weeks, good luck testing that 😉

Also consolidated some intent logic behind a veneer to be reusable by different screens.

Unable to add a widget test for the dashboard showing this new dialog, as we purposefully don't show the rating dialog during tests to keep e2e tests stable (this may not be an issue in flutter, it was in native so I'm also hiding it in tests here).

To test locally:
* Comment out the first block of the `showDialogIfPossible ` in the RatingDialog class so that the method just runs `showDialog`. This will show the dialog every time you open the app. 
* 4 & 5 stars go to the app store (native only did 5 stars, we're adding 4 here as we need the positive ratings)
* 3 and under are asked to leave a comment. If no comment is provided the dialog closes. If a comment is provided it opens an email (and closes the dialog)

### DIFF
--- a/apps/flutter_parent/lib/network/utils/api_prefs.dart
+++ b/apps/flutter_parent/lib/network/utils/api_prefs.dart
@@ -33,13 +33,16 @@ import 'package:shared_preferences/shared_preferences.dart';
 import 'dio_config.dart';
 
 class ApiPrefs {
+  static const String KEY_CAMERA_COUNT = 'camera_count';
+  static const String KEY_CURRENT_LOGIN_UUID = 'current_login_uuid';
+  static const String KEY_CURRENT_STUDENT = 'current_student';
   static const String KEY_HAS_MIGRATED = 'has_migrated_from_old_app';
   static const String KEY_HAS_CHECKED_OLD_REMINDERS = 'has_checked_old_reminders';
   static const String KEY_HAS_MIGRATED_TO_ENCRYPTED_PREFS = 'has_migrated_to_encrypted_prefs';
   static const String KEY_LOGINS = 'logins';
-  static const String KEY_CURRENT_LOGIN_UUID = 'current_login_uuid';
-  static const String KEY_CURRENT_STUDENT = 'current_student';
-  static const String KEY_CAMERA_COUNT = 'camera_count';
+  static const String KEY_RATING_SHOW_AGAIN_WAIT = 'date_show_again';
+  static const String KEY_RATING_DONT_SHOW_AGAIN = 'dont_show_again';
+  static const String KEY_RATING_FIRST_LAUNCH_DATE = 'date_first_launched';
 
   static EncryptedSharedPreferences _prefs;
   static PackageInfo _packageInfo;
@@ -236,6 +239,8 @@ class ApiPrefs {
     }
   }
 
+  /// Prefs
+
   static String getCurrentLoginUuid() => _getPrefString(KEY_CURRENT_LOGIN_UUID);
 
   static User getUser() => getCurrentLogin()?.currentUser;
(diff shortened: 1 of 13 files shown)


apps/flutter_parent/lib/utils/common_widgets/rating_dialog.dart
```diff
@@ -0,0 +1,253 @@
...
+  /// - when the user sees the dialog again there will be a "don't show again" button
+  /// - when the user sees the dialog, there are a few use cases for when to show the dialog again
+  ///
+  /// 1. User presses 5 stars -> take user to play store and don't show dialog again
+  /// 2. User presses < 5 stars with no comment -> show again 4 weeks later
+  /// 3. User presses < 5 stars with a comment -> show again 6 weeks later
+  /// 4. User presses back -> show again 4 weeks later
+  static Future<void> asDialog(BuildContext context) {
+    // Don't show dialog in tests, so that they run more stable. Testing this rating dialog can be achieved by
+    // using the method `RatingDialog.showDialogIfPossible()`
```
joehoag, line 45: Thanks for pro-actively disabling this during tests!

apps/flutter_parent/lib/utils/common_widgets/rating_dialog.dart
```diff
@@ -0,0 +1,255 @@
...
+  static const SIX_WEEKS = 42;
+  static const _RATING_TO_APP_STORE_THRESHOLD = 3;
+
+  /// Will show the rating dialog when:
+  /// - user has used the app for 4 weeks
+  /// - when the user sees the dialog again there will be a "don't show again" button
+  /// - when the user sees the dialog, there are a few use cases for when to show the dialog again
+  ///
+  /// 1. User presses 5 stars -> take user to play store and don't show dialog again
+  /// 2. User presses < 5 stars with no comment -> show again 4 weeks later
```
Blackchai, line 40: super minor nit pick, but this should say < 4 stars now, right?

apps/flutter_parent/lib/network/utils/api_prefs.dart
```diff
@@ -266,6 +271,23 @@ class ApiPrefs {
 
   static Future<void> setCameraCount(int count) => _setPrefInt(KEY_CAMERA_COUNT, count);
 
+  static int getRatingShowAgainWait() => _getPrefInt(KEY_RATING_SHOW_AGAIN_WAIT);
```
Blackchai, line 274: Definitely a nit pick, sorry. What are your thoughts on converting these prefs to Iso8601 Strings? That way we could use DateTime to work with them and convert back and forth. I don't feel particularly strongly about it though.

Blackchai: We could even make a setter/getter pref helper method for "DateTime" prefs so it is converted to and from usage.

apps/flutter_parent/test/utils/widgets/rating_dialog_test.dart
```diff
@@ -0,0 +1,280 @@
...
+import 'package:flutter_parent/utils/veneers/AndroidIntentVeneer.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/mockito.dart';
+
+import '../accessibility_utils.dart';
+import '../platform_config.dart';
+import '../test_app.dart';
+import '../test_helpers/mock_helpers.dart';
+
+void main() {
```
Blackchai, line 29: great tests!

apps/flutter_parent/lib/utils/common_widgets/rating_dialog.dart
CalvinKern: lol, oops. Copied from the old code and didn't update when we changed to 4 stars. Definitely should be updated.

apps/flutter_parent/lib/network/utils/api_prefs.dart
CalvinKern: Yeah, I didn't really think about it, but datetime is really easy to work with in Flutter so we could have these be the actual time rather than just an offset to wait from the last time.

What I'll probably do then is just set a 'next show time' and get rid of show again wait and first launch time.

PR #780, opened 2020-05-04
[Parent][MBL-14349] Additional course filtering

Blackchai:
Play around with an observer and its courses. Change the start dates, end dates, setting restrictions. Same with the sections/terms. See unit tests for various cases that are being covered. I have a user, but I'd rather someone else test/compare with their user so we get more varied testing in on this case.


apps/flutter_parent/lib/models/course.dart
```diff
@@ -158,6 +167,51 @@ abstract class Course implements Built<Course, CourseBuilder> {
...
   String contextFilterId() => 'course_${this.id}';
+
+  /// Verifies that the course should be displayed within the parent app
+  bool isValidForDate() {
+    if (accessRestrictedByDate) return false;
+
+    if (workflowState == 'completed') return false;
+
+    final now = DateTime.now();
+    final isWithinCourseDates = _isWithinDates(startAt, endAt, now);
```
joehoag, line 178: Nit: Do you need to worry about GMT/local time conversion issues?  I *believe* that we store startAt/endAt in GMT, though I could be wrong.  (I realize that, even if we got that wrong, it wouldn't make much of a difference for the purposes of this PR.)

apps/flutter_parent/lib/screens/courses/courses_screen.dart
```diff
@@ -56,28 +55,27 @@ class _CoursesScreenState extends State<CoursesScreen> {
     }
   }
 
-  Future<List<Course>> _loadCourses({bool forceRefresh = false}) => _interactor.getCourses(isRefresh: forceRefresh);
+  Future<List<Course>> _loadCourses({bool forceRefresh = false}) =>
+      _interactor.getCourses(isRefresh: forceRefresh, studentId: _student.id ?? null);
```
joehoag, line 59: Isn't "?? null" a no-op?

Blackchai: Oh woops, in my brain this was meant to replace an empty id with null. I'll fix it, thanks.

apps/flutter_parent/lib/models/course.dart
Blackchai: Hmm I had assumed that all dates involved here would pull from locale. I'll look into it!

joehoag: (Resolved offline.  Looks like the code is doing the right thing as is.)

joehoag:
OK, manually verified that not-yet-started courses do not show up, already-ended courses do not show up, and an already-ended section of a still-ongoing course does not show up.

PR #783, opened 2020-05-05
[Student][Groups][MBL-14234]: Pact tests for groups

joehoag:


libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/GroupApiPactTests.kt
```diff
@@ -0,0 +1,191 @@
...
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class GroupApiPactTests : ApiPactTestBase() {
+
+    val _PROVIDER_STATE_NAME="mobile courses with groups"
+
+    // Common logic
+    private fun createService(pathPrefix: String = "/api/v1/"): GroupAPI.GroupInterface {
```
Blackchai, line 37: In the past I've seen you do something like:
`private fun createService(caller: String = DEFAULT_MOBILE_STUDENT): `

For this createService function. Is this a different case?

joehoag: I could create a DEFAULT_MOBILE_PATH_PREFIX and use that instead of the string literal.  

I copied this logic over from a previous test and it turns out that we don't even need to be able to specify a non-standard path for these tests, so I could just remove the parameter as well.

I'll push some sort of merge to address the issue.

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/GroupApiPactTests.kt
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    //
+    // region request a user's groups
+    //
+
+    val getAccountGroupsQuery = "include[]=favorites&include[]=can_access"
+    val getAccountGroupsPath = "/api/v1/users/self/groups"
+    val getAccountGroupsConfig = PactGroupFieldConfig.fromQueryString(getAccountGroupsQuery)
+    val getAccountGroupsResponseBody = LambdaDsl.newJsonArray { arr ->
```
Blackchai, line 49: I don't think courses _always_ have groups. That's okay though, right? This pact is only for the case where we _have_ groups I assume?

joehoag: Correct.  I created a provider state where the courses have groups, and I'm running against that provider state.

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/GroupApiPactTests.kt
joehoag: OK, for now, I've just removed the pactPrefix parameter in that method, since we don't need it.  In future work, I will create a DEFAULT_MOBILE_PATH_PREFIX constant for use when the path prefix is variable.  (Don't want to make any sweeping changes now as I have done a ton of testing and would prefer to commit soon.)

PR #788, opened 2020-05-07
[Student][Teacher][MBL-14361]: Fix remote config overrides on physical devices

joehoag:
On physical devices, local override of remote config settings was not working.

Fix: Apparently onKeyListener doesn't work so well when the soft keyboard is in use, so I switched over to TextChangedListener, which works much better.

### DIFF
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
@@ -20,7 +20,6 @@ import android.os.Bundle
 import android.text.Editable
 import android.text.TextWatcher
 import android.util.Log
-import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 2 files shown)


libs/pandautils/src/main/java/com/instructure/pandautils/fragments/RemoteConfigParamsFragment.kt
```diff
@@ -66,16 +64,21 @@ private class RemoteConfigParamAdapter : RecyclerView.Adapter<RecyclerView.ViewH
...
             valueEditText.setText(RemoteConfigPrefs.getString(param.rc_name, param.safeValueAsString))
-            valueEditText.setOnKeyListener() {view, id, event ->
-                if(event.action == KeyEvent.ACTION_UP) {
-                    Log.d("RemoteConfigPrefs", "onKeyListener ${param.rc_name} content ${(view as EditText).text}")
-                    RemoteConfigPrefs.putString(param.rc_name, (view as EditText).text.toString())
+            valueEditText.addTextChangedListener(object : TextWatcher {
+                override fun afterTextChanged(s: Editable?) {
+                    Log.d("RemoteConfigPrefs", "afterTextChanged: ${param.rc_name} = ${s.toString()}")
+                    RemoteConfigPrefs.putString(param.rc_name, s.toString())
                 }
```
joehoag, line 71: Sadly, this callback fires twice per change on emulators (but only once on physical devices).  I looked around for a fix, but was unsuccessful.  This won't break functionality on emulators, it will just be a little inefficient.

PR #789, opened 2020-05-07
[Student][MBL-14379] Fix Flutter API calls when masquerading

JordanMarshall:


CalvinKern:
So... this works... but it does throw off the a11y offsets when the masquerade title is visible.
Not sure if that should be addressed separately or in this PR, or at all (since it's mainly a support thing to masquerade)

JordanMarshall:
Yeah, I figured it would upset a11y since the masquerade UI causes FlutterView to be misaligned with the window insets. I think it's fine for now and we'll just wait for a proper upstream fix from the flutter team.

PR #790, opened 2020-05-07
[Student][MBL-14363] Testing infrastructure for student flutter

CalvinKern:
Added a test for each directory to have that established. Migrated over all the infrastructure related things to testing, making a new 'testutils' directory for it (rather than being intermixed with actual 'utils' tests for production code)

Had to make some changes to colors/styles too get a11y to pass.

To test, just run the tests 😄
The code coverage tool should work for this module now. Lots of missing coverage from tests that still need to be ported over or written.

### DIFF
--- a/libs/flutter_student_embed/test/testutils/mock_helpers.dart
+++ b/libs/flutter_student_embed/test/testutils/mock_helpers.dart
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2020 - present Instructure, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, version 3 of the License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Create a mocked RemoteConfig object.
+// If valueSettings != null, then (1) a mocked settings fetch will occur, and (2) the retrieved
+// settings will correspond the specified values.
+import 'dart:io';
+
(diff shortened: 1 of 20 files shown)


libs/flutter_student_embed/test/testutils/mock_helpers.dart
```diff
@@ -0,0 +1,48 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Create a mocked RemoteConfig object.
```
joehoag, line 17: Comment seems out of place?

CalvinKern: Yupp, I'll clean those up. Thanks

libs/flutter_student_embed/test/testutils/test_app.dart
```diff
@@ -0,0 +1,243 @@
...
+  /// * [config] A [PlatformConfig] object passed to TestApp that is used during [setupPlatformChannels]
+  /// * [configBlock] If you want to do custom configuration of the app or prefs, this is an easy place to setup custom data
+  ///
+  /// Example call:
+  /// testWidgetsWithAccessibility((tester) {
+  ///   await showWidgetFromTap(
+  ///     tester,
+  ///     (context) => ErrorReportDialog.asDialog(context),
+  ///     locale: Locale('ar'),
+  ///     config: PlatformConfig(initWebView: true, initLoggedInUser: login, initRemoteConfig: mockRemoteConfig),
```
joehoag, line 62: Looks like we no longer support initRemoteConfig?

CalvinKern: Nope, I'll remove that from the comment as well.

I figured we would bring over remote config when we need it. We may end up with a different solution here as opposed to parent since we don't have to make the call to fetch in this module, student will have to provide the values to us and I'm not sure how that will work yet.

PR #795, opened 2020-05-08
[Student][Hotfix-6.7.1] Fix issue with Student view

StephenBrough:
Sometimes going into Student view would work, but would have some weird behavior that caused the Student app to return to the Landing screen. This could also wipe out the Teacher's token. 

This PR also solves another issue that allowed the user to go into Student view, then press the back button, returning to the landing screen while still being in masquerade mode.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/activity/CallbackActivity.kt
+++ b/apps/student/src/main/java/com/instructure/student/activity/CallbackActivity.kt
@@ -185,9 +185,8 @@ abstract class CallbackActivity : ParentActivity(), InboxFragment.OnUnreadCountI
             ApiType.API -> {
                 val oldLocale = ApiPrefs.effectiveLocale
                 // This has a habit of running after we've retrieved the Test user account when in Student View, which then
-                // overrides the Test user info in ApiPrefs. We only want to override if we're not in Student view, and
-                // the user in ApiPrefs is null.
-                if (!ApiPrefs.isStudentView && ApiPrefs.user == null)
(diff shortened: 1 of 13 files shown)


apps/student/src/main/java/com/instructure/student/activity/CallbackActivity.kt
```diff
@@ -184,7 +184,11 @@ abstract class CallbackActivity : ParentActivity(), InboxFragment.OnUnreadCountI
         when (type) {
             ApiType.API -> {
                 val oldLocale = ApiPrefs.effectiveLocale
-                ApiPrefs.user = user
+                // This has a habit of running after we've retrieved the Test user account when in Student View, which then
+                // overrides the Test user info in ApiPrefs. We only want to override if we're not in Student view, or
+                // the user in ApiPrefs is null.
+                if (!ApiPrefs.isStudentView && ApiPrefs.user == null)
+                    ApiPrefs.user = user
```
CalvinKern, line 191: Your boolean check doesn't match the comment logic. Just want to double check you want an && here instead of an ||.

StephenBrough: Ooh, good catch!

StephenBrough: Huh, I thought I could get rid of that second half of the condition check, but the app went into a crash loop, so I changed it back. Then I removed it again, and now it works. I think I'll take it out, I don't know the reasons behind why it's there, but I think Student view can be an exception.

PR #797, opened 2020-05-11
[Student][MBL-14374] Fix white planner screen after backgrounding the app

JordanMarshall:
To test: Tap multiple tabs on the bottom nav bar such that there are two more more planner fragments in the back stack. Background and then resume the app. Navigating back to a previous planner fragment should no longer result in a white screen and the planner should be fully interactive.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/fragment/FlutterCalendarFragment.kt
+++ b/apps/student/src/main/java/com/instructure/student/fragment/FlutterCalendarFragment.kt
@@ -19,17 +19,20 @@ package com.instructure.student.fragment
 
 import android.content.Context
 import android.os.Bundle
+import androidx.fragment.app.Fragment
 import com.google.gson.Gson
 import com.instructure.canvasapi2.models.PlannerItem
 import com.instructure.student.flutterChannels.FlutterComm
 import com.instructure.student.util.AppManager
-import io.flutter.embedding.android.*
+import io.flutter.embedding.android.FlutterFragment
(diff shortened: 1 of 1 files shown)


apps/student/src/main/java/com/instructure/student/fragment/FlutterCalendarFragment.kt
```diff
@@ -41,7 +44,35 @@ class FlutterCalendarFragment : FlutterFragment() {
...
+        // See comments in onStart
+        if (hidden) skipLifecycleCall() else super.onResume()
+    }
+
+    /**
+     *  Fakes a super call to a lifecycle method by setting 'mCalled' to true, allowing us to skip the side effects
+     *  of the super call while also avoiding a SuperNotCalledException.
+     */
+    private fun skipLifecycleCall() {
+        val field = Fragment::class.java.getDeclaredField("mCalled")
```
StephenBrough, line 69: lol

apps/student/src/main/java/com/instructure/student/fragment/FlutterCalendarFragment.kt
```diff
@@ -41,7 +44,35 @@ class FlutterCalendarFragment : FlutterFragment() {
         super.onActivityCreated(savedInstanceState)
     }
 
+    override fun onStart() {
+        /* If there are multiple calendar fragments in the back stack, backgrounding and resuming the app
```
StephenBrough, line 48: Excellent comments. _Excellent._

PR #800, opened 2020-05-12
MBL-14364, MBL-14365: Port calendar tests parent -> student

joehoag:
Copied/adapted the calendar tests in flutter-parent into the student app.

It's too bad that the conversions will be viewed as new files, so that reviewers can't see all of the tweaks that were done to the original tests.  For the most part, the heavy tweaking was done in calendar_day_list_tile_test, specifically in the Interactive tests therein.

Three tests were commented-out (2 calendar-filter tests, 1 test of launching the calendar to an arbitrary day), as they don't immediately apply to student.

### DIFF
--- a/libs/flutter_student_embed/lib/screens/calendar/calendar_screen.dart
+++ b/libs/flutter_student_embed/lib/screens/calendar/calendar_screen.dart
@@ -109,13 +109,13 @@ class CalendarScreenState extends State<CalendarScreen> {
         ],
       ),
       floatingActionButton: FloatingActionButton(
-        backgroundColor: Theme.of(context).buttonColor,
-        onPressed: () async {
-          var updatedDates = await locator<QuickNav>().push(context, CreateUpdateToDoScreen(initialDate: _currentDay));
-          _refreshDates(updatedDates);
-        },
-        child: Icon(Icons.add),
-      ),
+          backgroundColor: Theme.of(context).buttonColor,
+          onPressed: () async {
+            var updatedDates =
+                await locator<QuickNav>().push(context, CreateUpdateToDoScreen(initialDate: _currentDay));
+            _refreshDates(updatedDates);
+          },
+          child: Icon(Icons.add, semanticLabel: L10n(context).newToDo)),
       body: CalendarWidget(
         key: _calendarKey,
         fetcher: _fetcher,
--- a/libs/flutter_student_embed/lib/screens/calendar/planner_fetcher.dart
+++ b/libs/flutter_student_embed/lib/screens/calendar/planner_fetcher.dart
@@ -30,7 +30,7 @@ class PlannerFetcher extends ChangeNotifier {
 
   final String userDomain;
 
-  final String userId;
+  String userId;
 
   PlannerFetcher({DateTime fetchFirst, @required this.userDomain, @required this.userId}) {
     if (fetchFirst != null) getSnapshotForDate(fetchFirst);
@@ -154,6 +154,11 @@ class PlannerFetcher extends ChangeNotifier {
       ..userId = userId
       ..filters = SetBuilder(contexts));
     await locator<CalendarFilterDb>().insertOrUpdate(filter);
+    reset();
+  }
+
+  void setUserId(String newUserId) {
+    this.userId = newUserId;
     reset();
   }
 
--- a/libs/flutter_student_embed/lib/screens/calendar/planner_fetcher.dart
+++ b/libs/flutter_student_embed/lib/screens/calendar/planner_fetcher.dart
@@ -30,7 +30,7 @@ class PlannerFetcher extends ChangeNotifier {
 
   final String userDomain;
 
-  String userId;
+  final String userId;
 
   PlannerFetcher({DateTime fetchFirst, @required this.userDomain, @required this.userId}) {
     if (fetchFirst != null) getSnapshotForDate(fetchFirst);
@@ -154,11 +154,6 @@ class PlannerFetcher extends ChangeNotifier {
       ..userId = userId
       ..filters = SetBuilder(contexts));
     await locator<CalendarFilterDb>().insertOrUpdate(filter);
-    reset();
-  }
-
-  void setUserId(String newUserId) {
-    this.userId = newUserId;
     reset();
   }
 
--- a/libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
+++ b/libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
@@ -32,20 +32,21 @@ import 'package:mockito/mockito.dart';
 
 import '../../testutils/accessibility_utils.dart';
 import '../../testutils/canvas_model_utils.dart';
+import '../../testutils/mock_helpers.dart';
 import '../../testutils/platform_config.dart';
 import '../../testutils/test_app.dart';
 
 void main() {
   final studentId = '1337';
   final studentName = 'Instructure Panda';
-  _MockCalendarScreenChannel mockChannel;
+  MockCalendarScreenChannel mockChannel;
 
   final student = User((b) => b
     ..id = studentId
     ..name = studentName);
 
   setUp(() {
-    mockChannel = _MockCalendarScreenChannel();
+    mockChannel = MockCalendarScreenChannel();
   });
 
   group('Render', () {
@@ -405,5 +406,3 @@ Function(PlannerItem) _getItemClickHandler(WidgetTester tester, CalendarScreenCh
     }
   };
 }
-
-class _MockCalendarScreenChannel extends Mock implements CalendarScreenChannel {}
--- a/libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
+++ b/libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
@@ -1,0 +1,409 @@
+// Copyright (C) 2020 - present Instructure, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, version 3 of the License.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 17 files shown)


libs/flutter_student_embed/lib/screens/calendar/calendar_screen.dart
```diff
@@ -109,13 +109,13 @@ class CalendarScreenState extends State<CalendarScreen> {
...
-        },
-        child: Icon(Icons.add),
-      ),
+          backgroundColor: Theme.of(context).buttonColor,
+          onPressed: () async {
+            var updatedDates =
+                await locator<QuickNav>().push(context, CreateUpdateToDoScreen(initialDate: _currentDay));
+            _refreshDates(updatedDates);
+          },
+          child: Icon(Icons.add, semanticLabel: L10n(context).newToDo)),
```
joehoag, line 118: A11y fix

libs/flutter_student_embed/lib/screens/calendar/planner_fetcher.dart
```diff
@@ -157,6 +157,11 @@ class PlannerFetcher extends ChangeNotifier {
     reset();
   }
 
+  void setUserId(String newUserId) {
+    this.userId = newUserId;
+    reset();
+  }
+
```
joehoag, line 164: I made userId mutable so that the student version of a test in planner_fetcher_test would work.  If we're better off with an immutable userId, I can remove the affected test.

libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
```diff
@@ -0,0 +1,409 @@
...
+      ));
+      await tester.pump();
+
+      var icon = tester.widget<Icon>(find.byType(Icon));
+
+      expect(icon.icon, CanvasIcons.calendar_day);
+    });
+  });
+
+  group('Interaction', () {
```
joehoag, line 242: OK, for the most part these Interaction tests are really only testing that the _getItemClickHandler() helper method that I added is working correctly.  I'm open to ideas as to more rigorous test methods here, but keep in mind that we really can't test much beyond the channel call.

libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
```diff
@@ -0,0 +1,409 @@
...
+    testWidgetsWithAccessibilityChecks('tapping assignment plannable emits correct channel call', (tester) async {
+      var plannerItem = _createPlannerItem(plannableType: 'assignment', contextName: 'Tap me');
+
+      setupTestLocator((locator) => locator..registerFactory<QuickNav>(() => QuickNav()));
+
+      await tester.pumpWidget(TestApp(
+        CalendarDayListTile(plannerItem, _getItemClickHandler(tester, mockChannel)),
```
joehoag, line 249: I had to add the _getItemClickHandler() parameter to all of the CalendarDayListTile constructors, since we changed that constructor for student.

libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
```diff
@@ -0,0 +1,409 @@
...
+      await tester.pump();
+
+      await tester.tap(find.text('Tap me'));
+      await tester.pump();
+      await tester.pump();
+
+      verify(mockChannel.nativeRouteToItem(plannerItem)).called(1);
+    });
+
+    testWidgetsWithAccessibilityChecks('tapping calendar event plannable navigates to ToDo details screen',
```
joehoag, line 347: This is the one interaction test that actually brings you to a details screen, since the ToDo details screen is part of our Flutter logic.  So not quite as useless as the others.

libs/flutter_student_embed/test/screens/calendar/calendar_widget/calendar_widget_test.dart
```diff
@@ -0,0 +1,1131 @@
...
+      DateTime dateForDayBuilder = null;
+      await tester.pumpWidget(
+        calendarTestApp(
+          CalendarWidget(
+            dayBuilder: (_, day) {
+              dateForDayBuilder = day;
+              return Container();
+            },
+            fetcher: _FakeFetcher(),
+            onTodaySelected: (whatever) {},
```
joehoag, line 306: Had to add this no-op handler to all CalendarWidget constructor calls.

libs/flutter_student_embed/lib/screens/calendar/planner_fetcher.dart
CalvinKern: It doesn't seem like this setUserId is being used in the production code anywhere, so feels like we should just remove the test and keep the user ID final (since the id we're fetching for won't change while still logged in like it would for observees in parent).

libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
CalvinKern: Is it possible to pass in a mock as the item click handler? It seem like we really only care that the right planner item is being passed back in this instance (e2e tests will probably care more about actually navigating to the next screen).

If we can get away with that, we could clean up this test by removing quicknav and mockchannel.

libs/flutter_student_embed/lib/screens/calendar/planner_fetcher.dart
joehoag: 👍 I'll remove the test and revert the change here.

libs/flutter_student_embed/test/screens/calendar/calendar_widget/calendar_widget_test.dart
CalvinKern: We should have at least one test to verify the true and false case for this callback. Making a call to goToDate with a date other than today should return false for 'whatever', making the call with todays date should return true for 'whatever'.

If you can pass a mock here as a function, then verify it was called, that'd be great (not sure if it's possible). Otherwise, the pattern I've used before is to create a completer, and complete it in the `onTodaySelected` block. Then you can await the completer after the call to `goToDate` and do an expect check.

libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
joehoag: Taking a look...

joehoag: We do have the one test that actually shows a details screen ("tapping calendar event plannable navigates to ToDo details screen").  Do we want to give that up with a mocked onItemPressed?  (Still not sure if that is possible.)  Also, it seems kind of equally useless to check that the right planner item is being passed to the (mocked) onItemPressed, as we're the ones that created the planner item above -- it should always pass.

CalvinKern: > Do we want to give that up with a mocked onItemPressed

I think we want to give up on that one, since it's not testing the actual itemClickHandler that's used in prod (it's just copied code that could deviate). If it was the actual delegate that was being used in prod, then this would be reasonable. Since, like you said, this is really just testing that your function operates correctly (nothing in production), it's a pointless test.

> it seems kind of equally useless to check that the right planner item is being passed to the (mocked) onItemPressed...it should always pass

I mean, that's the point of this test. We're making sure that it's giving us the right item back as that's the 'contract'. If someone goes and adds more functionality to this class later to disable the press for some reason, or something similar, this could hopefully catch that.
I like to view tests as not only making sure we did we we intended, but to help prevent future changes from breaking us. So while it does seem naive to test something this simple, it's nice to have some reassurance that this won't silently break in the future.

libs/flutter_student_embed/test/screens/calendar/calendar_screen_test.dart
```diff
@@ -0,0 +1,322 @@
...
+  PlannerApi plannerApi = MockPlannerApi();
+  CalendarFilterDb filterDb = MockCalendarFilterDb();
+  CalendarFilterListInteractor filterInteractor = MockCalendarFilterListInteractor();
+
+  final String userDomain = 'user_domain';
+  final String userId = 'user_123';
+  final Set<String> contexts = {'course_123'};
+
+  when(filterDb.getForUser(any, any))
+      .thenAnswer((_) => Future.value(CalendarFilter((b) => b.filters = SetBuilder({'course_123'}))));
```
CalvinKern, line 51: This can be deleted, it's overridden immediately in the setup method.

libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
joehoag: OK.  I think that I can make the mocked click handler work.

joehoag: Addressed in latest commit.

libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
joehoag: Removed in latest commit.  (In order to simplify faking the click callback.)

libs/flutter_student_embed/test/screens/calendar/calendar_widget/calendar_widget_test.dart
joehoag: Added a test for this in the latest commit.

libs/flutter_student_embed/test/screens/calendar/calendar_screen_test.dart
joehoag: Done.

libs/flutter_student_embed/test/screens/calendar/calendar_day_list_tile_test.dart
CalvinKern: Awesome, looks good! Thanks!

joehoag:
Addressing a test failure on bitrise...

PR #806, opened 2020-05-13
[Flutter-Parent][MBL-14368] Fix webview crash when really large

CalvinKern:
* No longer embedding web views with other native content. Adds a tile to view description in another screen. Added a simple new screen to show html (rather than a url)
* Adds refresh icon to toolbar in course details (when syllabus or home page are visible) to refresh the course data (since PTR doesn't work on full screen web views).

To test:
See ticket for original crash details.
Also test assignment empty states (shows the "There's no description yet" message)
Event empty states (should not show any message with an empty description)
Announcements don't have empty states

Announcements, assignments, and events should all have a button to view web content, rather than an embedded webview.

Course details for Front Page and Syllabus should show a full screen webview (like before) but no pull to refresh functionality right now. Instead, the refresh button added to the course details should reload those screens with new content.


apps/flutter_parent/lib/l10n/app_localizations.dart
```diff
@@ -1345,6 +1346,14 @@ class AppLocalizations {
 
   String get unexpectedError => Intl.message('An unexpected error occurred');
 
+  String get descriptionTitle => Intl.message(
+        'Description',
+        desc: 'Title for screens that contain only a description from Canavs',
+      );
+
+  String get noDescriptionBody =>
+      Intl.message('There\'s no description yet', desc: 'Message used when the assignment has no description');
```
Blackchai, line 1355: Was this string approved by Peyton or Dave? The contraption use doesn't feel quite right. Super minor nit pick tho.

PR #814, opened 2020-05-15
[Student][Parent][MBL-14393] Upgrade to Flutter 1.17

JordanMarshall:
Updates Flutter for parent and student embed to 1.17.1. Due to 1.17 using a newer version of appcompat than Student, some changes (mostly nullability changes) had to be made to otherwise unrelated parts of the Student app. Also, I commented out a few failing Student tests related to l10n, assuming that they will be addressed as part of MBL-14395.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/dialog/FatalErrorDialogStyled.kt
+++ b/apps/student/src/main/java/com/instructure/student/dialog/FatalErrorDialogStyled.kt
@@ -54,8 +54,7 @@ class FatalErrorDialogStyled : DialogFragment() {
     }
 
     override fun onDestroyView() {
-        if (dialog != null && retainInstance)
-            dialog?.setDismissMessage(null)
+        if (retainInstance) dialog?.setDismissMessage(null)
         super.onDestroyView()
     }
 
--- a/apps/student/src/main/java/com/instructure/student/dialog/FatalErrorDialogStyled.kt
+++ b/apps/student/src/main/java/com/instructure/student/dialog/FatalErrorDialogStyled.kt
@@ -55,7 +55,7 @@ class FatalErrorDialogStyled : DialogFragment() {
 
     override fun onDestroyView() {
         if (dialog != null && retainInstance)
-            dialog.setDismissMessage(null)
+            dialog?.setDismissMessage(null)
         super.onDestroyView()
     }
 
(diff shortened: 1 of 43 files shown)


apps/student/src/main/java/com/instructure/student/fragment/ParentFragment.kt
```diff
@@ -198,18 +198,6 @@ abstract class ParentFragment : DialogFragment(), ConfigureRecyclerView, Fragmen
...
-        // Very important fix for the support library and child fragments.
-        try {
-            val childFragmentManager = Fragment::class.java.getDeclaredField("mChildFragmentManager")
-            childFragmentManager.isAccessible = true
-            childFragmentManager.set(this, null)
-        } catch (e: NoSuchFieldException) {
-            e.printStackTrace()
-        } catch (e: IllegalAccessException) {
-            e.printStackTrace()
-        }
```
JordanMarshall, line 211: This workaround is no longer needed in recent versions of appcompat. In fact, leaving it in here was causing crashes.

apps/student/src/main/java/com/instructure/student/dialog/HelpDialogStyled.kt
```diff
@@ -89,7 +89,7 @@ class HelpDialogStyled : DialogFragment() {
 
     override fun onDestroyView() {
         if (dialog != null && retainInstance)
```
StephenBrough, line 91: Could probably just get rid of the null check.

apps/student/src/main/java/com/instructure/student/dialog/FatalErrorDialogStyled.kt
```diff
@@ -55,7 +55,7 @@ class FatalErrorDialogStyled : DialogFragment() {
 
     override fun onDestroyView() {
         if (dialog != null && retainInstance)
```
StephenBrough, line 57: Could probably just get rid of the null check.

apps/student/src/main/java/com/instructure/student/dialog/LegalDialogStyled.kt
```diff
@@ -100,7 +100,7 @@ class LegalDialogStyled : AppCompatDialogFragment() {
     }
 
     override fun onDestroyView() {
-        if (dialog != null && retainInstance) dialog.setDismissMessage(null)
+        if (dialog != null && retainInstance) dialog?.setDismissMessage(null)
```
StephenBrough, line 103: Could probably just get rid of the null check.

PR #816, opened 2020-05-18
[Teacher][MBL-14300] Mark discussion reply as unread

StephenBrough:
Make sure to test RTL as well!

### DIFF
--- a/apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionBottomSheetMenuFragment.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionBottomSheetMenuFragment.kt
@@ -29,6 +29,11 @@ import kotlinx.android.synthetic.main.bottom_sheet_discussion_menu.view.*
 class DiscussionBottomSheetMenuFragment : BottomSheetDialogFragment() {
     lateinit var callback: (DiscussionBottomSheetChoice) -> Unit
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        retainInstance = true
+    }
+
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
         return inflater.inflate(R.layout.bottom_sheet_discussion_menu, container, false)
     }
--- a/apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionBottomSheetMenuFragment.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionBottomSheetMenuFragment.kt
@@ -24,16 +24,11 @@ import androidx.fragment.app.FragmentManager
 import com.google.android.material.bottomsheet.BottomSheetDialogFragment
 import com.instructure.pandautils.utils.dismissExisting
 import com.instructure.teacher.R
+import com.instructure.teacher.events.DiscussionOverflowMenuClickedEvent
 import kotlinx.android.synthetic.main.bottom_sheet_discussion_menu.view.*
+import org.greenrobot.eventbus.EventBus
 
 class DiscussionBottomSheetMenuFragment : BottomSheetDialogFragment() {
-    lateinit var callback: (DiscussionBottomSheetChoice) -> Unit
-
-    override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
-        retainInstance = true
-    }
-
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
         return inflater.inflate(R.layout.bottom_sheet_discussion_menu, container, false)
     }
@@ -41,30 +36,31 @@ class DiscussionBottomSheetMenuFragment : BottomSheetDialogFragment() {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         view.markAsUnread.setOnClickListener {
-            callback(DiscussionBottomSheetChoice.MARK_AS_UNREAD)
+            EventBus.getDefault().post(DiscussionOverflowMenuClickedEvent(DiscussionBottomSheetChoice.MARK_AS_UNREAD, entryId))
             this.dismiss()
         }
+
         view.edit.setOnClickListener {
-            callback(DiscussionBottomSheetChoice.EDIT)
+            EventBus.getDefault().post(DiscussionOverflowMenuClickedEvent(DiscussionBottomSheetChoice.EDIT, entryId))
             this.dismiss()
         }
+
         view.delete.setOnClickListener {
-            callback(DiscussionBottomSheetChoice.DELETE)
+            EventBus.getDefault().post(DiscussionOverflowMenuClickedEvent(DiscussionBottomSheetChoice.DELETE, entryId))
             this.dismiss()
         }
     }
 
     companion object {
-        fun newInstance(clickCallback: (DiscussionBottomSheetChoice) -> Unit): DiscussionBottomSheetMenuFragment {
-            return DiscussionBottomSheetMenuFragment().apply {
-                callback = clickCallback
-            }
-        }
+        var entryId: Long = -1
+
+        fun newInstance(): DiscussionBottomSheetMenuFragment = DiscussionBottomSheetMenuFragment()
 
         @JvmStatic
-        fun show(manager: FragmentManager, callback: (DiscussionBottomSheetChoice) -> Unit) {
+        fun show(manager: FragmentManager, id: Long) {
+            entryId = id
             manager.dismissExisting<DiscussionBottomSheetMenuFragment>()
-            val dialog = newInstance(callback)
+            val dialog = newInstance()
             dialog.show(manager, DiscussionBottomSheetMenuFragment::class.java.simpleName)
         }
     }
--- a/apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionBottomSheetMenuFragment.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionBottomSheetMenuFragment.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2020 - present Instructure, Inc.
+ *
+ *     This program is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU General Public License as published by
+ *     the Free Software Foundation, version 3 of the License.
+ *
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package com.instructure.teacher.fragments
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.fragment.app.FragmentManager
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment
+import com.instructure.pandautils.utils.dismissExisting
+import com.instructure.teacher.R
+import kotlinx.android.synthetic.main.bottom_sheet_discussion_menu.view.*
+
+class DiscussionBottomSheetMenuFragment : BottomSheetDialogFragment() {
+    lateinit var callback: (DiscussionBottomSheetChoice) -> Unit
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        return inflater.inflate(R.layout.bottom_sheet_discussion_menu, container, false)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        view.markAsUnread.setOnClickListener {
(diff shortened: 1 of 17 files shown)


apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionBottomSheetMenuFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+            callback(DiscussionBottomSheetChoice.DELETE)
+            this.dismiss()
+        }
+    }
+
+    companion object {
+        fun newInstance(clickCallback: (DiscussionBottomSheetChoice) -> Unit): DiscussionBottomSheetMenuFragment {
+            return DiscussionBottomSheetMenuFragment().apply {
+                callback = clickCallback
+            }
```
CalvinKern, line 56: This is not allowed by the android framework and causes a crash if the phone is rotated (and the user selects an option). This fragment (with its callback) will be destroyed by the framework and recreated without the callback set, causing a UninitializedPropertyAccessException.

apps/teacher/src/main/java/com/instructure/teacher/presenters/DiscussionsDetailsPresenter.kt
```diff
@@ -257,6 +257,27 @@ class DiscussionsDetailsPresenter(
         }
     }
 
+    @Suppress("EXPERIMENTAL_FEATURE_WARNING")
+    fun markAsUnread(id: Long) {
+        if (mDiscussionMarkAsUnreadApiCalls != null && mDiscussionMarkAsUnreadApiCalls!!.isActive) return
+        mDiscussionMarkAsUnreadApiCalls = weave {
+            var markedAsUnreadId: Long = -1
+
+            val response = awaitApiResponse<Void>{ DiscussionManager.markDiscussionTopicEntryUnread(canvasContext, discussionTopicHeader.id, id, it) }
```
CalvinKern, line 266: Not sure why, but this request is crashing the app for me when it fails (putting into airplane mode to simulate a fail).

apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionBottomSheetMenuFragment.kt
StephenBrough: Ah, I think a `retainInstance` will fix that.

apps/teacher/src/main/java/com/instructure/teacher/presenters/DiscussionsDetailsPresenter.kt
StephenBrough: Oops, I think it needs to be wrapped in a try/catch. EZ.

apps/teacher/src/main/res/layout/bottom_sheet_discussion_menu.xml
```diff
@@ -0,0 +1,76 @@
...
+  ~
+  -->
+
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:paddingTop="8dp">
+
+    <TextView
```
CalvinKern, line 25: Can you add the selectable item background to each of these so we get a ripple when touching the items?

apps/teacher/src/main/java/com/instructure/teacher/presenters/DiscussionsDetailsPresenter.kt
StephenBrough: I guess the mark as read call should also be wrapped in a try/catch. If they're not online, they won't be able to mark conversations as read or unread.

apps/teacher/src/main/res/layout/bottom_sheet_discussion_menu.xml
CalvinKern: Doesn't seem like these text views are using the right text size from the figma designs. Update to 16sp.

There's also a discrepancy of the unread icon between the  designs posted in the JIRA and the designs in Figma. You should reach out to Dave to see which icon he wants to use (if you haven't already).

StephenBrough: Oh wat... he changed that icon after I implemented it, lol. Ok, I'll get those changes in.

apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionBottomSheetMenuFragment.kt
CalvinKern: Hmm, that does fix the crash, but then the dialog goes away on rotate. I guess this is usable.

I hate to say it, but event bus might be a better solution here, rather than a callback passed through on construction. If that's not going to work or is not appropriate, than we can keep this going away on rotation.

libs/pandautils/src/main/assets/discussion_html_header_item_rtl.html
```diff
@@ -287,6 +287,23 @@
...
+				display: block;
+			}
+
+			.meatball_wrapper > div {
+			  width: 100%;
+			  height: 100%;
+			  display: flex;
+			  justify-constent: start;
+			  align-items: center;
+			}
```
CalvinKern, line 305: Ooof, formatting 😂

libs/pandautils/src/main/assets/discussion_html_template_item.html
```diff
@@ -41,23 +41,17 @@
...
-            <div class="edit_wrapper" id="edit___ENTRY_ID__" __EDIT_LISTENER__>
-                __EDIT_TEXT__
-            </div>
-
-            <div class="delete_vertical_divider" aria-hidden="true">__DELETE_DIVIDER__</div>
-
-            <div class="delete_wrapper" id="delete___ENTRY_ID__" __DELETE_LISTENER__>
-                __DELETE_TEXT__
+            <div class="menu_wrapper" id="menu___ENTRY_ID__" __MENU_LISTENER__>
+               __MENU_TEXT__
```
CalvinKern, line 47: The menu is not accessible by screenreader (like and reply still work though). May just need an aria label?

libs/pandautils/src/main/assets/discussion_html_template_item.html
```diff
@@ -41,23 +41,17 @@
...
             </div>
 
             <div class="likes_wrapper___ENTRY_ID__" __LIKE_LISTENER__>
-                <div role="checkbox" aria-label="__LIKE_ARIA__" aria-checked="__LIKE_CHECKED_ARIA__" class="likes_icon_wrapper___ENTRY_ID__">
+                <div class="likes_count___ENTRY_ID__" aria-label="__LIKE_COUNT_ARIA__">__LIKE_COUNT__</div>
+                <div class="likes_icon_wrapper___ENTRY_ID__" role="checkbox" aria-label="__LIKE_ARIA__" aria-checked="__LIKE_CHECKED_ARIA__">
                     <img aria-hidden="true" class="likes_icon___ENTRY_ID__" src="__LIKE_ICON__">
                 </div>
-                <div class="likes_count___ENTRY_ID__" aria-label="__LIKE_COUNT_ARIA__">__LIKE_COUNT__</div>
```
CalvinKern, line 60: Lmao, I just noticed this was backwards on master when using the screenreader. Nice fix making those the correct order!

StephenBrough: Hah, yeah there's some quirkiness in the HTML here.

libs/pandautils/src/main/java/com/instructure/pandautils/discussions/DiscussionEntryHtmlConverter.kt
```diff
@@ -265,29 +258,23 @@ class DiscussionEntryHtmlConverter {
...
+    private fun formatMenuText(canEdit: Boolean, canDelete: Boolean): String {
+        if (canEdit && canDelete) {
+            return """
+            <div class="meatball_wrapper menu">
+                <div>
+                    <div class="meatball"></div>
+                    <div class="meatball"></div>
+                    <div class="meatball"></div>
+                </div>
+            </div>
```
CalvinKern, line 270: Not sure if this place or the other place is better for providing a talkback label, but leaving a comment to give you options.

StephenBrough:
> RTL looks good and marking unread works!
> 
> Just need that talkback working and we're golden.

I've found that on rotation, as well as any editing done to any other discussion entry, causes the entire webview to refresh, which means all the unread class labels get wiped out. 

Fixing this will be gross, we'll have to keep track of all the entry IDs that have been marked, then filter the HTMl we get back and add the appropriate class to the HTML elements. We'll probably want to make the unread call(s) when the user leaves the details screen as well to cut down on the number of API calls we make. 

I think for now this will be OK. Super annoying though.

CalvinKern:
Looks like the latest commit broke some stuff. All the reply buttons now say `Reply" role="button">` instead. Does it look this way for you?

I don't think you needed that aria label for reply, since it had text it was reading it fine before this. Now it looks very wrong.

Also, just noticed on a marshmallow device that the icon for editing in the material drawer isn't colored correctly. So that will need to be fixed as well.

StephenBrough:
> Looks like the latest commit broke some stuff. All the reply buttons now say `Reply" role="button">` instead. Does it look this way for you?

Hmm... that wasn't there when I built yesterday, but I see it now!

>Also, just noticed on a marshmallow device that the icon for editing in the material drawer isn't colored correctly. So that will need to be fixed as well.

I'm not sure which button you're referring to, there's no drawer involved (maybe you're talking about the bottom sheet?). Ran a Marshmallow emulator, the app bar Edit icon looks OK to me, and the bottom sheet Edit logo doesn't appear to be different from the other two. Maybe there's a difference on a physical device?

CalvinKern:
> I'm not sure which button you're referring to, there's no drawer involved (maybe you're talking about the bottom sheet?). Ran a Marshmallow emulator, the app bar Edit icon looks OK to me, and the bottom sheet Edit logo doesn't appear to be different from the other two. Maybe there's a difference on a physical device?

Yeah, sorry, I did mean the bottom sheet. I was also testing on an emulator. The edit button icon that was added is tinted white, so I can only see it when the ripple is present (hard to screenshot that). But this is what I see normally on the API 23 emulator.

Screen Shot 2020-05-20 at 8 57 06 AM

libs/pandautils/src/main/java/com/instructure/pandautils/discussions/DiscussionEntryHtmlConverter.kt
```diff
@@ -218,10 +215,9 @@ class DiscussionEntryHtmlConverter {
                 .replace("__ENTRY_ID__", discussionEntry.id.toString())
                 .replace("__USER_ID__", userId)
                 .replace("__REPLY_TEXT__", reply)
-                .replace("__EDIT_TEXT__", edit)
-                .replace("__DELETE_TEXT__", delete)
-                .replace("__EDIT_DIVIDER__", editDivider)
-                .replace("__DELETE_DIVIDER__", deleteDivider)
+                .replace("__REPLY_BUTTON_LABEL__", reply)
```
CalvinKern, line 218: Can remove the `__REPLY_BUTTON_LABEL` here since we're no longer using it in libs/pandautils/src/main/assets/discussion_html_template_item.html

apps/teacher/src/main/res/layout/bottom_sheet_discussion_menu.xml
```diff
@@ -0,0 +1,115 @@
...
+  ~     but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~     GNU General Public License for more details.
+  ~
+  ~     You should have received a copy of the GNU General Public License
+  ~     along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  ~
+  -->
+
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
CalvinKern, line 19: You'll need to take a look at this layout file again. The padding values don't match the designs in Figma.

CalvinKern:
Screen reader looks good, rtl looks good, behavior is good, and the weird marshmallow tint issue is good now!

Just need the layout updated to match designs.

apps/teacher/src/main/res/layout/bottom_sheet_discussion_menu.xml
CalvinKern: Here are the differences I see in Figma:
* paddingStart should only be 16 dp for each markAsUnread, edit, and delete.
* marginEnd should only be 32 dp for each image in markAsUnread, edit, and delete.

Everything else looks like it matches up (textviews and drawable resources match size/padding)

StephenBrough: Those might be old, here's what I'm seeing in Figma now:

Screen Shot 2020-05-20 at 3 39 13 PM
Screen Shot 2020-05-20 at 3 39 05 PM

StephenBrough: wtf, that 32 changes to 34 sometimes
Screen Shot 2020-05-20 at 3 40 28 PM

StephenBrough: Ah, I think I see the difference now. Depends on what you have selected. I remember seeing the 32 and 16 numbers. My eyes are so bad.

CalvinKern: The problem with Figma is that Dave puts those icons in a bounding box, so you have to make sure you click the whole box from the sidebar instead of just the icon.

Screen Shot 2020-05-20 at 3 41 15 PM
Screen Shot 2020-05-20 at 3 41 05 PM

PR #818, opened 2020-05-18
[Student][Pact][MBL-14235] | Add Pact tests for conversations/inbox

joehoag:
Also commented out a couple of unused model fields.

### DIFF
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Message.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Message.kt
@@ -35,7 +35,8 @@ data class Message(
         var attachments: List<Attachment> = ArrayList(),
         @SerializedName("media_comment")
         var mediaComment: MediaComment? = null,
-        var submission: Submission? = null,
+        // The submission field is not supported by the API, and not used in our code
+        //var submission: Submission? = null,
         @SerializedName("forwarded_messages")
         var forwardedMessages: List<Message> = ArrayList(),
         @SerializedName("participating_user_ids")
--- a/libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/InboxApiPactTests.kt
+++ b/libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/InboxApiPactTests.kt
@@ -38,6 +38,71 @@ class InboxApiPactTests : ApiPactTestBase() {
     }
 
     //
+    // region create a conversation
+    //
+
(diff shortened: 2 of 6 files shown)


libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Message.kt
```diff
@@ -35,7 +35,8 @@ data class Message(
         var attachments: List<Attachment> = ArrayList(),
         @SerializedName("media_comment")
         var mediaComment: MediaComment? = null,
-        var submission: Submission? = null,
+        // The submission field is not supported by the API, and not used in our code
+        //var submission: Submission? = null,
```
StephenBrough, line 39: Huh. I wonder if this is a really old artifact of some past feature we no longer support. I know we had the conversation of leaving these in, commented out, but I hope if we move to Flutter we can filter these out. Maybe put them in a doc somewhere if we're really worried about unknowns.

joehoag: IMO, this constitutes the doc.  If we move to flutter, the dev can look at this field and decide whether or not to move it over (commented-out).

It's kind of strange that the API docs still mention a submission as being part of a conversation, but always along with a message that it is obsolete.

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/InboxApiPactTests.kt
```diff
@@ -0,0 +1,361 @@
...
+    //
+
+    val getAllConversationsQuery = "interleave_submissions=1&include[]=participant_avatars&scope="
+    val getAllConversationsPath = "/api/v1/conversations"
+    val getAllConversationsFieldConfig = PactConversationFieldConfig()
+    val getAllConversationsResponseBody = LambdaDsl.newJsonArray { arr ->
+        arr.`object`() { obj ->
+            obj.populateConversationFields(getAllConversationsFieldConfig)
+        }
+        arr.`object`() { obj ->
```
StephenBrough, line 116: Is this just creating a duplicate entry in the array?

joehoag: It's creating two separate conversation objects in the returned List<Conversation>.

joehoag: I think that it's possible to say, "Expect one or more conversation objects in the returned list/array", but I've run into trouble with that before.  I'll give that another try in future pact efforts.

PR #819, opened 2020-05-18
[Student][Teacher][MBL-12556] Push notification updates

CalvinKern:
This also removed support for old notifications that were stored using the user id and domain prior to MBL-13725. Removing here as we have now released two versions of teacher/student that should have put everyone on the new format for storing notifications.

This commit will now show each push notification received as its own notification on the device. On devices higher than API 23 (Android 6.0) we will always show each push as its own notification, as groups aren’t supported very well on this API version. Higher than 23 will collapse 2 or more notifications into a notification group. This group can be expanded letting users deal with each notification themselves. Clearing the group will clear all notifications.

The JIRA ticket allows for extra customizations if any were desired beyond grouping and expanding, though I couldn't think of anything with our current notification data. When we eventually get server work done, we can add different channels (to match the channels/categories in Canvas), actions (reply to inbox message), and more!

------To test------
Have the current master branch installed on various API levels (23 is our min, 29 is our max).
Have at least 2 notifications on the device.
Install this branch as an update to keep notifications.

* Verify old notifications persist on the update, and are now each their own notification
* Verify you can click a notification to have it open the app to the correct place (and is cleared from the notification list)
* Verify you can swipe to dismiss a notification and restarting the device doesn't show it again
* Verify that using the 'clear all' button on newer Android versions will clear all notifications, and restarting the device doesn't show it again
* Verify new notifications appear and behave as stated above

### DIFF
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/receivers/PushExternalReceiver.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/receivers/PushExternalReceiver.kt
@@ -137,7 +137,7 @@ abstract class PushExternalReceiver : FirebaseMessagingService() {
             channelId: String,
             notification: PushNotification
         ): Notification {
-            // Use the app name is we're marshmallow or lower, as it won't display the app name for us. On later versions,
+            // Use the app name if we're marshmallow or lower, as it won't display the app name for us. On later versions,
             // we'll still want a title, but it will already have our app name so instead we'll use this notification title
             val title =
                 if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.M) appName else context.getString(R.string.notificationPrimaryInboxTitle)
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/receivers/PushExternalReceiver.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/receivers/PushExternalReceiver.kt
@@ -47,22 +47,25 @@ abstract class PushExternalReceiver : FirebaseMessagingService() {
         val data = message?.data ?: return
 
         val from = message.from ?: ""
-        val alert = getMessage(data)
-        val userId = getUserId(data)
-        val htmlUrl = getHtmlUrl(data)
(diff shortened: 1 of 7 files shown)


libs/pandautils/src/main/java/com/instructure/pandautils/receivers/PushExternalReceiver.kt
```diff
@@ -47,110 +47,153 @@ abstract class PushExternalReceiver : FirebaseMessagingService() {
...
-            nm.notify(PushNotification.NOTIFY_ID, notification)
+        private fun createNotification(
+            context: Context,
+            startingActivity: Class<out Activity>,
+            appName: String,
+            @ColorRes appColor: Int,
+            channelId: String,
+            notification: PushNotification
+        ): Notification {
+            // Use the app name is we're marshmallow or lower, as it won't display the app name for us. On later versions,
```
JordanMarshall, line 140: >Use the app name _if_ we're marshmallow or lower

libs/pandautils/src/main/res/values/strings.xml
```diff
@@ -387,6 +387,10 @@
     <string name="notificationPrimaryInboxTitle">Canvas Notification</string>
     <string name="notificationChannelNamePrimary">General Canvas Notifications</string>
     <string name="notificationChannelDescriptionPrimary">Further configuration of notifications can be done within the Canvas Notification Preferences section.</string>
+    <plurals name="notificationGroupContent">
+        <!--Will always be 2 or more, but some languages to specific formatting for 2 items-->
```
JordanMarshall, line 391: > but some languages _require_ specific formatting

libs/pandautils/src/main/java/com/instructure/pandautils/receivers/PushExternalReceiver.kt
CalvinKern: Oof, both of these comments are going to bother me. PATCH INCOMING!

PR #821, opened 2020-05-20
Update translations

inst-danger:


CalvinKern:
This was a test to make sure that the script for extracting translations for flutter projects works as intended.
Closing as this will get run again later from bitrise (without my l10n changes as well)

PR #822, opened 2020-05-20
[Student][Teacher][MBL-14322] Html updates for newrce videos (and more)

Blackchai:
Okay, so the first thing you'll notice is that this touches a lot of files. A pretty hefty amount of that is me refactoring a couple of our CanvasWebView html helper functions (unfortunately a lot of these are in the old parent app, so its just a method name refactor for most of those). We had two functions called formatHtml and a function called loadHtml. They were a sort of weird mismatch of each other, so I made it so we have just one loadHtml() and one formatHtml(). loadHtml() formats and loads, formatHtml() just formats and returns as a string.

Additionally, I combined a few of our helper functions for parsing out iframes, cnvs_content ids, and lti src urls into a single function, called loadHtmlWithiFrames. This has also been added to a decent number of locations with RCE content that didn't previously support it (discussions/quizzes/assignments in student being the biggest areas).

To Test:
There are a few old pages/etc with LTI content in the Android Dev course, I've been using those to check backwards compat with old rce content and lti embeds, all looks well there so far. I've also been using the new rce content to add videos in all of the other content types for testing. Let me know if you have any questions!


apps/student/src/main/java/com/instructure/student/fragment/DiscussionDetailsFragment.kt
```diff
@@ -665,11 +668,11 @@ class DiscussionDetailsFragment : ParentFragment(), Bookmarkable {
...
-        if (CanvasWebView.containsStudioLTI(discussionTopicHeader.message.orEmpty(), "UTF-8")) {
-            // We are only handling Studio because there is not a predictable way for use to determine if a URL is and LTI launch
-            getAuthenticatedURL(discussionTopicHeader.message.orEmpty()) { authenticatedHtml, originalUrl -> loadHTMLTopic(authenticatedHtml, originalUrl) }
+        if (discussionTopicHeader.message?.contains("<iframe") == true) {
+            loadHeaderHtmlJob = discussionTopicHeaderWebView.loadHtmlWithIframes(requireContext(), isTablet,
+                    discussionTopicHeader.message.orEmpty(), ::loadHTMLTopic, discussionTopicHeader.title)
         } else {
-            loadHTMLTopic(discussionTopicHeader.message ?: "")
+            loadHTMLTopic(discussionTopicHeader.message.orEmpty(), discussionTopicHeader.title)
         }
```
CalvinKern, line 676: You've done this "<iframe" check in a few places now. Would it be reasonable to put that inside the loadHtmlWithIframes call? Then it could return the job if it needs to, otherwise it will just call the callback provided (essentially what the else cases are doing).

Not sure if they would help clean up, or if it would break somewhere else, but would reduce the number of places that have to do that iframe check.

apps/student/src/main/java/com/instructure/student/fragment/DiscussionDetailsFragment.kt
```diff
@@ -524,12 +527,12 @@ class DiscussionDetailsFragment : ParentFragment(), Bookmarkable {
     //endregion
 
     //region Loading
-    private fun loadHTMLTopic(html: String, ltiUrl: String? = null) {
+    private fun loadHTMLTopic(html: String, contentDescription: String?) {
         setupHeaderWebView()
-        discussionTopicHeaderWebView.loadHtml(DiscussionUtils.createDiscussionTopicHeaderHtml(requireContext(), isTablet, html, ltiUrl), discussionTopicHeader.title)
+        discussionTopicHeaderWebView.loadHtml(html, contentDescription)
```
CalvinKern, line 532: Do we not need to createDiscussionTopicHeaderHtml anymore?

apps/student/src/main/java/com/instructure/student/fragment/QuizStartFragment.kt
```diff
@@ -65,6 +66,7 @@ class QuizStartFragment : ParentFragment(), Bookmarkable {
 
     private var webViewClientCallback: CanvasWebView.CanvasWebViewClientCallback? = null
     private var embeddedWebViewCallback: CanvasWebView.CanvasEmbeddedWebViewCallback? = null
+    private var loadHtmlJob: Job? = null
```
CalvinKern, line 69: Need a cancel for this job.

apps/teacher/src/main/java/com/instructure/teacher/fragments/AssignmentDetailsFragment.kt
```diff
@@ -280,19 +280,19 @@ class AssignmentDetailsFragment : BasePresenterFragment<
...
 
         // If the html has a Studio LTI url, we want to authenticate so the user doesn't have to login again
-        if (CanvasWebView.containsLTI(description.orEmpty(), "UTF-8")) {
+        if (description?.contains("<iframe") == true) {
             descriptionWebView.addJavascriptInterface(JsExternalToolInterface {
                 val args = LTIWebViewFragment.makeLTIBundle(URLDecoder.decode(it, "utf-8"), requireContext().getString(R.string.utils_externalToolTitle), true)
                 RouteMatcher.route(requireContext(), Route(LTIWebViewFragment::class.java, mCourse, args))
             }, "accessor")
-            loadHtmlJob = descriptionWebView.loadHtmlWithLTIs(requireContext(), isTablet, description.orEmpty(), ::loadAssignmentHTML)
+            loadHtmlJob = descriptionWebView.loadHtmlWithIframes(requireContext(), isTablet, description.orEmpty(), ::loadAssignmentHTML, mAssignment.name)
```
CalvinKern, line 288: Do you want to use `mAssignment.name` here? I think it's technically fine, but you could just use `name` since your in a `with` on the assignment.

apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionsDetailsFragment.kt
```diff
@@ -404,23 +404,24 @@ class DiscussionsDetailsFragment : BasePresenterFragment<
...
                 val args = LTIWebViewFragment.makeLTIBundle(URLDecoder.decode(it, "utf-8"), getString(R.string.utils_externalToolTitle), true)
                 RouteMatcher.route(this@DiscussionsDetailsFragment.requireContext(), Route(LTIWebViewFragment::class.java, canvasContext, args))
             }, "accessor")
-            headerLoadHtmlJob = discussionTopicHeaderWebView.loadHtmlWithLTIs(requireContext(), isTablet, discussionTopicHeader.message.orEmpty(), this::loadHTMLTopic)
+            headerLoadHtmlJob = discussionTopicHeaderWebView.loadHtmlWithIframes(requireContext(), isTablet,
+                    discussionTopicHeader.message.orEmpty(), this::loadHTMLTopic, presenter.discussionTopicHeader.title)
         } else {
-            discussionTopicHeaderWebView.loadHtml(discussionTopicHeader.message, discussionTopicHeader.title)
+            loadHTMLTopic(discussionTopicHeader.message.orEmpty(), discussionTopicHeader.title)
         }
```
CalvinKern, line 416: Similar question here, do you want to use `presenter.discussionTopicHeader.title` here instead of the passed in `discussionTopicHeader.title`?

apps/teacher/src/main/java/com/instructure/teacher/fragments/QuizDetailsFragment.kt
```diff
@@ -363,15 +364,16 @@ class QuizDetailsFragment : BasePresenterFragment<
...
 
         // Load instructions
-        if (CanvasWebView.containsLTI(quiz.description ?: "", "UTF-8")) {
-            loadHtmlJob = instructionsWebView.loadHtmlWithLTIs(requireContext(), isTablet, quiz.description ?: "", ::loadQuizHTML)
+        if (quiz.description?.contains("<iframe") == true) {
+            loadHtmlJob = instructionsWebView.loadHtmlWithIframes(requireContext(), isTablet,
+                    quiz.description.orEmpty(), ::loadQuizHTML, presenter.mQuiz.title)
         } else {
-            loadQuizHTML(quiz.description ?: "")
+            loadQuizHTML(quiz.description.orEmpty(), presenter.mQuiz.title)
```
CalvinKern, line 371: Another one of those cases where you're using presenter instead of the passed object.
Again, these all are probably fine technically, but it's still weird to be using the passed in object and the presenter object in the same lines.

libs/pandautils/src/main/java/com/instructure/pandautils/utils/WebViewExtensions.kt
```diff
@@ -80,16 +86,46 @@ fun WebView.loadHtmlWithLTIs(context: Context, isTablet: Boolean, html: String,
...
+    val button = "</br><p><div class=\"lti_button\" onClick=\"onLtiToolButtonPressed('%s')\">%s</div></p>"
+    val htmlButton = String.format(button, ltiUrl, context.resources.getString(R.string.utils_launchExternalTool))
+
+    // Now we add the launch button along with the new iframe with the updated URL
+    return newIframe + htmlButton
+}
+
+fun newRceVideoElement(dataMediaId: String): String {
+    // We need to make a new src url with the dataMediaId
+    val newSrcUrl = "/users/self/media_download?entryId=$dataMediaId&media_type=video&redirect=1"
```
CalvinKern, line 115: Do we know that these RCE media_objects_iframe urls will always be for media_type video?

apps/student/src/main/java/com/instructure/student/fragment/DiscussionDetailsFragment.kt
Blackchai: Not a bad idea. At the very least I could make it another helper function, to prevent potential typos/etc.

apps/student/src/main/java/com/instructure/student/fragment/DiscussionDetailsFragment.kt
Blackchai: Oh doh, I had been planning to clean up the now un-needed functions in the Discussionutils. I'll do that, thanks!

apps/student/src/main/java/com/instructure/student/fragment/QuizStartFragment.kt
Blackchai: Damn, can't believe I missed one!

apps/teacher/src/main/java/com/instructure/teacher/fragments/AssignmentDetailsFragment.kt
Blackchai: Oh very true, I think it was just leftover from a copy pasta.

apps/teacher/src/main/java/com/instructure/teacher/fragments/DiscussionsDetailsFragment.kt
Blackchai: Copy pasta, I'll check!

apps/teacher/src/main/java/com/instructure/teacher/fragments/QuizDetailsFragment.kt
Blackchai: :+1:

libs/pandautils/src/main/java/com/instructure/pandautils/utils/WebViewExtensions.kt
Blackchai: Pretty confident, but that's a good question. I'll do some investigation.

apps/student/src/main/java/com/instructure/student/fragment/DiscussionDetailsFragment.kt
CalvinKern: It seemed like every place could do this. It will just require being even more relaxed on the javascript channels (which is currently more relaxed than before this change; before the channels were only registered if there was an `external_tools`, now they're registered if any iframe is in the content. 

May as well just do always so that this callback can be super clean, imo.

libs/pandautils/src/main/java/com/instructure/pandautils/utils/WebViewExtensions.kt
Blackchai: Just tested with all of the available types of rce video/audio, it's working for all of them, including things like assignments with embedded course file videos.

apps/student/src/main/java/com/instructure/student/fragment/DiscussionDetailsFragment.kt
CalvinKern: Nice idea adding the javascript interface callback to that method! That makes it way better!

Everything looks good, just need to test now. Will do first thing tomorrow!

PR #823, opened 2020-05-20
[Student][Parent][MBL-14395] Flutter l10n work

CalvinKern:
Fix for generated language files in Parent. In a previous commit, I renamed the messages_pt_PT and messages_zh_HK files to fix those locales not working with what Canvas servers give us. This is incorrect as those are generated files and the fix has to be handled specially (in parent_app). Reverting those file names and handling those special cases so our generated workflow still goes smoothly.

Also added support for flutter_student_embed (copying over some of the l10n fixes that have been done in parent). The last commit adds some dummy translations for Portuguese and Chinese to student to make testing easier. These should be fine to commit (they'll get overridden by real translations at some point) but if you'd like me to remove them before merging this I can.

See also: The bitrise project `canvas-android-tasks` now has a new workfflow `extract-flutter-for-translators` that will extract strings from each flutter projects AppLocalizations into the english arb files to be exported to S3 in the weekly export job.

-----To Test-----
In flutter-parent, verify the app shows correctly after changing your canvas user's language to Portuguese and both traditional/simplified Chinese. Had to revert part of a change from MBL-14357 to keep the language fix working after our resources have been generated again

In student, you should be able to see the "Calendar" toolbar title change for Portuguese (CalendarPT) and Chinese (simplified is CalendarZH and traditional is CalendarHK). To get a locale change from canvas recognized in the student app I had to change users and select the same user again. This does cause the app to crash when using a local build, but using a bitrise apk doesn't have this issue.


JordanMarshall:
As part of updating to Flutter 1.17, we had to disable some tests in `calendar_day_of_week_headers_test.dart` and `calendar_widget_test.dart` in Student due to l10n issues. Do you think the changes here will fix those tests, or would it be better address them in a separate commit?

CalvinKern:
> Do you think the changes here will fix those tests, or would it be better address them in a separate commit?

Looks like this does fix it! I'll uncomment those tests and see if they pass on bitrise now too.

PR #825, opened 2020-05-21
[Student][MBL-14353] Allow file uploads for question bank quizzes

StephenBrough:
See ticket for repro steps and creds. 

I have also setup a test quiz in my sandbox for april, if you don't want to be submitting to someone else's sandbox. 

Side note: I have no idea where the files go. I couldn't find them in the user files, nor the course files. Must be some special place for quiz file uploads.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/fragment/BasicQuizViewFragment.kt
+++ b/apps/student/src/main/java/com/instructure/student/fragment/BasicQuizViewFragment.kt
@@ -17,6 +17,7 @@
 
 package com.instructure.student.fragment
 
+import android.content.Intent
 import android.graphics.Bitmap
 import android.net.Uri
 import android.os.Bundle
@@ -24,6 +25,7 @@ import android.view.View
 import android.webkit.WebResourceRequest
 import android.webkit.WebView
 import android.webkit.WebViewClient
+import android.widget.Toast
 import com.instructure.canvasapi2.managers.OAuthManager
 import com.instructure.canvasapi2.managers.QuizManager
 import com.instructure.canvasapi2.models.*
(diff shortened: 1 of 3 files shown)


apps/student/src/main/java/com/instructure/student/fragment/BasicQuizViewFragment.kt
```diff
@@ -120,6 +128,44 @@ class BasicQuizViewFragment : InternalWebviewFragment() {
         }
     }
 
+    private fun setupFilePicker() {
+        getCanvasWebView()?.setCanvasWebChromeClientShowFilePickerCallback(object : CanvasWebView.VideoPickerCallback {
```
joehoag, line 132: So I guess a VideoPickerCallback can be used to choose any arbitrary file, and is not constrained to just picking video files?

apps/student/src/main/java/com/instructure/student/fragment/BasicQuizViewFragment.kt
```diff
@@ -120,6 +128,44 @@ class BasicQuizViewFragment : InternalWebviewFragment() {
...
+            PermissionUtils.makeArray(PermissionUtils.WRITE_EXTERNAL_STORAGE, PermissionUtils.CAMERA),
+            PermissionUtils.PERMISSION_REQUEST_CODE
+        )
+    }
+
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    fun onRequestPermissionsResult(result: PermissionRequester.PermissionResult) {
+        if (PermissionUtils.allPermissionsGrantedResultSummary(result.grantResults)) {
+            getCanvasWebView()?.clearPickerCallback()
+            Toast.makeText(requireContext(), R.string.pleaseTryAgain, Toast.LENGTH_SHORT).show()
```
joehoag, line 159: Under what conditions does this logic get called?  I had to grant a permission when I tested this, and I never saw the pleaseTryAgain message.

apps/student/src/main/java/com/instructure/student/fragment/BasicQuizViewFragment.kt
StephenBrough: Yeah, I think it's purpose has expanded since it was originally written.

apps/student/src/main/java/com/instructure/student/fragment/BasicQuizViewFragment.kt
StephenBrough: The NavigationActivity was catching the callback for me and I was sending out a bus event and catching it here. Might be something going on with different Android versions and how they handle sending system callbacks through the fragment/activity stack.

PR #827, opened 2020-05-21
[Parent][E2E][MBL-14433]: Fix reporting of retries in E2E tests

joehoag:
Should fix the message-formatting problems that prevented retries from being reported before.

### DIFF
--- a/apps/flutter_parent/run_all_ui_tests.bash
+++ b/apps/flutter_parent/run_all_ui_tests.bash
@@ -50,8 +50,11 @@ do
         # Allow for a single retry for a failed test
         if [ $? -ne 0 ]
         then
+          # Stop the clock
+          endTime=`date +"%s"`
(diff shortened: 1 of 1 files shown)


apps/flutter_parent/run_all_ui_tests.bash
```diff
@@ -50,8 +50,11 @@ do
         # Allow for a single retry for a failed test
         if [ $? -ne 0 ]
         then
+          # Stop the clock
+          endTime=`date +"%s"`
+          ((runSecs=endTime-startTime))
           echo "Aggregator: $driver failed; retrying..."
-          emitTestResult $driver retry
+          emitTestResult $driver retry $runSecs
```
joehoag, line 57: Omitting the runSecs parameter was causing Splunk retry events to be formatted incorrectly and therefore not properly transmitted.

PR #836, opened 2020-05-29
Extract translations

inst-danger:


JordanMarshall:
Only timestamps were changed, no need to merge this.

PR #838, opened 2020-06-02
[Parent][MBL-14388] Reply all fix

Blackchai:
Okay, so this got a lot grosser than I had hoped. In order to filter out improper recipients from all usages of reply all, we have to make a couple of api calls. I'm open to feedback and suggestions on this. For now, it's working and tested.

To Test (I found creating the repro case for this was easier from web):
-Create a message as a teacher in a course with an observer, teachers/tas, and some students
-Reply All to this message from the current parent app, notice that it replies to all, including observed and non-observed students
-This change will filter out all non-observed students, and ALL students / self if the "send messages" permission is off

### DIFF
--- a/apps/flutter_parent/lib/screens/inbox/reply/conversation_reply_interactor.dart
+++ b/apps/flutter_parent/lib/screens/inbox/reply/conversation_reply_interactor.dart
@@ -43,8 +43,6 @@ class ConversationReplyInteractor {
         recipientIds = [replyMessage.authorId];
       }
     } else {
-      recipientIds = replyMessage.participatingUserIds.toList();
-
       // We need to make sure the recipients list doesn't contain any off limit users, such as non-observed students.
       final courseId = conversation.getContextId();
       final userId = ApiPrefs.getUser().id;
--- a/apps/flutter_parent/lib/screens/inbox/reply/conversation_reply_interactor.dart
+++ b/apps/flutter_parent/lib/screens/inbox/reply/conversation_reply_interactor.dart
@@ -15,6 +15,8 @@
 import 'package:flutter/widgets.dart';
 import 'package:flutter_parent/models/conversation.dart';
 import 'package:flutter_parent/models/message.dart';
+import 'package:flutter_parent/network/api/course_api.dart';
+import 'package:flutter_parent/network/api/enrollments_api.dart';
 import 'package:flutter_parent/network/api/inbox_api.dart';
 import 'package:flutter_parent/network/utils/api_prefs.dart';
 import 'package:flutter_parent/screens/inbox/attachment_utils/attachment_handler.dart';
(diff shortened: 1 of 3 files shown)


apps/flutter_parent/lib/screens/inbox/reply/conversation_reply_interactor.dart
```diff
@@ -40,8 +42,37 @@ class ConversationReplyInteractor {
       } else {
         recipientIds = [replyMessage.authorId];
       }
-    } else if (message != null) {
+    } else {
       recipientIds = replyMessage.participatingUserIds.toList();
```
joehoag, line 46: The setting of recipientIds here seems superfluous.  In fact, a good compiler should ignore this line.

apps/flutter_parent/lib/screens/inbox/reply/conversation_reply_interactor.dart
```diff
@@ -40,8 +42,37 @@ class ConversationReplyInteractor {
...
+        if (enrollments == null) return false;
+        return enrollments.contains('TeacherEnrollment') || enrollments.contains('TaEnrollment');
+      });
+
+      final filteredRecipientIds = recipients.map<String>((recipient) => recipient.id);
+
+      recipientIds = replyMessage.participatingUserIds
+          .toList()
+          .where((participantId) => filteredRecipientIds.contains(participantId))
+          .toList();
```
joehoag, line 75: You're right -- this is pretty complicated logic!  But I wasn't able to find any holes in it.

joehoag:
I've tested this out and it seems OK.  One thing to note: Reply-all will only send to a TA if that TA is in the same section as the observed student.  Seems OK to me, but I just wanted to run it by you.

apps/flutter_parent/lib/screens/inbox/reply/conversation_reply_interactor.dart
Blackchai: How so? If this line isn't present, recipientIds will be [] from line 35.

apps/flutter_parent/lib/screens/inbox/reply/conversation_reply_interactor.dart
Blackchai: Thanks!

Blackchai:
> I've tested this out and it seems OK. One thing to note: Reply-all will only send to a TA if that TA is in the same section as the observed student. Seems OK to me, but I just wanted to run it by you.

I'll look into this case some more! Thanks for all the testing.

apps/flutter_parent/lib/screens/inbox/reply/conversation_reply_interactor.dart
Blackchai: OH... I'm setting the list again down below 🤦 . I get it, I'll remove this line.

PR #842, opened 2020-06-08
[Student][MBL-14378] Fix flutter appbar style for landscape and tablets

JordanMarshall:
Flutter's AppBar widget always uses the same fixed height, causing a height (and style) mismatch when compared to the native Toolbar in landscape mode and on tablets. This commit wraps the AppBars and enforces size and style changes to match the native app. Specifically, on tablets the height is extended to 64 and extra padding is applied, while in landscape the height is shortened to 48 and the title font size is reduced to 14.


JordanMarshall:
Nice catch; my sandbox uses white text so that slipped by me in testing.

PR #853, opened 2020-06-12
[Student][MBL-14398] Update for student app observer assignment deets

Blackchai:
To Test:
-Log into the student app with an observer account with a submission with some status (graded, etc).
-Note that the assignment details screen now displays all the correct submission details for your first observer (no other way to "pick" observers right now in student, we just default to #1).

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsEffectHandler.kt
+++ b/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsEffectHandler.kt
@@ -153,10 +153,23 @@ class AssignmentDetailsEffectHandler(val context: Context, val assignmentId: Lon
                 if(courseResult.isSuccess && courseResult.dataOrNull != null) {
                     val enrollment = courseResult.dataOrNull!!.enrollments!!.firstOrNull { it.isObserver }
 
-                    val assignmentResponse = awaitApiResponse<Assignment> {
-                        AssignmentManager.getAssignment(effect.assignmentId, effect.courseId, effect.forceNetwork, it)
+                    if(enrollment != null) {
+                        val assignmentResponse = awaitApiResponse<ObserverAssignment> {
+                            AssignmentManager.getAssignmentForObserver(effect.assignmentId, effect.courseId, effect.forceNetwork, it)
+                        }
+                        val assignmentWithObserverSubmission = assignmentResponse.body()!!.toAssignmentForObservee()
+                        if(assignmentWithObserverSubmission != null) {
+                            DataResult.Success(assignmentWithObserverSubmission)
+                        } else {
+                            DataResult.Fail(null)
+                        }
+                    } else {
+                        val assignmentResponse = awaitApiResponse<Assignment> {
+                            AssignmentManager.getAssignment(effect.assignmentId, effect.courseId, effect.forceNetwork, it)
+                        }
+                        DataResult.Success(assignmentResponse.body()!!)
                     }
-                    DataResult.Success(assignmentResponse.body()!!)
+
                 } else {
                     DataResult.Fail(null)
                 }
--- a/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsEffectHandler.kt
+++ b/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsEffectHandler.kt
@@ -40,8 +40,8 @@ import com.squareup.sqldelight.Query
 import kotlinx.coroutines.launch
 
 class AssignmentDetailsEffectHandler(val context: Context, val assignmentId: Long) :
-        EffectHandler<AssignmentDetailsView, AssignmentDetailsEvent, AssignmentDetailsEffect>(),
-        Query.Listener {
+    EffectHandler<AssignmentDetailsView, AssignmentDetailsEvent, AssignmentDetailsEffect>(),
+    Query.Listener {
 
     private var submissionQuery: Query<Submission>? = null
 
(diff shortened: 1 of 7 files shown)


libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/ObserverAssignment.kt
```diff
@@ -0,0 +1,150 @@
...
+        @SerializedName("anonymous_grading")
+        val anonymousGrading: Boolean = false,
+        @SerializedName("allowed_attempts")
+        val allowedAttempts: Long = -1, // API gives -1 for unlimited submissions
+        var isStudioEnabled: Boolean = false
+) : CanvasModel<Assignment>() {
+    override val comparisonDate get() = dueAt.toDate()
+    override val comparisonString get() = dueAt
+
+    fun toAssignmentForObservee(): Assignment? {
```
joehoag, line 100: So, what does this accomplish?  Only return an Assignment object if a submission has occurred?  

Is there any difference, field-wise, between ObserverAssignment() and the object returned by toAssignmentForObservee()?

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsEffectHandler.kt
```diff
@@ -138,11 +140,38 @@ class AssignmentDetailsEffectHandler(val context: Context, val assignmentId: Lon
...
+                            AssignmentManager.getAssignmentForObserver(effect.assignmentId, effect.courseId, effect.forceNetwork, it)
+                        }
+                        val assignmentWithObserverSubmission = assignmentResponse.body()!!.toAssignmentForObservee()
+                        if(assignmentWithObserverSubmission != null) {
+                            DataResult.Success(assignmentWithObserverSubmission)
+                        } else {
+                            DataResult.Fail(null)
+                        }
+                    } else {
+                        // No observer enrollment, business as usual
```
joehoag, line 166: Not *quite* business as usual.  This logic used to run regardless of whether or not getCourseWithGradeAsync() call succeeded.  Is this difference intentional?

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/ObserverAssignment.kt
```diff
@@ -0,0 +1,150 @@
...
+ *     limitations under the License.
+ */
+package com.instructure.canvasapi2.models
+
+import com.google.gson.annotations.SerializedName
+import com.instructure.canvasapi2.utils.toDate
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
+data class ObserverAssignment(
```
joehoag, line 23: Maybe a comment about why we need ObserverAssignment, and when to use it instead of Assignment?  And I'm assuming ObserverAssignment means "view of assignment given to observer", not that the assignment itself was assigned to an observer?

Blackchai: Yeah some documentation would probably be a good idea here, I can add that.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsEffectHandler.kt
Blackchai: Yeah it was intentional. Thinking on it now though, I suppose I could just fall back to the default calls even if the course call fails. My original thinking was that if we can't fetch a course, things are likely broken anyways and can't be handled properly for the observer case. But the default case is more important. I'll tweak this.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/ObserverAssignment.kt
Blackchai: Yeah so this just keeps the rest of assignment details simple. ObserverAssignment has a list of submissions, Assignment has a single submission, unfortunately, the api uses the same property for both, so I had to dupe the objects. This helper method just converts an ObserverAssignment to an Assignment, returning null if the submission list is empty just helps with error handling, as we want to display an error state in that case. I'll add some comments.

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/AssignmentDetailsEffectHandler.kt
```diff
@@ -210,14 +204,14 @@ class AssignmentDetailsEffectHandler(val context: Context, val assignmentId: Lon
...
             consumer.accept(
-                AssignmentDetailsEvent.DataLoaded(
-                    assignmentResult,
-                    isStudioEnabled,
-                    studioLTITool,
-                    ltiTool,
-                    dbSubmission,
-                    quizResult
-                )
+                    AssignmentDetailsEvent.DataLoaded(
```
Blackchai, line 207: Ugh, I did an auto reformat, not sure why it added these indents.

Blackchai:
> Looks good -- thanks for adding the extra comments and restoring some default functionality. I re-tested, and it looks like the right things are displayed in assignment details for both observer and student.
> 
> It's kind of a shame that we need to have two sizable, largely identical classes to represent assignments in this fashion. It seems like there ought to be a way to have one derive from the other, or have both derive from a base class with common fields. At first glance, this is not allowed for data classes in Kotlin, but maybe we'll find a way to do it down the road.

Yeah I agree, its a shame. I was able to work around this in flutter with a custom serializer for this field. I might be able to do the same in the future to replace this, but it exceeded the scope of this ticket. Thanks for the good feedback!

PR #865, opened 2020-06-29
[Student][Teacher][MBL-14340] | Pact tests for discussions

joehoag:
This ended up being somewhat smaller in scope than I originally imagined, because:
(1) The server-side pact logic was so tortuously hard to manipulate, and
(2) Andrew had already written some tests for other discussion ops (i.e., create a discussion, reply to a discussion message, etc...)


joehoag:
> LGTM. I wouldn't mind seeing those unused model fields removed entirely if you wanted to do that, but no biggie either way.

That might be cleaner, but I think that we agreed to keep them around as a way to denote "We're aware that these fields are part of the API, but we're not using them for now."

Thanks for the review.

PR #866, opened 2020-06-30
[Student][MBL-14324] Set course ID on window ENV object for Pages

JordanMarshall:


JordanMarshall:
There shouldn't be any noticeable behavior changes unless the page has javascript that specifically looks for the course id; as far as I know we don't have a real-world example to test against. iOS used web devtools to test their changes for this and I did something similar.

joehoag:
👍

PR #869, opened 2020-07-01
[Student][MBL-11503] Only show FAB in People details when message permission is granted

JordanMarshall:


apps/student/src/main/java/com/instructure/student/fragment/PeopleDetailsFragment.kt
```diff
@@ -123,7 +122,24 @@ class PeopleDetailsFragment : ParentFragment(), Bookmarkable {
...
+    }
+
+    private fun checkMessagePermission() {
+        permissionCall = tryWeave {
+            val id = canvasContext.id
+            val canMessageUser = when {
+                canvasContext.isGroup -> GroupManager.getPermissionsAsync(id).awaitOrThrow().send_messages
+                canvasContext.isCourse -> {
+                    val isTeacher = user?.enrollments?.any { it.courseId == id && (it.isTA || it.isTeacher) } ?: false
+                    isTeacher || CourseManager.getPermissionsAsync(canvasContext.id).awaitOrThrow().send_messages
```
Blackchai, line 136: Ugh, this is such a nitpick, but you could have re-used your id here for the CourseManager.getPermissionsAsync() call..

PR #877, opened 2020-07-07
[Parent][MBL-14287] QR Account creation chooser

Blackchai:
Adds the bottom sheet picker and modifies the QRPairingScreen to prep it for routing to account creation (coming soon **TM**).

### DIFF
--- a/apps/flutter_parent/lib/router/panda_router.dart
+++ b/apps/flutter_parent/lib/router/panda_router.dart
@@ -123,7 +123,8 @@ class PandaRouter {
 
   static String _qrPairing = '/qr_pairing';
 
-  static String qrPairing({String pairingUri}) {
+  static String qrPairing({String pairingUri, bool isCreatingAccount = false}) {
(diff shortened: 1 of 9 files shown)


apps/flutter_parent/lib/router/panda_router.dart
```diff
@@ -123,7 +123,8 @@ class PandaRouter {
 
   static String _qrPairing = '/qr_pairing';
 
-  static String qrPairing({String pairingUri}) {
+  static String qrPairing({String pairingUri, bool isCreatingAccount = false}) {
+    if (isCreatingAccount) return '$_qrPairing?${_RouterKeys.isCreatingAccount}=${isCreatingAccount ? 1 : 0}';
```
JordanMarshall, line 127: Is the ternary necessary here? You're already checking for `isCreatingAccount` so I imagine it will always evaluate to 1.

Blackchai: Oh durp, good point.

PR #879, opened 2020-07-08
[Student][MBL-14495] Remove native quiz support

JordanMarshall:


libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/apis/QuizAPI.kt
```diff
@@ -67,100 +64,14 @@ object QuizAPI {
         @GET("courses/{courseId}/quizzes/{quizId}/submissions")
         fun getFirstPageQuizSubmissions(@Path("courseId") courseId: Long?, @Path("quizId") quizId: Long?): Call<QuizSubmissionResponse>
```
joehoag, line 65: Would it be fair to say that you preserved some of these endpoints just to ensure that the associated Pact tests still built successfully?  Because it seems odd to preserve this endpoint while removing getNextPageQuizSubmissions() and QuizManager.getQuizSubmissions().

If that is the case, we should make a list of such "vestigial" endpoints, and I can make a second pass to remove them and their associated Pact tests.

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/QuizzesE2ETest.kt
```diff
@@ -1,144 +0,0 @@
-/*
```
joehoag, line 1: Ouch -- it hurts that this entire e2e test had to be deleted.  Was there nothing that could be preserved here?

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/QuizInteractionTest.kt
```diff
@@ -1,541 +0,0 @@
-/*
```
joehoag, line 1: Ouch again.  Did this whole test have to be deleted?  I understand that the testOpensNativeView_xxx tests were no longer relevant.  But could any of the other ones have been preserved?

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/QuizzesE2ETest.kt
JordanMarshall: Hmmm, I suppose we could keep the first part—verifying what is displayed in the quiz list.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/QuizInteractionTest.kt
JordanMarshall: Which ones should we preserve? It seems like they all rely on native quiz taking.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/apis/QuizAPI.kt
JordanMarshall: After removing the relevant feature set from the app, I simply removed what was marked as unused by the IDE's static analyzer. You're more familiar with the pact tests so I'll defer to you on what we should do here.

joehoag: OK, I can make a second pass.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/QuizInteractionTest.kt
joehoag: Sorry -- I admit to being a little confused about which logic is "native" and which is still relevant.  But, taking a closer look, it seems like even the testQuiz_xxx tests involve not merely displaying quizzes but taking them as well.

So no action necessary.  I'll create a ticket to myself to look into the possibility of replacement quiz interaction tests.

PR #882, opened 2020-07-13
[Parent][E2E][MBL-14526]: Fix Parent E2E tests

joehoag:
I was able to get rid of the "dart:ui" import in user_color.dart by changing the "color" field from Color to int, and making corresponding changes in other parts of the code.  (I *think* that the color field is actually transmitted as an int.)

It builds, the E2E tests pass, it looks right -- but I could use a review from someone more knowledgeable about how UserColor is transmitted / used.


joehoag:
Whoops -- looks like I need to fix up some tests.  Working on it...

joehoag:
Closing this PR, as Jordan came up with a better, simpler solution.

PR #897, opened 2020-07-20
Update translations

inst-danger:


apps/flutter_parent/lib/l10n/res/intl_fr_CA.arb
```diff
@@ -1895,8 +1979,8 @@
     "type": "text",
     "placeholders": {}
   },
-  "QR scanning requires camera access.": "La numérisation QR nécessite un accès à la caméra.",
-  "@QR scanning requires camera access.": {
+  "QR scanning requires camera access": "La numérisation par balayage du code QR nécessite un accès à la caméra",
+  "@QR scanning requires camera access": {
```
joehoag, line 1983: Really bizarre that, in some of these languages, removal of the period resulted in so much change to the translated string.

PR #911, opened 2020-07-30
[Teacher][MBL-14604] Java -> Kotlin, additional cleanup

JordanMarshall:


apps/teacher/src/main/java/com/instructure/teacher/utils/RoboTesting.kt
```diff
@@ -0,0 +1,18 @@
+package com.instructure.teacher.utils
```
Blackchai, line 1: Do we still need this? Do we still run any of the robo tests?

JordanMarshall: Hmm, I'll just go ahead and delete it given that both the gradle project and bitrise project for parent have since been wiped. Apparently Teacher also has a bitrise robo project, but it hasn't been run since 2018.

PR #917, opened 2020-08-03
[MBL-14444][Student/Teacher] Send context code when adding a new message

armstrongnate:
refs: MBL-14444
affects: student, teacher
release note: none

Test plan:
- Log in as the student in the ticket
- Inbox > Tap the only conversation
- Reply directly to the first message (begins with _I am a teacher_)
- View the conversation as the teacher (in app or on web)
- You should not see the reply that the student made

I don't know how we test this since it doesn't impact the UI for the
current user. Only difference is that the receiver does not actually
receive the message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,12 @@ Student | `./gradle/gradlew -p apps :student:assembleDevDebug`
 Teacher | `./gradle/gradlew -p apps :teacher:assembleDevDebug`
 Parent  | `./gradle/gradlew -p apps :parent:assembleDevDebug`
 
+## Running tests
+
+To run unit tests for Student and Teacher
+1. Open the Build Variants window and set the variant to `qaDebug` for the app that you wish to test.
+2. You can run the tests by tapping on the play button next to the test case or the test class.
+
 ## Applications:
 
 #### The Applications we have published on Google Play.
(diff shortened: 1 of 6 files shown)


libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/apis/InboxApi.kt
```diff
@@ -71,7 +71,7 @@ object InboxApi {
         fun deleteMessages(@Path("conversationId") conversationId: Long, @Query("remove[]") messageIds: List<Long>): Call<Conversation>
 
         @POST("conversations/{conversationId}/add_message?group_conversation=true")
-        fun addMessage(@Path("conversationId") conversationId: Long, @Query("recipients[]") recipientIds: List<String>, @Query(value = "body", encoded = true) body: String, @Query("included_messages[]") includedMessageIds: LongArray, @Query("attachment_ids[]") attachmentIds: LongArray): Call<Conversation>
+        fun addMessage(@Path("conversationId") conversationId: Long, @Query("recipients[]") recipientIds: List<String>, @Query(value = "body", encoded = true) body: String, @Query("included_messages[]") includedMessageIds: LongArray, @Query("attachment_ids[]") attachmentIds: LongArray, @Query("context_code") contextCode: String?): Call<Conversation>
```
armstrongnate, line 74: I assume the reason this work is because the back-end will use the context code to determine if the enrollment still exists or not. We always send the context code on iOS too.

README.md
```diff
@@ -32,6 +32,12 @@ Student | `./gradle/gradlew -p apps :student:assembleDevDebug`
 Teacher | `./gradle/gradlew -p apps :teacher:assembleDevDebug`
 Parent  | `./gradle/gradlew -p apps :parent:assembleDevDebug`
 
+## Running tests
```
Blackchai, line 35: Thanks for adding this haha :D

PR #919, opened 2020-08-03
[MBL-11767][Student] Display rating when only graders can like

armstrongnate:
refs: MBL-11767
affects: student
release note: Adding rating to discussions when only graders can like

Test plan:
- View a discussion that does not allow rating
  - Student & Teacher: No rating content should be displayed
- View a discussion that allows ratings
  - Student & Teacher: You should be able to like and see the like count
- View a discussion that allows ratings and only allows graders to like
  - Student:
    - You should see like counts when the count is greater than 0
    - You should not see the like button
    - You should not be able to edit likes
  - Teacher:
    - You should be able to like the replies and the count should be displayed

### DIFF
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/discussions/DiscussionUtils.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/discussions/DiscussionUtils.kt
@@ -305,6 +305,7 @@ object DiscussionUtils {
                 allowReplies(canvasContext, discussionTopicHeader),
                 allowEditing(canvasContext),
                 allowLiking(canvasContext, discussionTopicHeader),
+                discussionTopicHeader.allowRating,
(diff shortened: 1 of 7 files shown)


libs/pandautils/src/main/java/com/instructure/pandautils/discussions/DiscussionUtils.kt
```diff
@@ -391,16 +392,10 @@ object DiscussionUtils {
...
-                }
-            } else {
-                return true
-            }
-        }
-        return false
+        val isGrader = ((canvasContext as Course).isTeacher || canvasContext.isTA)
+        return header.allowRating && (
+            !header.onlyGradersCanRate || isGrader
+        )
```
armstrongnate, line 398: Small refactor to eliminate nested branches. Logic is the same.

PR #923, opened 2020-08-06
[Student][MBL-14567] Fixed an issue with RCE Image links

Blackchai:
See ticket for test plan. I've also tested various pages with internal / external links. Let me know if you need more info.


libs/pandautils/src/main/java/com/instructure/pandautils/utils/WebViewExtensions.kt
```diff
@@ -146,6 +146,31 @@ fun handleLTIPlaceHolders(placeHolderList: ArrayList<Placeholder>, html: String)
...
 
+/**
+ * Goal is to strip target="_blank" from a tags i.e. <a href="https://google.com" target="_blank">
+ *
+ * This enables us to keep canvasWebView.settings.setSupportMultipleWindows(true) set to true
+ * but still keep RCE image/links working properly in a webview
+ */
+fun handleBlankTarget(html: String): String {
+    if(html.contains("<a href")) {
+        val matcher = Pattern.compile("<a href=\"([^\"]+)\" target=\"_blank\">").matcher(html)
```
joehoag, line 157: As was discussed in standup, we should be able to come up with a new regex that doesn't assume that href/target are the only attributes in the link.

Also, do you consider it worrisome that there might be several links in "html", not just one?  Will this strategy handle that situation correctly?  It might be worth writing some unit tests for this call that test these types of edge cases.

Blackchai: As far as I can tell, the image type links that this is targeting are always going to be a href + target. As such, I think that part of it should stay the same. Those are the only links that need modification and I'd rather not touch any additional links. I'll do some testing though on multiple links + etc and add a couple unit tests.

PR #925, opened 2020-08-07
[Student][MBL-14606] Java -> Kotlin (Part 2)

JordanMarshall:


apps/student/src/main/java/com/instructure/student/activity/InternalWebViewActivity.kt
```diff
@@ -0,0 +1,121 @@
...
+    override fun onUpPressed() = finish()
+
+    override fun onBackPressed() {
+        val fragment = supportFragmentManager.findFragmentById(R.id.container) as? InternalWebviewFragment
+        if (fragment?.handleBackPressed() != true) super.onBackPressed()
+    }
+
+    private fun setActionBarStatusBarColors(actionBarColor: Int, statusBarColor: Int) {
+        val colorDrawable = ColorDrawable(actionBarColor)
+        supportActionBar!!.setBackgroundDrawable(colorDrawable)
```
Blackchai, line 75: safe call here?

apps/student/src/main/java/com/instructure/student/adapter/RecipientAdapter.kt
```diff
@@ -24,7 +24,5 @@ class RecipientAdapter(context: Context) : BaseRecipientAdapter(context) {
     override fun getRecipientManager() = canvasRecipientManager
 
     val canvasRecipientManager: CanvasRecipientManager
-        get() = CanvasRecipientManager.getInstance(context)
-            .setRecipientCallback(this)
-            .setPhotoCallback(this)
+        get() = CanvasRecipientManager.getInstance().setRecipientCallback(this)
```
Blackchai, line 27: Where did the photo callback go?

JordanMarshall: It literally did nothing, so it was removed

Blackchai: lmao, fair nuff.

apps/student/src/main/java/com/instructure/student/activity/InternalWebViewActivity.kt
Blackchai: On the off chance that the toolbar is null, I'd rather just have this do nothing than NPE.

JordanMarshall: 👍

PR #934, opened 2020-08-14
[MBL-14642][All] Hide help buttons on landing pages

armstrongnate:
This is temporary while we figure out a better solution to mobile login
issues.

Support is overwhelmed with these requests right now and are
having to mass-reply to them all so this button is doing more harm than
good right now.

Test plan: Landing pages in all the apps should not have a help button.

### DIFF
--- a/apps/flutter_parent/lib/screens/login_landing_screen.dart
+++ b/apps/flutter_parent/lib/screens/login_landing_screen.dart
@@ -85,6 +85,7 @@ class LoginLandingScreen extends StatelessWidget {
                   child: FullScreenScrollContainer(
                     horizontalPadding: 0,
                     children: <Widget>[
+                      // _helpRequestButton(context), // add back after we make mobile login better
                       Expanded(child: _body(context)),
                       SizedBox(height: 56.0), // Sizedbox to offset helpRequestButton
                     ],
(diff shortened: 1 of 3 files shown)


apps/flutter_parent/lib/screens/login_landing_screen.dart
```diff
@@ -85,7 +85,6 @@ class LoginLandingScreen extends StatelessWidget {
                   child: FullScreenScrollContainer(
                     horizontalPadding: 0,
                     children: <Widget>[
-                      _helpRequestButton(context),
```
joehoag, line 88: Consider commenting out instead of removing?  So it can be easily added back in later?

apps/flutter_parent/test/screens/login/login_landing_screen_test.dart
```diff
@@ -239,17 +239,6 @@ void main() {
...
-  testWidgetsWithAccessibilityChecks('Tapping help button shows help dialog', (tester) async {
-    await tester.pumpWidget(TestApp(LoginLandingScreen()));
-    await tester.pumpAndSettle();
-
-    await tester.tap(find.byIcon(CanvasIcons.question));
-    await tester.pumpAndSettle();
-
-    expect(find.byType(ErrorReportDialog), findsOneWidget);
-    verify(analytics.logEvent(any)).called(1);
-  });
```
joehoag, line 251: Ditto -- consider commenting out instead of removing, perhaps with some note as to why it is being (temporarily) removed.

joehoag:
I've verified that the help button is no longer on any of the apps' landing pages after these changes.  I'd to get feedback on the suggestion to commenting-out certain code sections instead of deleting them before approving.

armstrongnate:
I agree, commenting out instead of removing sounds like a better idea.

PR #942, opened 2020-08-19
Update readme with student/teacher build status

JordanMarshall:
Also updates the parent build command

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ Android Studio > Import Project > canvas-android/apps/build.gradle
 2. Make sure the `main.dart` configuration is selected
 3. Tap 'Run' (`^R`) to run the app
 
-App | Command
---- | ---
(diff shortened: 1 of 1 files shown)


joehoag:
I clicked on the build status links (on the main github page in your branch) and they sent me to a blank page.

JordanMarshall:
It seems GitHub automatically makes markdown images clickable and I don't think that can be disabled. Clicking on the build status loads the build status image in a new page (at least for me). I could make it link to the respective bitrise projects, but I don't believe those are publicly accessible.

joehoag:
Oh, ok, so it gives some kind of status while the build is taking place.  Now I get it.

JordanMarshall:
These badges are for the open source projects that could fail to build if we change secrets without updating the open source data.

PR #947, opened 2020-08-21
[WIP][Parent][MBL-14659] Use observees endpoint to get list of students

JordanMarshall:
This fixes a separate issue causing users to see the "Not A Parent" screen if observee course access is restricted before the course begins. See ticket for repro steps.


JordanMarshall:
Will be fixed on Canvas' end

PR #975, opened 2020-09-08
[Teacher][Student][MBL-14688] Chips update

JordanMarshall:


apps/student/src/androidTest/java/com/instructure/student/ui/pages/NewMessagePage.kt
```diff
@@ -68,7 +69,7 @@ class NewMessagePage : BasePage() {
     }
 
     fun assertRecipientsNotEmpty() {
-        onView(withId(R.id.recipientsView)).check(AssertTextEmpty())
+        chipGroup.check(matches(hasChildCount(0)))
```
joehoag, line 72: Wait -- isn't this asserting that the recipient list IS empty?

JordanMarshall: You're right, perhaps the function name was incorrect to begin with? It's performing the same check as before, the difference being that previously the recipients were populated inline inside the text field but now they're populated in a separate view group.

apps/student/src/main/res/values/themes_canvastheme.xml
```diff
@@ -19,7 +19,7 @@
 <!-- Generated with http://android-holo-colors.com -->
 <resources xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <style name="CanvasMaterialTheme_Default" parent="Theme.AppCompat.Light.DarkActionBar">
+    <style name="CanvasMaterialTheme_Default" parent="Theme.MaterialComponents.Light.DarkActionBar.Bridge">
```
joehoag, line 22: This must affect more than chips, right?

JordanMarshall: The 'Bridge' theme is essentially a compatibility theme that introduces new material styles without changing existing AppCompat styles: https://material.io/develop/android/docs/getting-started#bridge-themes

apps/teacher/src/main/java/com/instructure/teacher/fragments/AddMessageFragment.kt
```diff
@@ -256,7 +223,7 @@ class AddMessageFragment : BasePresenterFragment<AddMessagePresenter, AddMessage
 
         // Get courses and groups if this is a new compose message
         if (isNewMessage) {
-            presenter.getAllCoursesAndGroups(true)
+            presenter.getAllCoursesAndGroups(false)
```
joehoag, line 226: Did we really want to change this from true to false?

automation/canvas_espresso/src/main/kotlin/com/instructure/canvas/espresso/CustomMatchers.kt
```diff
@@ -63,3 +64,12 @@ fun isElementDisplayed(resourceId: Int) : Boolean {
...
 }
 
+inline fun <reified T : View> typedViewCondition(crossinline onCheckCondition: (T) -> Boolean): Matcher<View> {
+    return object : BaseMatcher<View>() {
+        override fun matches(item: Any?): Boolean = (item as? T)?.let(onCheckCondition) ?: false
+        override fun describeTo(description: Description?) {
+            description?.appendText("matches view type '${T::class.java.simpleName}' and fulfills the given condition")
+        }
+    }
+}
```
joehoag, line 74: Nice.

apps/teacher/src/main/java/com/instructure/teacher/fragments/AddMessageFragment.kt
JordanMarshall: This is part of a fix for a bug that would reset the course spinner every time onResume was called (partially due to how our MVP architecture works). The downside is that the course spinner will not automatically reflect new/removed courses for up to an hour if the user has recently used this page.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/BasicUser.kt
```diff
@@ -32,7 +32,7 @@ data class BasicUser constructor(
         fun userToBasicUser(user: User): BasicUser {
             val basicUser = BasicUser()
             basicUser.id = user.id
-            basicUser.name = user.name
+            basicUser.name = user.shortName
```
joehoag, line 35: This would seem to impact more than chips.  Are we assuming that nothing important got broken because our tests are passing?

JordanMarshall: I verified that this is only being used to populate initial recipients; these BasicUsers are being mapped to Recipients in a later step in the compose fragments. I also double-checked the API docs to confirm that the 'name' is indeed the user's short name. This means means that we previously had a mismatch between the names used to populate initial recipients and the names provided by the recipients API, and our tests just happened to be passing.

One thing we _could_ do to alleviate ambiguity is have this return a Recipient instead of a BasicUser, and change the internals of the consuming fragments to use Recipients, but I'd have to spend some time figuring out if that would be feasible and/or break anything.

apps/student/src/androidTest/java/com/instructure/student/ui/pages/NewMessagePage.kt
joehoag: Right -- the old one was not used.

joehoag:
One general question: I recently learned that a user could (if granted permission) change his user name (which I believe is "shortName").  If that were to occur, would it wreak havoc on email thread recipient tracking logic?  Or would it just all magically get worked out because we really traffic in user ids, not user names?

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/BasicUser.kt
joehoag: You could potentially open a low-priority ticket for that.  I think that, as long as nothing has been broken (i.e., nothing was dependent on name == name instead of name == shortName), this change is good as is.

JordanMarshall:
We do use IDs under the hood for everything, so if a user changes their name then any mismatch due to caching should theoretically only be a cosmetic issue.

apps/student/src/main/res/values/themes_canvastheme.xml
joehoag: 👍

apps/student/src/androidTest/java/com/instructure/student/ui/pages/NewMessagePage.kt
joehoag: I recommend removing this.  Other than that, it seems like we're good to go.

JordanMarshall: 👍  I'll get rid of it

joehoag (approved):
I recommend removing the erroneously named, never used assertRecipientsNotEmpty() method that was discussed.  Then feel free to merge.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/BasicUser.kt
JordanMarshall: 👍  Opened [MBL-14725](https://instructure.atlassian.net/browse/MBL-14725)

PR #977, opened 2020-09-09
[Student][MBL-14634] Fix issue with linking to disabled modules

Blackchai:


JordanMarshall (approved):
LGTM, tested with ticket creds as well as with a student and teacher in my own sandbox. The modules list remains accessible for teachers which I believe is the behavior we want. My only comment is that the lock message "Something went wrong" seems a bit generic, but this is such an edge case that it probably doesn't matter much.

PR #983, opened 2020-09-11
[Teacher][RC-1.11.1] Fix crashes in FileSearchFragment and AnnotationCommentListFragment

JordanMarshall:


apps/teacher/src/main/java/com/instructure/teacher/PSPDFKit/AnnotationComments/AnnotationCommentListFragment.kt
```diff
@@ -74,7 +74,7 @@ class AnnotationCommentListFragment : BaseListFragment<
         })
     }
 
-    override val recyclerView: RecyclerView = annotationCommentsRecyclerView
+    override val recyclerView: RecyclerView get() = annotationCommentsRecyclerView
```
joehoag, line 77: Educate me: What difference does this make?

JordanMarshall: So the `recyclerView` property here is part of the MVP architecture we have set up for the Teacher app, and `annotationCommentsRecyclerView` is a view in the fragment's layout. `recyclerView` is only ever accessed after the fragment's layout has been inflated, however the override here was erroneously set up as a property assignment (my bad) which occurs at fragment instantiation, resulting in a NPE when trying to access the `annotationCommentsRecyclerView` view that isn't available yet. Switching from an assignment to a getter fixes that.

joehoag: 👍  Thanks.

joehoag (approved):

PR #990, opened 2020-09-15
[Polls] Remove polls app

JordanMarshall:
Justification:
- It is no longer maintained. The last update was over three years ago (September 7, 2017).
- The last successful CI build was over a year ago.
- There are only ~2k total active installs.
- Only 8.5% increase in year-over-year active installs. Compare to:
  - Student: 59.2%
  - Teacher: 100.9%
  - Parent: 347.8%
- It's our lowest-rated app at 3.5 stars. Even Parent is rated higher at 3.9 stars.
- It has received a total of 9 user ratings over the past year, and only 8 written reviews in the past three and a half years, of which the only 5-star review reads "nice."
- We no longer receive crash reports for it. We'd have to update it to migrate to Firebase.
- The existing app can remain published in the play store for as long as we need.


armstrongnate:
OMG `-15k`. That is amazing.

We can definitely delete the code. Documentation is working on getting a post up about the official deprecation. The plan is to remove it from the stores in January of 2021. 👍 👍

Blackchai (approved):

PR #995, opened 2020-09-16
[Student][RC 6.9] Hotfix for modules

Blackchai:


JordanMarshall (approved):

apps/student/src/main/java/com/instructure/student/adapter/ModuleListRecyclerAdapter.kt
```diff
@@ -365,7 +365,8 @@ open class ModuleListRecyclerAdapter(
             val tabs = awaitApi<List<Tab>> { TabManager.getTabs(courseContext, it, isRefresh) }
                     .filter { !(it.isExternal && it.isHidden) }
 
-            if (tabs.find { it.tabId == "modules" } != null) {
+            // We only want to show modules if its a course nav option OR set to as the homepage
+            if (tabs.find { it.tabId == "modules" } != null || (courseContext as Course).homePage?.apiString == "modules") {
```
joehoag, line 369: Are you sure that courseContext is a course?  If not, then the safe way to write the line above would be

            if (tabs.find { it.tabId == "modules" } != null || (courseContext as? Course)?.homePage?.apiString == "modules")

PR #997, opened 2020-09-17
[Teacher][RC-1.11.1] Fix chips crash on tablets

JordanMarshall:


joehoag (approved):
A good argument for having tablet runs of our teacher tests.

PR #999, opened 2020-09-17
[Teacher][RC-1.11.1] Fix message thread crash

JordanMarshall:
This fixes a crash that can occur when a conversation fails to load from the API after routing to MessageThreadFragment from a push notification. After some investigation is appears that this was an existing crash, but shows up as a new crash due to line changes.


Blackchai (approved):
LGTM

PR #1008, opened 2020-09-21
[All][MBL-14746] Fix for webview keyboards

Blackchai:
Keyboards weren't showing  up in places like essay quizzes, looks like it was the same issue as the RCE webview problem.


joehoag (approved):
The change looks good, and I took the bitrise debug student app for a spin and you can now input an answer to an essay quiz question.

Could there conceivably be any more places where we need to make this change?

JordanMarshall (approved):
> Could there conceivably be any more places where we need to make this change?

I did some searching but couldn't find any other classes extending WebView, and there are no classes that extend CanvasWebView, so 🤞  we're hopefully good here. It's still possible that somewhere we extend from a WebView subclass that we don't own, but I can't think of any cases like that aside from RCE, which we've already fixed.

PR #1012, opened 2020-09-23
[Student][Teacher][MBL-14763] Fix discussions on API 23

JordanMarshall:
This change removes the recently-added HTML sanitizer dependency which was triggering a [D8/AGP bug](https://issuetracker.google.com/issues/157681341) that caused a NoClassDefFoundError on API 23 when viewing discussions. Sanitization has been moved to `Pronouns.html()` and uses HTML encoding rather than element removal. While this approach continues to prevent the XSS issues resolved by #1011, it does mean that unwanted elements (like script tags) will be shown as non-executable text instead of being removed entirely.

### DIFF
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/Pronouns.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/Pronouns.kt
@@ -25,6 +25,7 @@ import android.text.Spanned
 import android.text.SpannedString
 import android.text.style.StyleSpan
 import androidx.annotation.StringRes
+import androidx.core.text.TextUtilsCompat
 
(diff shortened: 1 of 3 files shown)


Blackchai (approved):
LGTM

joehoag:
The code looks good to me as well.  I've kicked off a fresh "multi-device" test run against this branch to verify that the API-23 tests do better than they did last night.

Has anyone actually tested that this new solution prevents the XSS issues that were reported?

JordanMarshall:
I did manually test these changes using the creds from the [original ticket](https://instructure.atlassian.net/browse/MBL-14758).

joehoag:
OK, I tested them as well and it looks like these changes are indeed preventing the XSS issues that were reported.

PR #1015, opened 2020-09-24
[Student][Inbox][MBL-14772]: Fixed problems with new inbox tests on API-27/28/29

joehoag:
I was matching against a content description to pull up a menu, and that content description could get extra junk characters embedded in it on API 27/28/29.

The normal PR logic won't test the fix, since the issue only manifests on API 27/28/29.  But I did manually run an FTL job with this code against the InboxInteractionTest on API 27/28/29 and it passed.  So I *think* that I've fixed the issue(s).

refs: MBL-14772
affects: Student
release note: Fixed new inbox tests for later android versions

### DIFF
--- a/apps/student/src/androidTest/java/com/instructure/student/ui/pages/DiscussionDetailsPage.kt
+++ b/apps/student/src/androidTest/java/com/instructure/student/ui/pages/DiscussionDetailsPage.kt
@@ -130,11 +130,12 @@ class DiscussionDetailsPage : BasePage(R.id.discussionDetailsPage) {
         }
 
         // If we have not yet seen the desired reply, try one more time, allowing up to
-        // 30 seconds for the page to render and the reply to appear.
+        // 3 retries for the page to render and the reply to appear.  (Each attempt
+        // waits ~10 secs before failing.)
         // (It can take a *long* time for the reply to get rendered to the webview on
         // tablets (in FTL, anyway).)
         onWebView(withId(R.id.discussionRepliesWebView))
-                .withElement(findElement(Locator.ID, "message_content_${reply.id}"))
+                .withElementRepeat(findElement(Locator.ID, "message_content_${reply.id}"), 3)
                 .check(webMatches(getText(),containsString(reply.message)))
     }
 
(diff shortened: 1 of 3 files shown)


apps/student/src/androidTest/java/com/instructure/student/ui/pages/DiscussionDetailsPage.kt
```diff
@@ -134,7 +134,7 @@ class DiscussionDetailsPage : BasePage(R.id.discussionDetailsPage) {
         // (It can take a *long* time for the reply to get rendered to the webview on
         // tablets (in FTL, anyway).)
         onWebView(withId(R.id.discussionRepliesWebView))
-                .withElementRepeat(findElement(Locator.ID, "message_content_${reply.id}"), 30)
+                .withElement(findElement(Locator.ID, "message_content_${reply.id}"))
```
joehoag, line 137: Not really related to the issue at hand, but the "repeat" here was counter-productive.  The findElement() call itself waits for 10 secs before failing, so doing that 30 times could result in long-ish failures.

apps/student/src/androidTest/java/com/instructure/student/ui/pages/DiscussionDetailsPage.kt
```diff
@@ -130,11 +130,12 @@ class DiscussionDetailsPage : BasePage(R.id.discussionDetailsPage) {
...
         // If we have not yet seen the desired reply, try one more time, allowing up to
-        // 30 seconds for the page to render and the reply to appear.
+        // 3 retries for the page to render and the reply to appear.  (Each attempt
+        // waits ~10 secs before failing.)
         // (It can take a *long* time for the reply to get rendered to the webview on
         // tablets (in FTL, anyway).)
         onWebView(withId(R.id.discussionRepliesWebView))
-                .withElementRepeat(findElement(Locator.ID, "message_content_${reply.id}"), 30)
+                .withElementRepeat(findElement(Locator.ID, "message_content_${reply.id}"), 3)
                 .check(webMatches(getText(),containsString(reply.message)))
```
joehoag, line 139: In this case, 30 was a bit much, since findElement() waits 10 seconds before failing.

JordanMarshall (approved):

PR #1016, opened 2020-09-24
[Student][Teacher][MBL-14725] Pass Recipient instead of BasicUser when composing messages

JordanMarshall:
Also:
 - Fixes recipient search and selection not working for 'message students who'
 - Fixes crash when trying to use 'message students who' with groups
 - Constrains recipient search input to a single line


joehoag (approved):
It looks good to me, and I'm going to rely on our Inbox tests to verify that the changes didn't break anything.

Small nit: One thing that made the review a little confusing was that we had to convert between List<Recipient> and ArrayList<Recipient> in a few places.  But I'm guessing that the type differences are optimizations for the operations that have to be done in their respective areas.

PR #1027, opened 2020-10-02
[Parent][MBL-14775] Use enrollment state to filter courses

JordanMarshall:
We were previously using course state, date checks, and various flags to filter the dashboard courses client-side. This PR changes the courses API call to specify the enrollment state, thereby eliminating the need for those other filters and aligning the logic with iOS. See the [ticket](https://instructure.atlassian.net/browse/MBL-14775) for repro steps.

refs: MBL-14775
affects: Parent
release note: Fixed an issue where the app would not show ended courses that allow continued user participation


apps/flutter_parent/test/models/course_test.dart
```diff
@@ -83,168 +83,6 @@ void main() {
     });
   });
 
-  group('isValidForParent', () {
```
joehoag, line 86: It's a shame that we had to get rid of so many tests.  Could any of them have been repurposed to test the new dashboard enrollment_state policy?

JordanMarshall: We _could,_ but it would essentially be testing the API. It might be useful for E2E tests though, dunno.

joehoag: OK, I'll think about an E2E test.

joehoag (approved):
The change looks simple enough.  And I was able to verify via the ticket repro that two classes showed up after the changes, whereas only one showed up before the changes.

PR #1028, opened 2020-10-05
[Student] Fix crash in CalendarEventFragment

JordanMarshall:
This fixes a crash that occurs in CalendarEventFragment when the fragment is removed before its initial API call completes. To repro, open a calendar event (from the calendar, to do list, or syllabus) and then quickly press the back button.


Blackchai (approved):
A classic crash fix  haha

PR #1029, opened 2020-10-05
[Student] Fix crash during submission file upload

JordanMarshall:
I was unable to repro the crash, so this is a speculative fix. As far as I can tell the crash occurs when `NumberHelper.readableFileSize()` is passed what appears to be a very incorrect file size (e.g. 1 exabyte or more), causing an ArrayIndexOutOfBoundsException when attempting to look up the unit name from an array that only handles units up to the terabyte. I'm not sure where the bad files size is coming from, but it appears to disproportionately affect OPPO devices.

This change fixes the crash by capping the max unit to terabyte, and hardens the function with additional range checks. It also changes how file sizes are rounded, fixing an issue where file sizes slightly below the threshold of the next-highest unit were displayed in a potentially confusing way. For example, 1023.96 KB would display as "1024 KB" which is technically 1 MB, but with this change it will display as "1023.9 KB" instead.


Blackchai (approved):
LGTM

PR #1032, opened 2020-10-07
[Ops][MBL-14770] Update Android Gradle Plugin

JordanMarshall:
refs: MBL-14770
affects: Student, Teacher
release note: none

### DIFF
--- a/libs/annotations/src/main/java/com/instructure/annotations/FileCaching/FetchFileAsyncTask.kt
+++ b/libs/annotations/src/main/java/com/instructure/annotations/FileCaching/FetchFileAsyncTask.kt
@@ -27,6 +27,8 @@ import com.instructure.canvasapi2.utils.ContextKeeper
 import com.instructure.canvasapi2.utils.validOrNull
 import okhttp3.Request
 import okio.Okio
+import okio.buffer
+import okio.sink
 import java.io.File
 import java.io.FileInputStream
 import java.io.FileNotFoundException
@@ -98,7 +100,7 @@ class FetchFileAsyncTask private constructor(
             val total = response.body()?.contentLength()?.toFloat() ?: 0f
 
             // Set up source and sink
-            val sink = Okio.buffer(Okio.sink(tmpFile))
+            val sink = tmpFile.sink().buffer()
             val source = response.body()?.source()
 
             var downloaded = 0L
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/weave/AwaitQL.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/weave/AwaitQL.kt
@@ -44,9 +44,9 @@ suspend fun <DATA> awaitQL(managerCall: ApolloManagerCall<DATA>): DATA {
         val callback = object : QLCallback<DATA>() {
             override fun onResponse(response: Response<DATA>, type: ApiType) {
                 when {
-                    response.hasErrors() -> onFail(ApolloException(response.errors().first().message()))
-                    response.data() == null -> onFail(ApolloException("Response data is null!"))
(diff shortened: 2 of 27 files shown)


joehoag:
I've kicked off a ui-test-coverage run against this branch to make sure that our coverage/jacoco logic is still happy.

https://app.bitrise.io/build/42c0130a3f910a57

joehoag:
ui-test-coverage run looks good!

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/weave/AwaitQL.kt
```diff
@@ -44,9 +44,9 @@ suspend fun <DATA> awaitQL(managerCall: ApolloManagerCall<DATA>): DATA {
         val callback = object : QLCallback<DATA>() {
             override fun onResponse(response: Response<DATA>, type: ApiType) {
                 when {
-                    response.hasErrors() -> onFail(ApolloException(response.errors().first().message()))
-                    response.data() == null -> onFail(ApolloException("Response data is null!"))
-                    else -> continuation.resumeSafely(response.data()!!)
+                    response.hasErrors() -> onFail(ApolloException(response.errors?.firstOrNull()?.message ?: "Unknown error"))
```
joehoag, line 47: Nit: If hasErrors() is true, then would "!!" be a more appropriate operator than "?" ?

So: response.errors!!.first().message might suffice.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/weave/AwaitQLPaginated.kt
```diff
@@ -84,7 +84,7 @@ class WeaveQLPager<DATA>(
             if (isCanceled) return@responseBlock
             weaveCoroutine.onUI {
                 if (response.hasErrors()) {
-                    val message = response.errors().first().message()
+                    val message = response.errors?.firstOrNull()?.message ?: "Unknown error"
```
joehoag, line 87: Same thing here.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/utils/weave/AwaitQL.kt
JordanMarshall: True, if `hasErrors()` is true then we should be safe to use `!!`. I'll update these.

libs/annotations/src/main/java/com/instructure/annotations/FileCaching/FetchFileAsyncTask.kt
```diff
@@ -98,7 +100,7 @@ class FetchFileAsyncTask private constructor(
             val total = response.body()?.contentLength()?.toFloat() ?: 0f
 
             // Set up source and sink
-            val sink = Okio.buffer(Okio.sink(tmpFile))
+            val sink = tmpFile.sink().buffer()
```
joehoag, line 103: I didn't see that we changed the okio version anywhere.  Are we just making this change (in a few places) because the new code is a better way to do it?

JordanMarshall: We didn't explicitly update okio, but I wouldn't be surprised if the newer apollo version pulled it in. Regardless, that code was marked as an error and was failing the build so I had to change it.

joehoag: 👍

joehoag:
Just because I'm paranoid, I'm running a multi-device test suite against this branch.  I would be pretty surprised if it surfaced any issues.

https://app.bitrise.io/build/713beae118f5dedc

joehoag (approved):
Looks good!

PR #1033, opened 2020-10-08
[Student][MBL-14795] Remember collapsed modules

hermannakos:


apps/student/src/main/java/com/instructure/student/util/CollapsedModulesStore.kt
```diff
@@ -0,0 +1,30 @@
+package com.instructure.student.util
```
joehoag, line 1: You should include our copyright notice at the top of new code modules.  You can just copy it from another file.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ModuleInteractionTest.kt
```diff
@@ -325,6 +325,7 @@ class ModuleInteractionTest : StudentTest() {
 
     // Module can't be accessed until the availability date has passed
     @Test
+    @Stub
```
joehoag, line 328: Did you create a Jira ticket for me to get this test working again, as we discussed?

hermannakos: Not yet, wanted to do that once the changes were accepted.

joehoag (changes requested):
So, from some testing, the new behavior seems to be:
(1) Modules default to open
(2) Modules' open/close state will be saved across app restarts

Is that the intended behavior?

I was able to verify that the ModulesE2ETest still passes.

My one complaint (applies to the test that is failing) is that empty modules should not show the expand/collapse icons.  Because you can't expand or collapse them.  I guess you could ask UX or Product (both Nate for now) what the proper behavior should be in that case.

joehoag:
OK, I locally altered the failing test to include an assignment inside the locked module, and in that case the expand/collapse support made sense and did the right thing.  We should just remove expand/contract support for empty modules.

joehoag (approved):
Sounds like this is acceptable as is -- we'll address empty-module logic in a separate ticket.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ModuleInteractionTest.kt
hermannakos: I created the ticket, MBL-14842

PR #1034, opened 2020-10-08
[Student][Teacher] Retire remote configs set to true

Blackchai:
This PR removes the existing remote configs currently set to true in student/teacher. To test, simply check that the features are on and working. This includes qr_login and pairing (from student).


joehoag (approved):
The code looks good, and I was able to verify that QR-login still worked on all three apps, and "pair with observer" worked as well for creating a parent account.

PR #1036, opened 2020-10-09
[Teacher] Update test orchestrator

JordanMarshall:


joehoag (approved):
Looks good, as long as the tests pass.  (I think that they run on API-25.)

PR #1037, opened 2020-10-09
[Student][MBL-14806] Group remains on dashboard when un-favorited

tamaskozmer:
The reproduction steps and the expected behavior can be seen in the ticket.

refs: MBL-14775
affects: Student
release note: Fixed an issue where a group remains on the dashboard when unfavorited.


Blackchai (approved):
Looks good to me!

PR #1042, opened 2020-10-12
[Teacher][MBL-11775] Changing a PDF file name doesn't show the name change in the preview fragment after saving

tamaskozmer:
The reproduction steps and the expected behavior can be seen in the ticket. To test this you need a user, who has a PDF in their files.

refs: MBL-11775
affects: Teacher
release note: Fixed an issue where a the edited file's name doesn't update.


Blackchai (approved):
Looks good to me!

PR #1043, opened 2020-10-12
MBL-10743 - Add counter to Speedgrader Files tab

hermannakos:


JordanMarshall (changes requested):
This looks good so far, however we'll also want the file count to update when the user selects a different submission from the student's submission history.

JordanMarshall (approved):

PR #1044, opened 2020-10-12
[Student][Teacher] Fix crash in UploadFilesDialog

JordanMarshall:
Repro: In the file/folder list, choose the file upload option and select two or more files to be uploaded. Then attempt to remove the files by tapping the items' 'X' buttons in rapid sequence, which will result in a crash.

refs: none
affects: Student, Teacher
release note: Fixed a crash when rapidly removing files from the file upload dialog


Blackchai (approved):
LGTM!

PR #1045, opened 2020-10-12
[Student] Fix crash in ShareFileDestinationDialog

JordanMarshall:
Repro: In the files list select and view a PDF, and then in the menu overflow choose "Upload To Canvas". This will show a red screen and will begin animating a dialog into position (if it requests permissions, grant them and then back out and start over). If you press the back button after the dialog begins animating but before it finishes, the app will crash.

refs: none
affects: Student
release note: Fixed a crash when closing the share dialog shortly after opening it


Blackchai (approved):
LGTM!

PR #1048, opened 2020-10-13
[Student][Teacher][RemoteConfig][MBL-14822]: Fix problem on remote config settings page

joehoag:
Fixed a problem where old TextWatchers were still registered with a ViewHolder when it got recycled on the Remote Config settings page, resulting in multiple responses to field changes.  This was causing the RemoteConfig E2E test to fail, but it was actually an app issue.

refs: MBL-14822
affects: Teacher, Student
release note: Fixed internally used Remote Config settings page


Blackchai (approved):
LGTM!

PR #1049, opened 2020-10-13
[Teacher][MBL-14786] Infinite spinner in To Do list after selecting and item and returning

tamaskozmer:
The repro steps are in the ticket.

refs: MBL-14786
affects: Teacher
release note: Fixed an issue where an infinite loading indicator is shown when returning from a To Do item.


Blackchai (approved):
Looks good to me!

PR #1051, opened 2020-10-13
[Student] Fix crash on logout

JordanMarshall:
During logout, the app attempts to retrieve the FCM registration ID so that the canvas push communication channel can be cleaned up. However, on some devices (particularly devices without GMS, e.g. some Huawei, Xioami, and Amazon devices) this process can fail and crash the app. This PR catches that failure and allows the logout process to skip that step and proceed normally.

refs: none
affects: Student
release note: Fixed an uncommon crash when logging out


Blackchai (approved):
LGTM

PR #1052, opened 2020-10-14
[Parent][MBL-14783] Resize the refresh button in the course details AppBar

tamaskozmer:
I also changed the icon on the `CourseRoutingShellScreen` to maintain consistency, however I couldn't find where this screen can be seen. From the router it seems to me that it is unused currently, but maybe I am missing something.

refs: MBL-14783
affects: Parent
release note: Changed refresh button size


Blackchai (approved):
LGTM

PR #1066, opened 2020-10-19
[Teacher][MBL-14841] Empty state added for modules with no items

hermannakos:
refs: MBL-14841
affects: Teacher
release note: Empty state added for modules with no items.

test plan:

### DIFF
--- a/apps/teacher/src/main/java/com/instructure/teacher/features/modules/list/ui/binders/ModuleListEmptyItemBinder.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/features/modules/list/ui/binders/ModuleListEmptyItemBinder.kt
@@ -27,4 +27,6 @@ class ModuleListEmptyItemBinder : ListItemBinder<ModuleListItemData.EmptyItem, M
         get() = R.layout.adapter_module_empty_item
 
     override val bindBehavior = NoBind()
+
+    override fun getItemId(item: ModuleListItemData.EmptyItem): Long = item.moduleId
 }
--- a/apps/teacher/src/main/java/com/instructure/teacher/features/modules/list/ui/binders/ModuleListEmptyItemBinder.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/features/modules/list/ui/binders/ModuleListEmptyItemBinder.kt
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2019 - present Instructure, Inc.
+ *
+ *     This program is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU General Public License as published by
+ *     the Free Software Foundation, version 3 of the License.
+ *
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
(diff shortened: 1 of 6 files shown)


JordanMarshall (changes requested):
Just one issue with multiple empty modules, but otherwise it looks good.

apps/teacher/src/main/res/layout/adapter_module_empty_item.xml
```diff
@@ -0,0 +1,36 @@
...
+  ~
+  ~     This program is distributed in the hope that it will be useful,
+  ~     but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~     GNU General Public License for more details.
+  ~
+  ~     You should have received a copy of the GNU General Public License
+  ~     along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
tamaskozmer, line 16: Do we need a `ConstraintLayout` for this? I'm sure it doesn't have too much performance overhead, if it has any, but I would use a `FrameLayout` in cases like this. It's not an action item, just adding my 2 cents.

apps/teacher/src/main/res/layout/adapter_module_empty_item.xml
```diff
@@ -0,0 +1,36 @@
...
+    android:paddingStart="16dp"
+    android:paddingTop="8dp"
+    android:paddingEnd="16dp"
+    android:paddingBottom="8dp">
+
+    <TextView
+        android:id="@+id/emptyModuleItemTitle"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="@string/noItemsToDisplayShort"
```
tamaskozmer, line 30: I see that you are reusing this string resource here, and just checked that it is already reused in many places. It is a common practice on the project @JordanMarshall ?

tamaskozmer:
I have added two comments, but neither of them requires action, the PR looks good.

apps/teacher/src/main/res/layout/adapter_module_empty_item.xml
hermannakos: We already use this for the same thing in Student, that's why I use it here.

apps/teacher/src/main/java/com/instructure/teacher/features/modules/list/ui/binders/ModuleListEmptyItemBinder.kt
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+package com.instructure.teacher.features.modules.list.ui.binders
+
+import com.instructure.teacher.R
+import com.instructure.teacher.adapters.ListItemBinder
+import com.instructure.teacher.features.modules.list.ui.ModuleListCallback
+import com.instructure.teacher.features.modules.list.ui.ModuleListItemData
+
+class ModuleListEmptyItemBinder : ListItemBinder<ModuleListItemData.EmptyItem, ModuleListCallback>() {
```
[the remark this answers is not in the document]
hermannakos: Thanks, it's fixed now.

apps/teacher/src/main/res/layout/adapter_module_empty_item.xml
JordanMarshall: 👍  It's not uncommon for us to reuse generic empty/error messages, especially if there isn't specific product/design direction.

JordanMarshall (approved):

PR #1067, opened 2020-10-19
[Student][Teacher] Add comment indicators to annotations

Blackchai:
Added indicators to annotations with comments associated.

refs: MBL-14681
affects: Teacher, Student
release note: Added a visual indicator for annotations with comment replies

Test Plan:
-Install Teacher/Student
-Add comments to annotations present on submissions in both apps
-Confirm that the blue discussion icon indicators appear next to those annotations (they should also go away when the comments are deleted)


joehoag:
OK, I'm seeing the blue comment markers in the app when I comment on annotations.  However, when I delete a comment (via canvas-web), I still see the blue comment marker in the app; it's just that when I press it, I get an "Add Comment" dialog instead of a dialog showing me the comment.

Is this the desired behavior?

joehoag (approved):
OK, this would seem to be a series of resource tweaks accompanying some hacks to get comment indicators working.  I'm not going to pretend that I understood the hacks.  But I did test the code out and it appeared to display the comment indicators when expected and remove them when the comment was removed.

tamaskozmer:
I also tried to test this, and experienced some weird behavior:

Sometimes when adding a new Note or highlight with a comment the comment indicator didn't appear, however after I added a new text note the comment indicator appeared on the previously added notes. Maybe it just refreshes the state of the view later?
Also for some Notes the indicator is not there when I reload the PDF.

I'm still getting familiar with this part of the application, actually just explored this part today for the first time, so maybe I am missing something. Maybe Note annotations don't even need to have this indicator? We can talk about this on Slack/Zoom.

Blackchai:
> I also tried to test this, and experienced some weird behavior:
> 
> Sometimes when adding a new Note or highlight with a comment the comment indicator didn't appear, however after I added a new text note the comment indicator appeared on the previously added notes. Maybe it just refreshes the state of the view later?
> Also for some Notes the indicator is not there when I reload the PDF.m.

This should be fixed. Notes (the little  "pins") should never have comment indicators. I had missed one spot where I was suppose to check for them as well as update the notehinter provider. So both issues should be fixed now.

PR #1068, opened 2020-10-20
[all] Clean up some no longer needed TODOs

Blackchai:
Just removed some old todos that we don't need anymore

### DIFF
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Tab.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Tab.kt
@@ -24,19 +24,19 @@ import kotlinx.android.parcel.Parcelize
 @Parcelize
 data class Tab(
         @SerializedName("id")
-        val tabId: String = "", // TODO: null?
+        val tabId: String = "",
         val label: String? = null,
         val type: String = TYPE_INTERNAL,
         @SerializedName("html_url")
         val htmlUrl: String? = null, // Internal url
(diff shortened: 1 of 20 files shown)


libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Tab.kt
```diff
@@ -24,19 +24,19 @@ import kotlinx.android.parcel.Parcelize
...
-        val tabId: String = "", // TODO: null?
+        val tabId: String = "",
         val label: String? = null,
         val type: String = TYPE_INTERNAL,
         @SerializedName("html_url")
         val htmlUrl: String? = null, // Internal url
         @SerializedName("full_url")
         val externalUrl: String? = null, // External url
-        val visibility: String = "none", // possible values are: public, members, admins, and none // TODO: enum?
+        val visibility: String = "none", // possible values are: public, members, admins, and none
```
joehoag, line 34: I'm curious: We do use enums in similar situations elsewhere.  How do we decide whether or not to use an enum in this kind of scenario?

joehoag (approved):
LGTM.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/models/Tab.kt
Blackchai: I think this is the sorta thing we'd convert to an enum when we rewrote the underlying code to mobius or something. Imo.

PR #1078, opened 2020-10-26
[Student][Teacher] Update file size calculation to match web

JordanMarshall:
After noticing that the file sizes displayed for the same files were different between the apps and web, I realized that the app is using binary calculations while web is using metric (e.g. 1KB = 1024B vs 1KB = 1000B). This PR updates the app logic to match [web](https://github.com/instructure/canvas-lms/blob/7b1d44f99899937e5269fcdecdcd1ce4bc4166b2/app/coffeescripts/util/friendlyBytes.js).

refs: none
affects: Student, Teacher
release note: File size calculation now matches web

test plan: Compare displayed file sizes against web. They should be identical in every case except for bytes, where the app will show 'B' and the web will show 'bytes'


Blackchai (approved):
LGTM

PR #1082, opened 2020-11-04
[Teacher][Tests][MBL-14757] | Enable teacher start nightly runs

davidvarga93:
Added .yml files to create new bitrise workflows for UI Tests.

Tablet mode:
https://app.bitrise.io/build/7722b0f9b838e19a#?tab=log

### DIFF
--- a/apps/teacher/flank_landscape.yml
+++ b/apps/teacher/flank_landscape.yml
@@ -1,0 +1,25 @@
+gcloud:
+  project: delta-essence-114723
+  # Use the next two lines to run locally
+  #  app: ./build/outputs/apk/qa/debug/teacher-qa-debug.apk
+  #  test: ./build/outputs/apk/androidTest/qa/debug/teacher-qa-debug-androidTest.apk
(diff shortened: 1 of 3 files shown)


apps/teacher/flank_landscape.yml
```diff
@@ -0,0 +1,25 @@
...
+  test-targets:
+    - notAnnotation com.instructure.canvas.espresso.E2E, com.instructure.canvas.espresso.Stub
+  device:
+    - model: NexusLowRes
+      version: 25
+      locale: en_US
+      orientation: landscape
+
+flank:
+  testShards: 10
```
joehoag, line 23: We used a testShards value of 10 for the Student tests.  But the "primary" workflow for the teacher UI tests uses testShards=16.  I'm just pointing out the difference -- if you like 10, we can stay at 10.

joehoag (approved):
This looks good to me.  The new workflows are suffering failures, but they are real failures that we need to address as opposed to mistakes that we need to fix in our scripts / workflows.

So let's get these changes merged in, and then we can start focussing on all of the failures (a11y and otherwise) occurring in the runs.

PR #1084, opened 2020-11-05
[PDF][MBL-14764] Part 1 of PSPDFKit api updates, toolbar items

Blackchai:
This PR updates pspdfkit to a new minor version and refactors the annotation toolbar grouping to use the newest api/method, which is faaar nicer.

refs: MBL-14765
affects: Teacher, Student
release note:

test plan: Make sure the annotation tools look roughly the same as they did before. The new method results in the undo/redo buttons being grouped, as well as the eraser at times, this is expected behavior.


libs/annotations/src/main/java/com/instructure/annotations/AnnotationCreationGroupingRule.kt
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    override fun areGeneratedGroupItemsSelectable(): Boolean = true
+
+    private val eightItemGrouping = listOf(
+        MenuItem(com.pspdfkit.R.id.pspdf__annotation_creation_toolbar_item_stamp),
+        MenuItem(com.pspdfkit.R.id.pspdf__annotation_creation_toolbar_item_freetext),
+        MenuItem(com.pspdfkit.R.id.pspdf__annotation_creation_toolbar_item_highlight),
+        MenuItem(com.pspdfkit.R.id.pspdf__annotation_creation_toolbar_item_strikeout),
+        MenuItem(com.pspdfkit.R.id.pspdf__annotation_creation_toolbar_group_markup,
```
tamaskozmer, line 41: I think this group should be `com.pspdfkit.R.id.pspdf__annotation_creation_toolbar_group_drawing`. In the other groupings these items are added into that group.

tamaskozmer (changes requested):

libs/annotations/src/main/java/com/instructure/annotations/AnnotationCreationGroupingRule.kt
Blackchai: Oh good catch, thanks!

tamaskozmer (approved):

PR #1086, opened 2020-11-06
[Teacher][Tests][MBL-14939] | Address API-23 test failures for the teacher app

joehoag:
The multi_device test run now passes with these changes.
https://app.bitrise.io/build/40030a15d2f265bc

### DIFF
--- a/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/InboxPageTest.kt
+++ b/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/InboxPageTest.kt
@@ -52,7 +52,7 @@ class InboxPageTest: TeacherTest() {
 
         // Test expects single conversation; filter down to starred conversation
         val unwanted = data.conversations.filter() {entry -> !entry.value.isStarred}
-        unwanted.forEach() {id, conversation -> data.conversations.remove(id)}
+        unwanted.forEach() { (id, conversation) -> data.conversations.remove(id)}
 
         val token = data.tokenFor(teacher)!!
         tokenLogin(data.domain, token, teacher)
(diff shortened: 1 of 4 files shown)


apps/teacher/src/androidTest/java/com/instructure/teacher/ui/InboxPageTest.kt
```diff
@@ -52,7 +52,7 @@ class InboxPageTest: TeacherTest() {
 
         // Test expects single conversation; filter down to starred conversation
         val unwanted = data.conversations.filter() {entry -> !entry.value.isStarred}
-        unwanted.forEach() {id, conversation -> data.conversations.remove(id)}
+        unwanted.forEach() { (id, conversation) -> data.conversations.remove(id)}
```
joehoag, line 55: Without this change, we were getting "java.lang.NoClassDefFoundError: com.instructure.teacher.ui.InboxPageTest$displaysConversation$1" on API-23.

Blackchai: Weird

apps/teacher/src/main/java/com/instructure/teacher/fragments/EditAssignmentDetailsFragment.kt
```diff
@@ -386,6 +387,9 @@ class EditAssignmentDetailsFragment : BaseFragment() {
         }
         // when the RCE editor has focus we want the label to be darker so it matches the title's functionality
         descriptionEditor.setLabel(assignmentDescLabel, R.color.defaultTextDark, R.color.defaultTextGray)
+
+        // Take down progress bar
+        descriptionProgressBar.setGone()
```
joehoag, line 392: For both this scenario and the one below, the progress bar was being set "visible" but never set back to "gone".  Even though the progress bar was not even visible, this behavior still caused API-23 tests to hang based on the progress bar never going idle.  So I added a line to take down the progress bar.

Blackchai (approved):
LGTM

PR #1088, opened 2020-11-10
[Student][MBL-14946] Calendar event won't show time if start time and end time is the same

tamaskozmer:
Testing: Repro steps are in the ticket.

refs: MBL-14946
affects: Student
release note: Fixed an issue when the time is not shown on calendar events in some cases.


hermannakos (approved):
lgtm!

PR #1090, opened 2020-11-10
[PDF][MBL-14944] Fix bug with offpage annotations

Blackchai:
This PR fixes a bug with off page annotations causing the annotation toolbar to disappear _forever_... It also adds a simple clean up to the annotation/edit lists.

refs: MBL-14944
affects: Teacher, Student
release note: Fixed a bug with the annotation tools disappearing for off page clicks.

test plan: On a landscape pdf, attempt to annotate on the blank space arround the page of the document, not on the document/page itself. The annotation toolbar will flicker, but won't disappear! This is the best I could get, and that was with help from PSPDFKits support staff haha.

### DIFF
--- a/libs/annotations/src/main/java/com/instructure/annotations/PdfSubmissionView.kt
+++ b/libs/annotations/src/main/java/com/instructure/annotations/PdfSubmissionView.kt
@@ -153,8 +153,11 @@ abstract class PdfSubmissionView(context: Context) : FrameLayout(context), Annot
         annotationsJob?.cancel()
         pdfContentJob?.cancel()
         fileJob?.cancel()
+        exitHandler?.removeCallbacks(enterCreationMode)
     }
 
+    //private val exitHandler: Handler = Handler(Looper.getMainLooper())
+    //    private val enterCreationMode = Runnable { pdfFragment?.enterAnnotationCreationMode() }
     protected fun unregisterPdfFragmentListeners() {
         pdfFragment?.removeOnAnnotationCreationModeChangeListener(this)
         pdfFragment?.removeOnAnnotationEditingModeChangeListener(this)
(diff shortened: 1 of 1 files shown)


libs/annotations/src/main/java/com/instructure/annotations/PdfSubmissionView.kt
```diff
@@ -652,6 +652,8 @@ abstract class PdfSubmissionView(context: Context) : FrameLayout(context), Annot
     }
 
     override fun onExitAnnotationCreationMode(p0: AnnotationCreationController) {
+        exitHandler.post(enterCreationMode)
```
tamaskozmer, line 655: To be honest this seems a little hacky to me, the PSPDFKit has a method `onExitAnnotationCreationMode` and the first thing we do here is to re-enter the annotation creation mode, which we just exited. Could this cause any additional problems? 

From this method name I assume that this is the expected behavior of PSPDFKit, so if the user clicks outside a PDF we exit the annotation creation mode. If you click inside the PDF again, the annotation creation mode is entered again and the toolbar will be visible again. If we want a behavior when the annotation toolbar is always visible, and we are always in creation mode then this might be our only option. Can't we disable that clicking outside the PDF will trigger this event?

tamaskozmer (changes requested):

libs/annotations/src/main/java/com/instructure/annotations/PdfSubmissionView.kt
Blackchai: We chatted about this on slack and came to the conclusion that it's worth fixing. The toolbar goes away if you attempt to annotate "off document" and will not come back unless you are able to click on an existing annotation.

tamaskozmer (approved):

PR #1093, opened 2020-11-11
[MBL-14885][Teacher] Native view for scheduled items from syllabus summary

tamaskozmer:
Testing: Open a scheduled calendar event from the syllabus page of a course and verify if all the information on the page looks good. You can compare it to the student app. Example user can be found in the JIRA ticket.

refs: MBL-14885
affects: Teacher
release note: Added option to open a calendar event from the syllabus.


apps/teacher/src/main/java/com/instructure/teacher/features/calendar/event/CalendarEventFragment.kt
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    private fun setupViews(viewState: CalendarEventViewState) {
+        dateTitle.text = viewState.dateTitle
+        dateSubtitle.text = viewState.dateSubtitle
+        locationTitle.text = viewState.locationTitle
+        locationSubtitle.text = viewState.locationSubtitle
+
+        calendarEventWebView?.setGone()
```
hermannakos, line 118: It is already set from xml

apps/teacher/src/main/java/com/instructure/teacher/features/calendar/event/CalendarEventStateTransformer.kt
```diff
@@ -0,0 +1,91 @@
...
+        return CalendarEventViewState(
+            scheduleItem.title ?: "",
+            dateTitle,
+            dateSubtitle,
+            locationTitle,
+            locationSubtitle,
+            scheduleItem.description ?: "")
+    }
+
+    private fun getFullDateString(context: Context, date: Date?): String {
```
hermannakos, line 83: I think this could be moved to the DateHelper.

apps/teacher/src/main/res/layout/fragment_calendar_event.xml
```diff
@@ -0,0 +1,115 @@
...
+        tools:text="Deak Ferenc utca 10" />
+
+    <com.instructure.pandautils.views.CanvasWebView
+        android:id="@+id/calendarEventWebView"
+        android:layout_width="wrap_content"
+        android:layout_height="0dp"
+        android:layout_marginTop="20dp"
+        android:padding="16dp"
+        android:visibility="gone"
+        app:layout_constraintStart_toStartOf="parent"
```
hermannakos, line 111: I'm not sure, but shouldn't we add the end constraint.

hermannakos (approved):
Just a few minor things, otherwise looks good.

apps/teacher/src/main/java/com/instructure/teacher/features/calendar/event/CalendarEventViewState.kt
```diff
@@ -0,0 +1,26 @@
...
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package com.instructure.teacher.features.calendar.event
+
+data class CalendarEventViewState(
```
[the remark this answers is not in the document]
tamaskozmer: It can, but I don't think it has any added benefits. It's easier to work with non nullable properties.

apps/teacher/src/main/java/com/instructure/teacher/features/calendar/event/CalendarEventFragment.kt
tamaskozmer: Right, I will check this. Maybe we don't even need to set the visibility at all and it can be always visible, it is just something I copied from the student app.

apps/teacher/src/main/res/layout/fragment_calendar_event.xml
tamaskozmer: Actually I think it won't be a problem for the WebView, but I just realised from this comment that maybe I should add it for the text fields and add some padding as well, because if the text is too long it would look bad at the end of the screen.

apps/teacher/src/main/java/com/instructure/teacher/features/calendar/event/CalendarEventStateTransformer.kt
tamaskozmer: I will check how specific is this for this screen and consider making this global.

PR #1094, opened 2020-11-11
[Student][MBL-14870] Course participation setting prevents students from seeing invitations on the dashboard

tamaskozmer:
Testing: Repro steps are well defined in the ticket. I would suggest creating test data in your own sandbox, instead of using the one specified in the ticket, because dates are important here, and the ticket was created 3 years ago, also you can edit the dates of the course to check edge cases.

refs: MBL-14870
affects: Student
release note: Fixed a bug where in some cases students can't see course invites


hermannakos (approved):
lgtm!

PR #1101, opened 2020-11-16
[MBL-11022][Teacher] Grading slider added for assignments

hermannakos:
refs: MBL-11022
affects: Teacher
release note: Slider added for SpeedGrader for point assignments

test plan:
Open a point assignment.
Slider should be visible, use it to grade the assignment.
If you hold at the minimum value for more than a second, you can clear the grade.
If you hold at the maximum value for more than a second, you can excuse the student


apps/teacher/src/main/java/com/instructure/teacher/view/TooltipView.kt
```diff
@@ -0,0 +1,284 @@
...
+                playSequentially(animIn, animOut)
+                start()
+            }
+        } else {
+            animIn.start()
+        }
+    }
+
+    private fun buildLayout(text: String, maxWidth: Int): Layout {
+        return if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
```
tamaskozmer, line 233: This branch can be eliminated, since our minSdk is 23 (Marshmallow), so this will never get executed.

apps/teacher/src/main/java/com/instructure/teacher/view/TooltipView.kt
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /** Returns the length of the longest line in this layout */
+    private fun Layout.maxLineWidth() = (0 until lineCount).map { getLineWidth(it) }.max()?.toInt()
+            ?: 0
+
+    /** Returns the length of the visible text */
+    private val Layout.visibleTextLength get() = (0 until lineCount).sumBy { getLineVisibleEnd(it) }
+
+    companion object {
```
tamaskozmer, line 277: I don't think we need a companion object, because these constants are never used outside of this file. You can move them to top level constants.

apps/teacher/src/main/java/com/instructure/teacher/view/grade_slider/SpeedGraderSlider.kt
```diff
@@ -0,0 +1,134 @@
...
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package com.instructure.teacher.view.grade_slider
```
tamaskozmer, line 17: As far as I know package names in Kotlin/Java should be camel case.

apps/teacher/src/main/java/com/instructure/teacher/view/grade_slider/SpeedGraderSlider.kt
```diff
@@ -0,0 +1,134 @@
...
+import org.greenrobot.eventbus.EventBus
+import kotlin.properties.Delegates
+
+class SpeedGraderSlider @JvmOverloads constructor(
+        context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
+) : FrameLayout(context, attrs, defStyleAttr) {
+
+    var onGradeChanged: (String, Boolean) -> Unit by Delegates.notNull()
+
+    private lateinit var mAssignment: Assignment
```
tamaskozmer, line 43: I would move away from Hungarian notation in new code we write (I know there is a lot of old code with it), I don't think it has any added benefit, but we can discuss this later.

apps/teacher/src/main/java/com/instructure/teacher/view/grade_slider/SpeedGraderSlider.kt
```diff
@@ -0,0 +1,134 @@
...
+    init {
+        View.inflate(context, R.layout.view_speed_grader_slider, this)
+    }
+
+    fun setData(assignment: Assignment, submission: Submission?, assignee: Assignee) {
+        mAssignment = assignment
+        mSubmission = submission
+        mAssignee = assignee
+
+        if (mAssignment.rubric != null && mAssignment.rubric!!.isNotEmpty() && mAssignment.gradingType?.let { Assignment.getGradingTypeFromAPIString(it) } == Assignment.GradingType.POINTS) {
```
tamaskozmer, line 70: I don't think this logic belongs here. Do we have a Presenter or something for the Fragment where this kind of logic can be placed? Ideally it would be better to make a decision there and depending on that show the correct view.

tamaskozmer (changes requested):

apps/teacher/src/main/java/com/instructure/teacher/view/TooltipView.kt
```diff
@@ -7,33 +22,40 @@ import android.annotation.SuppressLint
...
-    private var textLayout: Layout? = null
+    protected var textLayout: Layout? = null
 
     /** Rect of the anchor view */
-    private var anchorRect = Rect()
+    protected var anchorRect = Rect()
 
     /** The animation used to fade the tooltip in an out */
-    private var animator: Animator? = null
+    protected var animator: Animator? = null
```
tamaskozmer, line 58: I think some of these don't need to be protected.

apps/teacher/src/main/java/com/instructure/teacher/view/TooltipView.kt
```diff
@@ -7,33 +22,40 @@ import android.annotation.SuppressLint
...
 
 abstract class TooltipView @JvmOverloads constructor(
         context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
 ) : View(context, attrs, defStyleAttr) {
 
+    /** Maximum line count before truncating */
+    private val MAX_LINES = 3
+
+    /** Maximum tooltip width in DP */
+    private val MAX_WIDTH_DP = 400
```
tamaskozmer, line 46: I thought these should be at the top of the file under the imports, in that case you can use the `const` keyword to make them compile time constants.

tamaskozmer (approved):
Great work, tested every previous issue and haven't found any problem. I added some additional comments, but you can merge it back.

PR #1102, opened 2020-11-16
[PDF][MBL-14954] Fixed bug with stamp color and duplicate annotations

Blackchai:
This PR fixes two bugs with the stamp/point annotations (looks like a pin). They would often not retain their color when updating or when created. In addition, duplicate annotations were occurring very regularly, this adds a race condition check to prevent them from being created twice.

refs: MBL-14954
affects: Teacher, Student
release note: Fixed a bug with the point annotation not retaining its color on updates

test plan: Create a stamp annotatation, any color other than blue. Exit the pdf and re-open it, confirm its still the same color. Now update the annotation, move it, resize it, etc. Exit/re-enter and confirm the color remains the same. For the duplicate fix, just check that there isn't a second one hiding underneath :).


tamaskozmer (approved):
Nice fix

PR #1106, opened 2020-11-18
[MBL-14899][Student][Teacher] New Quizzes assignments do not display on quiz index

tamaskozmer:
Testing: In the JIRA ticket

refs: MBL-14899
affects: Teacher, Student
release note: Fixed a bug where quizzes created with New Quizzes don't show in the list.

### DIFF
--- a/libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/QuizzesApiPactTests.kt
+++ b/libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/QuizzesApiPactTests.kt
@@ -43,7 +43,7 @@ class QuizzesApiPactTests : ApiPactTestBase() {
     //
 
     val getAllQuizzesTeacherQuery: String? = null
-    val getAllQuizzesTeacherPath = "/api/v1/courses/3/quizzes"
+    val getAllQuizzesTeacherPath = "/api/v1/courses/3/all_quizzes"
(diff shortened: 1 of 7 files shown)


hermannakos (approved):

libs/canvas-api-2/src/test/java/com/instructure/canvasapi2/pact/canvas/apis/QuizzesApiPactTests.kt
```diff
@@ -43,7 +43,7 @@ class QuizzesApiPactTests : ApiPactTestBase() {
     //
 
     val getAllQuizzesTeacherQuery: String? = null
-    val getAllQuizzesTeacherPath = "/api/v1/courses/3/quizzes"
+    val getAllQuizzesTeacherPath = "/api/v1/courses/3/all_quizzes"
```
joehoag, line 46: So, technically, we should re-publish our pact contracts after making this change, since the current pact tests run against the /quizzes endpoint, not the /all_quizzes endpoint.  And that would mean cloning canvas-web on your machine, changing some contracts, testing, potentially getting a canvas-web PR passed, and re-publishing our contracts. 

So I'm OK with leaving our contract as is for this minor change, with the knowledge that we aren't testing the /all_quizzes endpoint in pact.

joehoag (approved):
This generally looks good.  As noted, I'm a little queasy about the changes to our pact test logic, but I think that we can let them stand (without publishing new contracts) for now.

PR #1111, opened 2020-11-19
[Student][MBL-14960] Potential modules fix - needs product approval

Blackchai:
So the way our expandable recycler view code is set up doesn't play well with paginated modules _or_ the empty view stuff that was added back in when trying to "scroll to" a group position. This is a possible solution to get linking to modules linking fixed quickly, but it disables empty views and exhausts the pagination. A more involved solution is likely viable, but would require much more serious changes to the modules code. I think this one should go to product.

refs: MBL-14960
affects: Student
release note: Fixed a bug with links to modules

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/adapter/ModuleListRecyclerAdapter.kt
+++ b/apps/student/src/main/java/com/instructure/student/adapter/ModuleListRecyclerAdapter.kt
@@ -83,7 +83,7 @@ open class ModuleListRecyclerAdapter(
 
         }
         isExpandedByDefault = false
-        isDisplayEmptyCell = true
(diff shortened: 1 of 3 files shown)


tamaskozmer (approved):

Blackchai:
@tamaskozmer Do we have approval from product to merge this?

tamaskozmer:
To be honest, I totally missed the title here. Why do we need a product approval for this? Because we get all the modules instead of pagination? We can discuss this today on Slack, and after I will talk with József.

tamaskozmer:
We decided to merge this back and create a follow up ticket to fix the issue.

PR #1128, opened 2020-12-01
[MBL-14791][Parent] Manage remote config flags

hermannakos:


apps/flutter_parent/lib/screens/settings/settings_interactor.dart
```diff
@@ -28,6 +29,10 @@ class SettingsInteractor {
   void routeToThemeViewer(BuildContext context) {
     locator<QuickNav>().push(context, ThemeViewerScreen());
   }
+  
+  void routeToRemoteConfig(BuildContext context) {
+    locator<QuickNav>().push(context, RemoteConfigScreen());
```
tamaskozmer, line 34: This method is deprecated. Is there an easy option to change to the alternative solution?

apps/flutter_parent/test/screens/remote_config_params/remote_config_interactor_test.dart
```diff
@@ -0,0 +1,31 @@
...
+import '../../utils/platform_config.dart';
+import '../../utils/test_app.dart';
+import '../../utils/test_helpers/mock_helpers.dart';
+
+void main() {
+  test('getRemoteConfigParams returns correct map', () async {
+    final mockRemoteConfig = setupMockRemoteConfig(valueSettings: {'test_string': 'fetched value', 'mobile_verify_beta_enabled' : 'false'});
+    await setupPlatformChannels(
+        config: PlatformConfig(initRemoteConfig: mockRemoteConfig));
+    expect(RemoteConfigInteractor().getRemoteConfigParams(), equals({RemoteConfigParams.TEST_STRING: 'fetched value', RemoteConfigParams.MOBILE_VERIFY_BETA_ENABLED: 'false'}));
```
tamaskozmer, line 29: I think that you are verifying here, that the `getRemoteConfigParams` returns the values that you set to your mock. If that's right, this test will fail once you change the remote config name for one of those enum instances changed. I suggest using just inlined strings for this test.

apps/flutter_parent/test/utils/remote_config_utils_test.dart
```diff
@@ -70,4 +70,15 @@ void main() {
 
     expect(RemoteConfigUtils.getStringValue(RemoteConfigParams.TEST_STRING), 'fetched value');
   });
+
+  test('update cached value', () async {
```
tamaskozmer, line 74: Maybe the same applies to this test as to the other one, but I'm not sure about this one. However this won't fail because we just update the TEST_STRING, but I think your intention was here to update the `rc_test_string`.

tamaskozmer (changes requested):
Looks good, I added some minor comments, but I still need to test it.

tamaskozmer:
Tested, works as intended.

apps/flutter_parent/test/utils/remote_config_utils_test.dart
hermannakos: the rc_test_string is the key to the TEST_STRING in the SharedPreferences. We use rc_test_string to mock it, but when accessing it we want to use the enum value.

apps/flutter_parent/test/screens/remote_config_params/remote_config_interactor_test.dart
hermannakos: I cannot use the inlined strings because the functionality is based on the enum values. I agree that we should only use the TEST_STRING  since it is the only one for testing purposes only.

tamaskozmer (approved):

PR #1143, opened 2020-12-10
[Teacher][MBL-14864] Edit Syllabus

tamaskozmer:
The feature is ready, but I still need to add some interaction tests. It can be reviewed already, but please don't merge.

Test Plan: 
- Test editing the syllabus with a teacher.
- Test if teacher assistants does not have permission to edit syllabus, if this is turned off in the Canvas settings.
- Test success cases: in this case if the syllabus is succesfully edited, the user should return to the syllabus screen and it should be updated.
- Test error case: We should se an error message and remain on the screen.
- Test if going back without saving triggers confirmation dialog.

refs: MBL-14864
affects: Teacher
release note: Added the possibility to edit course syllabus.

### DIFF
--- a/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/EditSyllabusPageTest.kt
+++ b/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/EditSyllabusPageTest.kt
@@ -49,7 +49,7 @@ class EditSyllabusPageTest : TeacherTest() {
         goToEditSyllabus()
 
         editSyllabusPage.editSyllabusBody("Syllabus edited")
-        editSyllabusPage.changeShowSummary()
+        editSyllabusPage.toggleShowSummary()
         editSyllabusPage.saveSyllabusEdit()
 
         syllabusPage.assertDisplaysSyllabus("Syllabus edited", false)
--- a/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/EditSyllabusPageTest.kt
+++ b/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/EditSyllabusPageTest.kt
@@ -1,0 +1,101 @@
+/*
+ * Copyright (C) 2020 - present Instructure, Inc.
+ *
+ *     This program is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU General Public License as published by
+ *     the Free Software Foundation, version 3 of the License.
+ *
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package com.instructure.teacher.ui
+
+import com.instructure.canvas.espresso.mockCanvas.*
+import com.instructure.canvasapi2.models.Assignment
+import com.instructure.canvasapi2.models.CanvasContextPermission
+import com.instructure.canvasapi2.models.CourseSettings
+import com.instructure.canvasapi2.models.Tab
+import com.instructure.dataseeding.util.days
+import com.instructure.dataseeding.util.fromNow
+import com.instructure.dataseeding.util.iso8601
+import com.instructure.espresso.ActivityHelper
+import com.instructure.teacher.ui.utils.TeacherTest
+import com.instructure.teacher.ui.utils.tokenLogin
+import org.junit.Test
+
+class EditSyllabusPageTest : TeacherTest() {
+
+    override fun displaysPageObjects() = Unit
+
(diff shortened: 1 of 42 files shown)


hermannakos:
The intent is not fired after I give every permission to take images, I have to click the action buttons again.

apps/teacher/src/main/java/com/instructure/teacher/features/syllabus/ui/SyllabusFragment.kt
```diff
@@ -40,6 +40,16 @@ class SyllabusFragment : MobiusFragment<SyllabusModel, SyllabusEvent, SyllabusEf
 
     override fun makeInitModel() = SyllabusModel(canvasContext.id)
 
+    override fun onStart() {
+        super.onStart()
+        view.registerEventBus()
```
hermannakos, line 45: It feels like to me that we should move this to the view to onConnect/onDispose. I may be wrong here, but it feels weird to register/unregister it outside.

apps/teacher/src/main/res/layout/fragment_edit_syllabus.xml
```diff
@@ -0,0 +1,133 @@
...
+                app:rce_editor_padding="12dp" />
+
+            <include
+                android:id="@+id/divider"
+                layout="@layout/view_divider"
+                android:layout_width="match_parent"
+                android:layout_height="0.5dp"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toBottomOf="@id/contentRCEView"
+                tools:layout_editor_absoluteX="-6dp"
```
hermannakos, line 87: I know this is just for the editor but these numbers seem weird to me.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/CourseManager.kt
```diff
@@ -187,6 +203,20 @@ object CourseManager {
         CourseAPI.updateCourse(courseId, queryParams, adapter, callback, params)
     }
 
+    fun editCourseSyllabusAsync(courseId: Long, syllabusBody: String): Deferred<DataResult<Course>> {
+        return apiAsync { editCourseSyllabus(courseId, syllabusBody, it) }
+    }
+
+    private fun editCourseSyllabus(courseId: Long, syllabusBody: String, callback: StatusCallback<Course>) {
+        val queryParams = HashMap<String, String>()
+        queryParams["course[syllabus_body]"] = syllabusBody
```
hermannakos, line 212: We've already talked about this. I don't prefer to use queryParams here, we'll see if we can fix this.

hermannakos (commented):
Awesome! Works great, some minor issues we've already talked about.

apps/teacher/src/main/java/com/instructure/teacher/features/syllabus/ui/SyllabusFragment.kt
tamaskozmer: It was there first, I spent half a day debugging, and it seems like in that case when the event is fired the EventBus is not already registered, and we can't get the event. Mobius magic.

apps/teacher/src/main/res/layout/fragment_edit_syllabus.xml
tamaskozmer: I don't know why, but if I don't set absolute coordinates this view will just show in the top of the editor, and everything as well what is constrainted to this.

libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/managers/CourseManager.kt
tamaskozmer: On it

apps/teacher/src/androidTest/java/com/instructure/teacher/ui/EditSyllabusPageTest.kt
```diff
@@ -0,0 +1,101 @@
...
+        syllabusPage.assertDisplaysSyllabus("Syllabus edited", true)
+        syllabusPage.assertSuccessfulSave(ActivityHelper.currentActivity())
+    }
+
+    @Test
+    fun testSaveEditedSyllabusWhenShowSummaryWasDisabled() {
+        goToEditSyllabus()
+
+        editSyllabusPage.editSyllabusBody("Syllabus edited")
+        editSyllabusPage.changeShowSummary()
```
hermannakos, line 52: It's not clear what state should the summary be in after changing it. Maybe it could be clarified.

tamaskozmer: I can't really clarify it, since it performs a click on the switch, which toggles it's state. Maybe toggleShowSummary could be a more descriptive name.

hermannakos:
If you add a syllabus description and after deleting everything and leaving it blank it successfully saves it and moves back to the empty description screen. If it is not the intended behaviour, then I think in this case we should remove that tab just like when we have no description.

hermannakos (approved):

PR #1156, opened 2020-12-22
[MBL-15058][Student] Assignment Details | Not graded assignments

tamaskozmer:
Test plan: Testing steps and test user are in the linked bug ticket. It would be also good to test different types of graded assignments, if everything else is working as it worked before.

refs: MBL-15058
affects: Student
release note: Corrected submission details view for not graded assignments.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/ui/SubmissionDetailsEmptyContentViewState.kt
+++ b/apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/ui/SubmissionDetailsEmptyContentViewState.kt
@@ -17,9 +17,9 @@ package com.instructure.student.mobius.assignmentDetails.submissionDetails.conte
 
 sealed class SubmissionDetailsEmptyContentViewState {
     data class Loaded(
-        val isAllowedToSubmit: Boolean = true,
-        val dueDateText: String = "",
         val submitButtonText: String = "",
-        val isObserver: Boolean = false
+        val emptyViewTitleText: String = "",
+        val emptyViewSubtitleText: String = "",
+        val submitButtonVisible: Boolean = false
     ) : SubmissionDetailsEmptyContentViewState()
 }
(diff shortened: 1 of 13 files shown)


apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/ui/SubmissionDetailsEmptyContentViewState.kt
```diff
@@ -17,9 +17,9 @@ package com.instructure.student.mobius.assignmentDetails.submissionDetails.conte
 
 sealed class SubmissionDetailsEmptyContentViewState {
     data class Loaded(
-        val isAllowedToSubmit: Boolean = true,
-        val dueDateText: String = "",
         val submitButtonText: String = "",
-        val isObserver: Boolean = false
+        val emptyViewTitleText: String = "",
```
hermannakos, line 21: I would prefer nullable properties instead of empty strings, but this is fine as well.

libs/pandares/src/main/res/values/strings.xml
```diff
@@ -93,6 +94,7 @@
     <string name="submissionDetailsAssignmentLockedByModuleName">Your assignment is locked by module \"%1$s\"</string>
     <string name="submissionDetailsAssignmentLockedByModulePrereqs">Your assignment is locked by a module requirement</string>
     <string name="submissionDetailsAssignmentLocked">Assignment Locked</string>
+    <string name="submissionDetailsNoSubmissionAllowed">No Submission Allowed</string>
```
hermannakos, line 97: I don't think we should start every word with capital letters here.

hermannakos (approved):
LGTM!

apps/student/src/main/java/com/instructure/student/mobius/assignmentDetails/submissionDetails/content/emptySubmission/ui/SubmissionDetailsEmptyContentViewState.kt
tamaskozmer: These will never be null, so it wouldn't be good to make them nullable. I can remove the default values, but I would need to adjust a lot of unit tests by adding some boilerplate to tests where I only check one property. In general I think it's not ideal to use nullable values in these view state object, because it will introduce additional logic in the views, so it could only be covered by render tests instead of unit tests.

libs/pandares/src/main/res/values/strings.xml
tamaskozmer: Right, it's lowercase in the ticket. I will ask Józsi about this, because the string this replaces are all capital letters.

PR #1186, opened 2021-01-26
[MBL-15139][Teacher] Move TouchTarget delegate to extension

tamaskozmer:
Test plan: Test if the touch is detected on the SpeedGrader buttons when clicking slightly outside of the border. Show pointer location setting can be helpful.

refs: MBL-15139
affects: Teacher
release note: -


hermannakos (approved):
Really nice!

PR #1199, opened 2021-02-03
[MBL-15116][Student] Hard-completed enrollments cannot be accessed

tamaskozmer:
Test plan: In the ticket. You can compare the courses in the all courses list with the web. It would be also good to test all the screens which were affected by the api call change.

refs: MBL-15116
affects: Student
release note: Fixed a bug where hard completed courses didn't show up in course list.


hermannakos (approved):
LGTM

PR #1260, opened 2021-04-14
[MBL-15188][Student] Dashboard edit

hermannakos:
refs: MBL-15188
affects: Student
release note: Updated functionality for editing the Dashboard.

test plan: See ticket for use cases.


apps/student/src/test/java/com/instructure/student/features/dashboard/edit/EditDashboardViewModelTest.kt
```diff
@@ -0,0 +1,593 @@
...
+import org.junit.Assert.*
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.threeten.bp.DateTimeUtils
+import org.threeten.bp.OffsetDateTime
+import org.threeten.bp.ZoneId
+
+@ExperimentalCoroutinesApi
+class EditDashboardViewModelTest {
```
tamaskozmer, line 53: I would also add some tests for the case when we successfully get courses and groups from the service, the data is correctly mapped and ItemViewModels are created with the correct data. There is a fair amount of logic there when we create the different categories, some of that might be covered with tests.

tamaskozmer (changes requested):

tamaskozmer:
Looks good, great work. Most of the requested changes are minor/styling issues.

apps/student/src/main/res/layout/fragment_edit_dashboard.xml
```diff
@@ -22,9 +22,12 @@
         <variable
             name="viewModel"
             type="com.instructure.student.features.dashboard.edit.EditDashboardViewModel" />
+
+        <import type="android.view.View" />
```
tamaskozmer, line 26: This import seems unnecessary

libs/pandautils/src/main/java/com/instructure/pandautils/binding/BindingAdapters.kt
```diff
@@ -58,6 +59,11 @@ fun bindItemViewModels(recyclerView: RecyclerView, itemViewModels: List<ItemView
     adapter.updateItems(itemViewModels)
 }
 
+@BindingAdapter("refreshState")
+fun bindRefreshState(swipeRefreshLayout: SwipeRefreshLayout, state: ViewState?) {
+    swipeRefreshLayout.isRefreshing = state == ViewState.Loading
```
tamaskozmer, line 64: Instead of this I would rather create a `ViewState.Refresh` state that can indicate that the user is refreshing the content. With this we can handle the initial loading and the refresh separately and we can show the normal loading indicator on initial loading. You could also eliminate all the changes to the `EmptyView` class.

tamaskozmer (changes requested):

tamaskozmer (approved):

PR #1273, opened 2021-05-03
[MBL-15343][Student] Change font if K5

hermannakos:
refs: MBL-15343
affects: Student
release note: None

test plan: Check the fonts if you login as a standard vs. as a K5 user.


libs/pandautils/src/main/java/com/instructure/pandautils/typeface/TypefaceUtil.kt
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.instructure.pandautils.typeface
+
+import android.content.Context
+import android.graphics.Typeface
+import java.lang.reflect.Field
+
+object TypefaceUtil {
```
tamaskozmer, line 23: I think we should include this logic into the `TypefaceBehavior` class, don't need to create an other singleton util for this.

apps/student/src/main/java/com/instructure/student/activity/NavigationActivity.kt
```diff
@@ -204,6 +209,8 @@ class NavigationActivity : BaseRouterActivity(), Navigation, MasqueradingDialog.
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        TypefaceUtil.overrideFont(applicationContext, typefaceBehavior.typefaceMap)
+        setContentView(contentResId())
```
tamaskozmer, line 213: This seems odd. Why do we need to call `setContentView` again? I guess it won't apply the fonts after setting the content view. In that case would it make sense to call `overrideFont` before the `super.onCreate`?

tamaskozmer (changes requested):

tamaskozmer (approved):

PR #1305, opened 2021-06-14
[MBL-15431][Student] K5 subject cards positioning

tamaskozmer:
Test plan: Verify that the we show 1 subject card in portrait mode and 2 in landscape.

refs: MBL-15431
affects: Student
release note: none


libs/pandautils/src/main/java/com/instructure/pandautils/features/elementary/homeroom/HomeroomFragment.kt
```diff
@@ -87,11 +87,7 @@ class HomeroomFragment : Fragment() {
 
                 val calculatedSpan = (width - spacing * 4) / cardRequiredSpace
 
-                val span = when {
-                    calculatedSpan < 3 -> 2
-                    calculatedSpan >= 4 -> 4
-                    else -> 3
-                }
+                val span = if (calculatedSpan < 2) 1 else 2
```
hermannakos, line 90: Do we need this span calculation still, wouldn't it be safer to set it based on the configuration?

tamaskozmer: We would still need to create multiple configurations, because of the tablet behavior. It will be different on devices with sw-720dp and on devices that are below it.

hermannakos (approved):

PR #1326, opened 2021-07-14
add Vietnamese translation

annhienktuit:


hermannakos:
Sorry we cannot accept translations from outside contributors, since we have no means to validate them.

PR #1329, opened 2021-07-19
[MBL-15422][Student] K5 Schedule content

hermannakos:
refs: MBL-15422
affects: Student
release note: none


libs/pandautils/src/main/res/layout-large/item_schedule_course.xml
```diff
@@ -0,0 +1,100 @@
...
+  ~
+  ~ This program is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
```
tamaskozmer, line 16: Are you using the `layout-large` qualifier for the tablet design, or is there some other requirement that requires this? If it's the first, please use the `sw720dp` qualifier to remain consistent with the other parts of the Homeroom screens.

libs/pandautils/src/main/java/com/instructure/pandautils/features/elementary/schedule/ScheduleViewModel.kt
```diff
@@ -0,0 +1,384 @@
...
+        getData()
+        jumpToToday()
+    }
+
+    fun refresh() {
+        _state.postValue(ViewState.Refresh)
+        getData()
+    }
+
+    fun jumpToToday() {
```
tamaskozmer, line 88: This function should be private.

libs/pandautils/src/main/java/com/instructure/pandautils/features/elementary/schedule/ScheduleViewModel.kt
```diff
@@ -0,0 +1,384 @@
...
+        _events.postValue(Event(ScheduleAction.JumpToToday(todayPos)))
+    }
+
+    private fun getData() {
```
tamaskozmer, line 92: I think we should have a force network parameter here, we probably don't want to load from network every time we open the page.

libs/pandautils/src/main/java/com/instructure/pandautils/features/elementary/schedule/ScheduleViewModel.kt
```diff
@@ -0,0 +1,384 @@
...
+
+                plannerItems = plannerManager.getPlannerItemsAsync(
+                        true,
+                        weekStart.toApiString(),
+                        DateHelper.getNextSaturday(startDate).toApiString())
+                        .await()
+                        .dataOrNull
+                        .orEmpty()
+
+                assignmentMap = plannerItems
```
tamaskozmer, line 111: Do we need to fetch the assignments here for each planner item? I see that we only get these to determine the chips to show, but most of those states can be found under the `planner_item/submission` object. The only thing I not found is how to get the `replies` chip, what only applies to discussions. Maybe there is a way to determine that as well. It would worth a try asking the web team about this, since they didn't mention in the API document that we have to request each individual assignment. 

If this is not possible we can still eliminate a lot of unnecessary network requests by requesting only the discussions instead of all the assignments. We can know from the `PlannableType`, which items are discussions.

libs/pandautils/src/main/java/com/instructure/pandautils/features/elementary/schedule/ScheduleViewModel.kt
```diff
@@ -0,0 +1,384 @@
...
+            PlannableType.ASSIGNMENT -> _events.postValue(Event(ScheduleAction.OpenAssignment(plannerItem.canvasContext, plannerItem.plannable.id)))
+            PlannableType.CALENDAR_EVENT -> _events.postValue(Event(ScheduleAction.OpenCalendarEvent(plannerItem.canvasContext, plannerItem.plannable.id)))
+            PlannableType.DISCUSSION_TOPIC -> _events.postValue(Event(ScheduleAction.OpenDiscussion(plannerItem.canvasContext, plannerItem.plannable.id, plannerItem.plannable.title)))
+            PlannableType.QUIZ -> {
+                if (plannerItem.plannable.assignmentId != null) {
+                    // This is a quiz assignment, go to the assignment page
+                    _events.postValue(Event(ScheduleAction.OpenAssignment(plannerItem.canvasContext, plannerItem.plannable.id)))
+                } else {
+                    var htmlUrl = plannerItem.htmlUrl.orEmpty()
+                    if (htmlUrl.startsWith('/')) htmlUrl = ApiPrefs.fullDomain + htmlUrl
```
tamaskozmer, line 333: It would be better to inject the `ApiPrefs` through the `ViewModel` constructor.

tamaskozmer (changes requested):

libs/pandautils/src/main/java/com/instructure/pandautils/binding/BindingAdapters.kt
```diff
@@ -124,3 +128,20 @@ fun bindImageWithOverlay(imageView: ImageView, imageUrl: String?, overlayColor:
     }
 }
 
+@BindingAdapter(value = ["borderColor", "borderWidth", "backgroundColor", "borderCornerRadius"], requireAll = false)
+fun addBorderToContainer(view: View, borderColor: Int?, borderWidth: Int?, backgroundColor: Int?, borderCornerRadius: Int?) {
```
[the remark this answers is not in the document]
hermannakos: It would be easier but reusability would be an issue then. We would do the same thing as we do here only with drawables in that case.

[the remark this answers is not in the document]
tamaskozmer: This might be a bit more reusable, but generally the responsibility of Data Binding and Binding Adapters is to bind some kind of data to the view, and not to set the style of the view. I think setting a color can be considered Data Binding since color is dynamic and we receive it from the API, but adding a background is part of the design. I think this is not a big issue, so you can leave it like this, if this would require big effort to change.

tamaskozmer (approved):

PR #1345, opened 2021-08-11
[Student][MBL-15573] Resources screen - Important Info section

tamaskozmer:
Test plan: Specification and the new design for multiple homeroom courses are in the ticket. Also verify when only 1 course has important links, the header is not shown. Easiest way to do this is to remove the important links content from the Homeroom 2 course of the test user.

refs: MBL-15573
affects: Student
release note: none


balintbartok (approved):
QA + 1

PR #1363, opened 2021-09-03
[MBL-15350][Teacher] Added actions to the end labels for SpeedGrader

hermannakos:
refs: MBL-15350
affects: Teacher
release note: Added the option to click the end labels of the SpeedGrader slider.

test plan: Tap on the min/max value, the grade should change.

### DIFF
--- a/apps/teacher/src/main/java/com/instructure/teacher/view/grade_slider/SpeedGraderSlider.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/view/grade_slider/SpeedGraderSlider.kt
@@ -78,7 +78,7 @@ class SpeedGraderSlider @JvmOverloads constructor(
             accessibleTouchTarget()
         }
 
-        maxGradeValue.apply {
+        maxGrade.apply {
             setOnClickListener {
                 updateGrade(maxGradeValue)
                 notGraded = false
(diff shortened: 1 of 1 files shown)


apps/teacher/src/main/java/com/instructure/teacher/view/grade_slider/SpeedGraderSlider.kt
```diff
@@ -68,7 +69,25 @@ class SpeedGraderSlider @JvmOverloads constructor(
...
+        minGrade.apply {
+            setOnClickListener {
+                updateGrade(0)
+                notGraded = false
+                isExcused = false
+            }
+            accessibleTouchTarget()
+        }
+
+        maxGradeValue.apply {
```
tamaskozmer, line 81: This should be `maxGrade`, because right now this is a reference to the Int value you defined and not to the actual view. Unfortunately AS still allows you to call the `setOnClickListener` and `accessibleTouchTarget` functions in the apply block, because we are in the scope of the `SpeedGraderSlider`. In this case it will be applied to the whole slider, you can test it by clicking under the slider, it will jump to the max grade value.

tamaskozmer (changes requested):

tamaskozmer (approved):

PR #1382, opened 2021-10-14
fix testDashboardE2E test navigate back button, and stabilize test

kdeakinstructure:
Navigate Up button was called from the wrong page, it has been refactored.
Also, there was some refactor to make the test more stabile.


tamaskozmer (approved):

inst-danger:
<p>Teacher Install Page</p>

PR #1388, opened 2021-10-21
[MBL-15659][Student] Populate subject page WebViews

hermannakos:
refs: MBL-15659
affects: Student
release note: none

test plan: Check if the correct tab loads for each course. The LTI tools should not be visible. Check if the state is persisted on config change. Sometimes the webviews can log you out, it's going to be fixed in another ticket before going to production.


inst-danger:
<p>Teacher Install Page</p>

PR #1398, opened 2021-11-05
[MBL-15605][Parent] Keyboard does not open on login screens when using Android 12

tamaskozmer:
Test plan: Bug description is in the ticket. This should be verified on Android 12, but test also lower versions if WebViews work correctly.

refs: MBL-15605
affects: Parent
release note: Fixed a bug where keyboard didn't open on the login screen using Android 12.


inst-danger:
<p>Parent Install Page</p>

hermannakos (approved):

PR #1412, opened 2021-11-18
[MBL-15719][Parent] Omit userId from enrolments call

hermannakos:
refs: MBL-15719
affects: Parent
release note: Fixed a bug where total grades were not visible for past and future enrolments.

test plan: See ticket.


inst-danger:
<p>Parent Install Page</p>

tamaskozmer (approved):

PR #1471, opened 2022-02-02
[MBL-15817][Student] Important dates interaction tests

hermannakos:
refs: MBL-15817
affects: Student
release note: none

test plan:

### DIFF
--- a/apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
+++ b/apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
@@ -37,6 +37,7 @@ class ImportantDatesInteractionTest : StudentTest() {
     override fun displaysPageObjects() = Unit
 
     @Test
+    //The UI is different on tablet, so we only check the phone version
     @StubTablet
     @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
     fun testShowCalendarEvents() {
@@ -76,7 +77,7 @@ class ImportantDatesInteractionTest : StudentTest() {
 
     @Test
     @StubTablet
-    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
+    @TestMetaData(Priority.P1, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
     fun testPullToRefresh() {
         val data = createMockData(courseCount = 1)
         val course = data.courses.values.toList()[0]
@@ -91,7 +92,7 @@ class ImportantDatesInteractionTest : StudentTest() {
 
     @Test
     @StubTablet
-    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
+    @TestMetaData(Priority.P1, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
     fun testOpenCalendarEvent() {
         val data = createMockData(courseCount = 1)
         val course = data.courses.values.toList()[0]
@@ -99,14 +100,14 @@ class ImportantDatesInteractionTest : StudentTest() {
 
         goToImportantDatesTab(data)
         importantDatesPage.assertItemDisplayed(event.title!!)
-        importantDatesPage.clickItem(event.title!!)
+        importantDatesPage.clickImportantDatesItem(event.title!!)
         calendarEventPage.verifyTitle(event.title!!)
         calendarEventPage.verifyDescription(event.description!!)
     }
 
     @Test
     @StubTablet
-    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
+    @TestMetaData(Priority.P1, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
     fun testOpenAssignment() {
         val data = createMockData(courseCount = 1)
         val course = data.courses.values.toList()[0]
@@ -116,7 +117,7 @@ class ImportantDatesInteractionTest : StudentTest() {
 
         goToImportantDatesTab(data)
         importantDatesPage.assertItemDisplayed(assignment.name!!)
-        importantDatesPage.clickItem(assignment.name!!)
+        importantDatesPage.clickImportantDatesItem(assignment.name!!)
         assignmentDetailsPage.verifyAssignmentDetails(assignment)
     }
 
--- a/apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
+++ b/apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
@@ -1,0 +1,144 @@
+/*
+ * Copyright (C) 2022 - present Instructure, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, version 3 of the License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.instructure.student.ui.interaction
+
+import com.instructure.canvas.espresso.StubTablet
+import com.instructure.canvas.espresso.mockCanvas.*
+import com.instructure.canvasapi2.models.Assignment
+import com.instructure.dataseeding.util.days
+import com.instructure.dataseeding.util.fromNow
+import com.instructure.dataseeding.util.iso8601
+import com.instructure.panda_annotations.FeatureCategory
+import com.instructure.panda_annotations.Priority
+import com.instructure.panda_annotations.TestCategory
+import com.instructure.panda_annotations.TestMetaData
+import com.instructure.student.ui.pages.ElementaryDashboardPage
+import com.instructure.student.ui.utils.StudentTest
+import com.instructure.student.ui.utils.tokenLoginElementary
+import dagger.hilt.android.testing.HiltAndroidTest
+import org.junit.Test
+
+@HiltAndroidTest
+class ImportantDatesInteractionTest : StudentTest() {
+    override fun displaysPageObjects() = Unit
+
+    @Test
+    @StubTablet
+    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
+    fun testShowCalendarEvents() {
+        val data = createMockData(courseCount = 1)
+        val course = data.courses.values.toList()[0]
+
+        val event = data.addCourseCalendarEvent(course.id, 2.days.fromNow.iso8601, "Important event", "Important event description", true)
+
+        goToImportantDatesTab(data)
+        importantDatesPage.assertItemDisplayed(event.title!!)
+    }
+
+    @Test
+    @StubTablet
+    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
+    fun testShowAssignment() {
+        val data = createMockData(courseCount = 1)
+        val course = data.courses.values.toList()[0]
+
+        val assignment = data.addAssignment(courseId = course.id, submissionType = Assignment.SubmissionType.ONLINE_TEXT_ENTRY)
+        data.addAssignmentCalendarEvent(course.id, 2.days.fromNow.iso8601, assignment.name!!, assignment.description!!, true, assignment)
+
+        goToImportantDatesTab(data)
+        importantDatesPage.assertItemDisplayed(assignment.name!!)
+    }
+
+    @Test
+    @StubTablet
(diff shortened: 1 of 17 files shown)


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
```diff
@@ -0,0 +1,144 @@
...
+import com.instructure.student.ui.utils.StudentTest
+import com.instructure.student.ui.utils.tokenLoginElementary
+import dagger.hilt.android.testing.HiltAndroidTest
+import org.junit.Test
+
+@HiltAndroidTest
+class ImportantDatesInteractionTest : StudentTest() {
+    override fun displaysPageObjects() = Unit
+
+    @Test
```
kdeakinstructure, line 39: A comment would be nice in which we explain why we don't run these on tablet (I know, because it's design is different than from the mobile, but we might forgot this in the future)

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    @StubTablet
+    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
+    fun testShowCalendarEvents() {
+        val data = createMockData(courseCount = 1)
+        val course = data.courses.values.toList()[0]
+
+        val event = data.addCourseCalendarEvent(course.id, 2.days.fromNow.iso8601, "Important event", "Important event description", true)
+
+        goToImportantDatesTab(data)
```
kdeakinstructure, line 48: Please use our general tab navigating method, you can see it within the ElementaryDashboardPage class (selectTab method)

image

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
```diff
@@ -0,0 +1,144 @@
...
+        val data = createMockData(courseCount = 1)
+
+        goToImportantDatesTab(data)
+
+        importantDatesPage.assertEmptyViewDisplayed()
+    }
+
+    @Test
+    @StubTablet
+    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
```
kdeakinstructure, line 79: I'd take this refresh test case to P1 priority as we marked in the other classes. (also, the name could be simply testRefresh(), we used this name in the other classes)

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
```diff
@@ -0,0 +1,144 @@
...
+        val assignment = data.addAssignment(courseId = course.id, submissionType = Assignment.SubmissionType.ONLINE_TEXT_ENTRY)
+        data.addAssignmentCalendarEvent(course.id, 2.days.fromNow.iso8601, assignment.name!!, assignment.description!!, true, assignment)
+
+        goToImportantDatesTab(data)
+        importantDatesPage.assertItemDisplayed(assignment.name!!)
+        importantDatesPage.clickItem(assignment.name!!)
+        assignmentDetailsPage.verifyAssignmentDetails(assignment)
+    }
+
+    private fun goToImportantDatesTab(data: MockCanvas) {
```
kdeakinstructure, line 123: If u use our general tab navigating method which I mentioned above, this won't be needed here, so it can be removed

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
```diff
@@ -0,0 +1,144 @@
...
+        val eventToCheck = data.addCourseCalendarEvent(course.id, 2.days.fromNow.iso8601, "Important event 2", "Important event 2 description", true)
+
+        importantDatesPage.pullToRefresh()
+        importantDatesPage.assertItemDisplayed(eventToCheck.title!!)
+    }
+
+    @Test
+    @StubTablet
+    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
+    fun testOpenCalendarEvent() {
```
kdeakinstructure, line 95: I'm seeing two ways here:

1. To merge this method with the testShowCalendarEvents method, and leave it on P0 priority.

2. They remain separate methods because testShowCalendarEvents is happening directly on the Important Dates page, so that's mandatory to check in these tests, so that would remain on P0, but this testOpenCalendarEvent case should be on P1, since it's just and "indirect", "secondary link" function of the Important Dates function.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
```diff
@@ -0,0 +1,144 @@
...
+        importantDatesPage.assertItemDisplayed(event.title!!)
+        importantDatesPage.clickItem(event.title!!)
+        calendarEventPage.verifyTitle(event.title!!)
+        calendarEventPage.verifyDescription(event.description!!)
+    }
+
+    @Test
+    @StubTablet
+    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
+    fun testOpenAssignment() {
```
kdeakinstructure, line 110: Same as with the CalendarEvent tests

apps/student/src/androidTest/java/com/instructure/student/ui/pages/ImportantDatesPage.kt
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.instructure.student.ui.pages
+
+import com.instructure.espresso.assertDisplayed
+import com.instructure.espresso.click
+import com.instructure.espresso.page.*
```
kdeakinstructure, line 21: Hint: You can increase the threshold of the imports to 1000 (or some big enough number) so it won't replace the direct imports automatically with *. (It would be clearer to see the classes which are really used). But that's just a personal preference from my side...:)

apps/student/src/androidTest/java/com/instructure/student/ui/pages/ImportantDatesPage.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    fun assertEmptyViewDisplayed() {
+        onView(withId(R.id.importantDatesEmptyView)).assertDisplayed()
+    }
+
+    fun pullToRefresh() {
+        onView(withId(R.id.importantDatesRecyclerView)).swipeDown()
+    }
+
+    fun clickItem(title: String) {
```
kdeakinstructure, line 40: I'd name this as "clickOnCalendarItem", would be a bit more accurate when we read the code

automation/espresso/src/main/kotlin/com/instructure/canvas/espresso/StubTabletAnnotation.kt
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.instructure.canvas.espresso
+
+@Target(AnnotationTarget.FUNCTION)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class StubTablet
```
kdeakinstructure, line 21: Cool! :)

automation/espresso/src/main/kotlin/com/instructure/canvas/espresso/mockCanvas/MockCanvas.kt
```diff
@@ -550,7 +551,32 @@ fun MockCanvas.addCourseCalendarEvent(courseId: Long, date: String, title: Strin
...
+    )
+
+    var calendarEventList = courseCalendarEvents[courseId]
+    if(calendarEventList == null) {
+        calendarEventList = mutableListOf<ScheduleItem>()
+        courseCalendarEvents[courseId] = calendarEventList
+    }
+    calendarEventList.add(newScheduleItem)
+
+    return newScheduleItem
```
kdeakinstructure, line 565: I'd consider to return with the whole calendarEventList here, so we could get the any element from it on the caller side.

automation/espresso/src/main/kotlin/com/instructure/canvas/espresso/mockCanvas/MockCanvas.kt
```diff
@@ -550,7 +551,32 @@ fun MockCanvas.addCourseCalendarEvent(courseId: Long, date: String, title: Strin
...
+}
+
+fun MockCanvas.addAssignmentCalendarEvent(courseId: Long, date: String, title: String, description: String, isImportantDate: Boolean = false, assignment: Assignment): ScheduleItem {
```
kdeakinstructure, line 568: Can we have a short conversation about handling assignments and scheduleItems? Just to make sure we are on the same page and we think them the same.

kdeakinstructure: If we can discuss and agree in how we handle them, my above comment might won't be interesting any more.

kdeakinstructure (changes requested):

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
hermannakos: That's the one we use.

automation/espresso/src/main/kotlin/com/instructure/canvas/espresso/mockCanvas/MockCanvas.kt
hermannakos: We can get the whole list from the MockData, with this method we only care about the one that we just created.

kdeakinstructure (approved):

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
tamaskozmer: An even better solution would be to add a string parameter to the `StubTablet` annotations.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
tamaskozmer: I would vote for the 2nd option. We don't need to merge tests that test different things. The first thing tests if the items are showing, and this tests if it opens correctly. I think it's better to separate test cases, so if one fails we know where the problem is exactly.

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
```diff
@@ -0,0 +1,145 @@
...
+    @StubTablet
+    @TestMetaData(Priority.P0, FeatureCategory.K5_DASHBOARD, TestCategory.INTERACTION)
+    fun testShowCalendarEvents() {
+        val data = createMockData(courseCount = 1)
+        val course = data.courses.values.toList()[0]
+
+        val event = data.addCourseCalendarEvent(course.id, 2.days.fromNow.iso8601, "Important event", "Important event description", true)
+
+        goToImportantDatesTab(data)
+        importantDatesPage.assertItemDisplayed(event.title!!)
```
tamaskozmer, line 50: I see that there is no assertions that the date is visible. Using the `DateTimeProvider` in the ViewModel instead of hardcoding a Calendar instance could solve this, because in that case you can mock the time in tests. I don't think this is high prio right now, but maybe @kdeakinstructure you can take a look when you work on the test refinements.

tamaskozmer (approved):

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
kdeakinstructure: Good idea!

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
kdeakinstructure: Yeah, exactly. I didn't know if there is an opportunity to define attributes/parameters to custom annotations in Kotlin, but that's really nice.

apps/student/src/androidTest/java/com/instructure/student/ui/interaction/ImportantDatesInteractionTest.kt
kdeakinstructure: What you are saying is basically right generally, but in these particular cases, the situation is that one case is actually the subset of the other, so we test the same thing twice.

I'd advice that if we go with the 2nd way (to remain in separate test methods) to remove the assertion which has already done in the simple "show" case, thus we would really test another function (actually opening). Right now, if the assertItemDisplayed will fail (whether it was because a real bug, or just flakyness), the "opening" method will fail as well, so we won't know if the opening works well or not because the test method does not get there.

PR #1491, opened 2022-02-10
[Student][Teacher][MBL-15881] Some Android devices don't recognize available apps that can open external files

tamaskozmer:
Test plan: See ticket. Should be also tested in the Teacher app, because it had the same issue. With this fix the teacher works as well, but doesn't have the loading screen when opening a file (never had).

refs: MBL-15881
affects: Student, Teacher
release note: Fixed a bug where some linked files wouldn't open the associated apps.


kdeakinstructure (approved):
QA 👍  (Also checked on Teacher app, works well there too.)

hermannakos (approved):

PR #1497, opened 2022-02-17
[MBL-9773][Student] Module navigation buttons are not included when accessing module items from links

tamaskozmer:
Test plan: I added a test user in the JIRA ticket. Each type of module item there has 3 links, 1 that has `module_item_id` and 1 that doesn't. These two should work the same, navigation buttons should be visible. The 3rd link is a content that is not included in the course modules. This link should open the same page, but redirect to the standalone screen after that.

refs: MBL-9773
affects: Student
release note: Module items opened from links now show module navigation buttons.


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

PR #1505, opened 2022-02-24
[MBL-15930][Teacher] Crash in SlidingPanel in Speed Grader.

tamaskozmer:
Test plan: It's not easy to reproduce this issue, but the crash is pretty self-explanatory, so instead of trying to reproduce it I would suggest to smoke test the Speed Grader and the Comment Library if everything works as intended.

refs: MBL-15930
affects: Teacher
release note: Fixed a crash in Speed Grader


inst-danger:
<p>Teacher Install Page</p>

hermannakos (approved):

PR #1518, opened 2022-03-08
[MBL-15936][Student] Canvas for Elementary theme is applied to courses which don't have it enabled.

tamaskozmer:
Test plan: Description in the ticket. There are also more cases where this bug could happen. It can also happen when opening a non K5 course grades from the K5 grades page. In that case we should redirect to the standard grades page.

refs: MBL-15936
affects: Student
release note: Fixed a bug where elementary subject page would open instead of course page.

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/features/elementary/course/ElementaryCourseFragment.kt
+++ b/apps/student/src/main/java/com/instructure/student/features/elementary/course/ElementaryCourseFragment.kt
@@ -23,15 +23,20 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.viewModels
+import androidx.lifecycle.Observer
 import com.google.android.material.tabs.TabLayout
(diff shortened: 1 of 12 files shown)


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

apps/student/src/main/java/com/instructure/student/features/elementary/course/ElementaryCourseFragment.kt
```diff
@@ -92,6 +97,12 @@ class ElementaryCourseFragment : Fragment() {
 
             }
         })
+
+        viewModel.events.observe(viewLifecycleOwner, Observer { event ->
```
hermannakos, line 101: There's a weird flashing when redirecting to the standard subject page. If we move this and the data fetching to the onCreateView could that perhaps make it a bit smoother?

hermannakos (approved):

PR #1536, opened 2022-04-01
[MBL-15873][Student][Teacher] Layout selection for dashboard

tamaskozmer:
Test plan:
- Verify that both apps has the layout selection switch and it works correctly.
- The state of the layout should be persisted between app restarts.
- In the Teacher App the layout should always be the same on the dashboard and on the all courses screen.
- Landscape mode should have list view as well if it is enabled.
- Smoke test editing the dashboard cards with both layouts.

refs: MBL-15873
affects: Teacher, Student
release note: Added option to switch between grid view and list view on the dashboard.


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

kdeakinstructure (approved):
QA 👍

apps/student/src/main/java/com/instructure/student/fragment/DashboardFragment.kt
```diff
@@ -149,8 +154,40 @@ class DashboardFragment : ParentFragment() {
...
+            StudentPrefs.listDashboard = false
+        } else {
+            item.setIcon(R.drawable.ic_grid_dashboard)
+            item.setTitle(R.string.dashboardSwitchToGridView)
+            StudentPrefs.listDashboard = true
+        }
+
+        recyclerView.fadeAnimationWithAction {
+            courseColumns = if (StudentPrefs.listDashboard) 1 else resources.getInteger(R.integer.course_card_columns)
+            groupColumns = if (StudentPrefs.listDashboard) 1 else resources.getInteger(R.integer.group_card_columns)
```
hermannakos, line 187: This is going to be always 1. Should we even bother with this?

apps/student/src/main/java/com/instructure/student/fragment/DashboardFragment.kt
```diff
@@ -149,8 +154,40 @@ class DashboardFragment : ParentFragment() {
...
+            item.setTitle(R.string.dashboardSwitchToListView)
+            StudentPrefs.listDashboard = false
+        } else {
+            item.setIcon(R.drawable.ic_grid_dashboard)
+            item.setTitle(R.string.dashboardSwitchToGridView)
+            StudentPrefs.listDashboard = true
+        }
+
+        recyclerView.fadeAnimationWithAction {
+            courseColumns = if (StudentPrefs.listDashboard) 1 else resources.getInteger(R.integer.course_card_columns)
```
hermannakos, line 186: This is kinda magic numbery to me. We should probably move this to a resource as well.

apps/student/src/main/java/com/instructure/student/fragment/DashboardFragment.kt
```diff
@@ -174,12 +211,11 @@ class DashboardFragment : ParentFragment() {
 
     private fun configureRecyclerView() {
         // Set up GridLayoutManager
-        val courseColumns = resources.getInteger(R.integer.course_card_columns)
-        val groupColumns = resources.getInteger(R.integer.group_card_columns)
-        val totalColumns = courseColumns * groupColumns
+        courseColumns = if (StudentPrefs.listDashboard) 1 else resources.getInteger(R.integer.course_card_columns)
```
hermannakos, line 214: These could be local variables.

apps/student/src/main/java/com/instructure/student/fragment/DashboardFragment.kt
tamaskozmer: No. `group_card_columns` have different values for different sizes.

apps/student/src/main/java/com/instructure/student/fragment/DashboardFragment.kt
tamaskozmer: I don't consider 0 and 1 as magic numbers, most linters neither, but I can extract it to a constant.

apps/student/src/main/java/com/instructure/student/fragment/DashboardFragment.kt
tamaskozmer: We change these in the `changeDashboardLayout` function when the user taps the menu item, instead of calling this and configuring the whole `RecyclerView` again, so this needs to be a property.

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

hermannakos (approved):

PR #1570, opened 2022-05-03
Allow backups

upintheairsheep:
There's no reason for it to be off. It's not a game.


tamaskozmer:
Hi,

Thank you for creating a PR. Currently we won't add this change, because there is no data that needs to be backed up to Google cloud. The app only works after logging in, and all the necessary data is tied to the user, so once the users log in they will have all the data that they had on their previous device.

PR #1583, opened 2022-05-17
[MBL-16015][Student] To do list widget does not update changes

tamaskozmer:
Test plan: See ticket. There are 2 scenarios in the ticket. The 1st is not valid, it only occurred because no curse was favorited. The second one should be verified, but you cannot reproduce it anymore with the attached user, so you need to create the data yourself.

refs: MBL-16015
affects: Student
release note: Fixed a bug where To Do widget displayed completed assignments.


inst-danger:
<p>Student Install Page</p>

PR #1653, opened 2022-07-20
Fix breaking multi_api_level tests

kdeakinstructure:
Fix breaking render tests on 'multi_api_level' jobs. (It was stubbed on Teacher because of the same issue, now, both Teacher and Student were fixed so Stubs has removed as well.)


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

PR #1685, opened 2022-08-24
[All][MBL-16150] Update target API level to 12

tamaskozmer:
Test plan:
- Smoke test all 3 apps thoroughly on a device running Android 12.
- If something seems to be not working as intended also check it with the current production app on the same device. In this case we can identify if the problem is caused by the target SDK or it already affects Android 12 users.
- Focus more on the platform interactions:
  - Push notifications
  - Services (file upload, file download, file submission)
  - Widgets
  - Share extension

refs: MBL-16150
affects: All
release note: none


inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

kdeakinstructure (approved):
QA 👍

hermannakos (approved):

PR #1698, opened 2022-09-06
[MBL-16227][Parent] - Rename build apk names to contains the appname

kdeakinstructure:
add variant section

define an apk build name which contains the word 'parent' instead of using a default build name (app-debug.apk or app-release.apk)

add timestamp to debug builds to be more accurate

refs: MBL-16227
affects: Parent
release note:

test plan:


inst-danger:
<p>Parent Install Page</p>

kristofnemere (approved):

tamaskozmer (approved):

PR #1762, opened 2022-10-27
[MBL-16212][Teacher] - Add search, compose message, user roles check to PeopleE2E test in Teacher

kdeakinstructure:
Add search, compose message, user roles check to PeopleE2E test in Teacher.
Extend data seeding with parent seeding.
Refactor StudentContextPage, add PersonContextPage (not only students can be on the PeoplePage), so PersonContextPage will be the base class and StudentContextPage will be specified detailer class of a student.
Refactor some interaction test according to the changes above.

refs: MBL-16212
affects: Teacher
release note: none


apps/teacher/src/androidTest/java/com/instructure/teacher/ui/PersonContextPageTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    fun displaysTeacherInfo() {
+        val (teacher, course) = getToPersonContextPage()
+        personContextPage.assertDisplaysCourseInfo(course)
+        personContextPage.assertSectionNameView(PersonContextPage.UserRole.TEACHER)
+        personContextPage.assertPersonNameIsDisplayed(teacher.name)
+    }
+
+    private fun getToPersonContextPage(assignmentCount: Int = 1): Triple<CanvasUserApiModel, CourseApiModel, List<AssignmentApiModel>> {
```
tamaskozmer, line 46: Seems like you never use the returned assignment list, so you might as well change this to `Pair` and remove that.

tamaskozmer (approved):

apps/teacher/src/androidTest/java/com/instructure/teacher/ui/PersonContextPageTest.kt
kdeakinstructure: Good catch, thanks!

PR #1786, opened 2022-11-23
[MBL-16318][Student][Teacher] Show Acceptable Use Policy (AUP) when needed

tamaskozmer:
Test plan:
- Basic flow is covered in the ticket. The user most see the acceptable use policy when it is changed or not accepted yet.
- There are a lot of scenarios that need to be checked for both apps:
  - The user is already logged into the app: In this case when opening the app the AUP should be instantly displayed and the user cannot see the dashboard until acceptance. Closing or going back will logout the user.
  - The user is logging into the app with password and username: In this case the AUP should be displayed after the login and the user cannot see the dashboard until acceptance. Closing or going back will logout the user.
  - The user is logging into the app with previously saved user: In this case the AUP should be displayed after tapping the previously saved user and the user cannot see the dashboard until acceptance. Closing or going back will logout the user.
- It should be verified in all of these scenarios that accepting the AUP will navigate the user to the dashboard.
- The whole login flow was refactored a bit so it should be thoroughly smoke tested, since it is the most critical part of the app.
- Token invalidation should also be tested. See MBL-15391
- Test Masquerading
- Verify that accepting the terms with a K5 users works and the K5 dashboard is shown after acceptance.

refs: MBL-16318
affects: Student, Teacher
release note: Added acceptable use policy screen on login.

## Checklist

- [x] A11y checked
- [x] Approve from product or not needed


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

libs/login-api-2/src/main/res/layout/activity_acceptable_use_policy.xml
```diff
@@ -0,0 +1,147 @@
...
+                android:elevation="6dp"
+                app:theme="@style/ToolBarStyleDark" />
+
+            <TextView
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="24dp"
+                android:text="@string/acceptableUsePolicyDescription"
+                android:textSize="16sp"
+                android:textFontWeight="400"
```
kristofnemere, line 53: Attribute textFontWeight is only used in API level 28 and higher (current min is 26)

apps/student/src/main/java/com/instructure/student/features/login/StudentAcceptableUsePolicyRouter.kt
```diff
@@ -0,0 +1,55 @@
...
+        activity.startActivity(intent)
+    }
+
+    override fun startApp() {
+        PushNotificationRegistrationWorker.scheduleJob(activity, ApiPrefs.isMasquerading)
+
+        CookieManager.getInstance().flush()
+
+        val intent = Intent(activity, NavigationActivity.startActivityClass)
+        if (activity.getIntent() != null && activity.getIntent().extras != null) {
```
kristofnemere, line 44: You could use let in this case to make it prettier

libs/login-api-2/src/main/java/com/instructure/loginapi/login/LoginNavigation.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.instructure.loginapi.login.viewmodel.LoginResultAction
+import com.instructure.loginapi.login.viewmodel.LoginViewModel
+
+private const val CANVAS_FOR_ELEMENTARY = "canvas_for_elementary"
+
+abstract class LoginNavigation(
+    private val activity: FragmentActivity
+) {
+
+    protected abstract val checkElementary: Boolean
```
kristofnemere, line 31: This is always true, do we need it to be overridable?

libs/login-api-2/src/main/java/com/instructure/loginapi/login/features/acceptableusepolicy/AcceptableUsePolicyViewModel.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    val events: LiveData<Event<AcceptableUsePolicyAction>>
+        get() = _events
+    private val _events = MutableLiveData<Event<AcceptableUsePolicyAction>>()
+
+    init {
+        viewModelScope.launch {
+            try {
+                val termsOfService =
+                    userManager.getTermsOfServiceAsync(true).await().dataOrThrow.content ?: ""
```
kristofnemere, line 44: You could use String.orEmpty()

libs/login-api-2/src/main/java/com/instructure/loginapi/login/features/acceptableusepolicy/AcceptableUsePolicyViewModel.kt
```diff
@@ -0,0 +1,91 @@
...
+                _data.value = _data.value?.copy(policy = termsOfService)
+            } catch (e: Exception) {
```
kristofnemere, line 46: We should handle the error somehow

kristofnemere (approved):

libs/login-api-2/src/main/res/layout/activity_acceptable_use_policy.xml
tamaskozmer: Actually this doesn't do anything, I just copied the style from Figma, but the 2 texts were the same. I wanted to ask Ákos about how we use the font weights, but apparently we don't use these anywhere else so I will just remove it.

libs/login-api-2/src/main/java/com/instructure/loginapi/login/LoginNavigation.kt
tamaskozmer: Well, that should be `false` in the Teacher app 😄

libs/login-api-2/src/main/java/com/instructure/loginapi/login/features/acceptableusepolicy/AcceptableUsePolicyViewModel.kt
tamaskozmer: It's actually handled in a way that we have an initial state in the `_data` LiveData, so that will be the state in case of an error. We don't need any specific handling because all the data what we show on the screen is static, and we load something that can be seen only after a click. In that case we just try to reload that data.

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

kdeakinstructure (approved):
QA 👍

PR #1851, opened 2023-02-07
[MBL-16385][Student] - Refactor and create new Assignment E2E tests

kdeakinstructure:
- Extend some Assignment E2E test case (because of new assignment list page, and missing some assertions)

- Refactor an Assignment E2E test case to do the submission flow on the UI (at least in one test case, because we did not have a single test case which tested the submission flow on the UI....it was generated everywhere by the API, but it's a very important feature so it's essential to check it at least in one test case. I did not implemented the UI flow in all test cases because if one case checks it, it's enough to check the feature's basic flow, and the rest of the cases could remain with API because it's much more faster of course.)

- Implement a brand new test case which is focused on the multiple attempt selection on the Assignment List Page (which was the new feature of the assignment enhancement implemented yet)

refs: MBL-16385
affects: Student
release note: none


inst-danger:
<p>Student Install Page</p>

kristofnemere (approved):

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/AssignmentsE2ETest.kt
```diff
@@ -582,4 +595,78 @@ class AssignmentsE2ETest: StudentTest() {
...
+            courseId = course.id,
+            assignmentId = pointsTextAssignment.id,
+            studentToken = student.token,
+            fileIds = emptyList<Long>().toMutableList()
+        )
+
+        Log.d(STEP_TAG, "Refresh the page.")
+        assignmentDetailsPage.refresh()
+
+        Log.d(STEP_TAG, "Assert that when only there is one attempt, the spinner is displayed and the last/newest attempt is selected.")
```
tamaskozmer, line 664: I think this log step was the victim of copy paste :)

apps/student/src/androidTest/java/com/instructure/student/ui/pages/AssignmentDetailsPage.kt
```diff
@@ -96,6 +97,10 @@ open class AssignmentDetailsPage : BasePage(R.id.assignmentDetailsPage) {
         onView(withId(R.id.gradeCell)).scrollTo().click()
     }
 
+    fun assertSubmissionAndRubricLabel() {
+        onView(allOf(withId(R.id.submissionAndRubricLabel), withText(R.string.submissionAndRubric)))
```
tamaskozmer, line 101: Do we need the `withText` matcher?

tamaskozmer (changes requested):

apps/student/src/androidTest/java/com/instructure/student/ui/pages/AssignmentDetailsPage.kt
kdeakinstructure: We do not, for the selector I guess (althought I think it just makes it more accurate), but there is really a problem with this that i forgot to put 'assertDisplayed()' at the end.

inst-danger:
<p>Student Install Page</p>

tamaskozmer (approved):

PR #1907, opened 2023-03-22
Extend Submission API to handle attempts so that fixes the breaking FilesE2E tests

kdeakinstructure:
Fix breaking FilesE2E test in both apps (Teacher and Student). --> These tests broke because I've turned on the assignment enhancement feature flag on the mobileqa.beta environment (so that we can test the new assignment logic like attempt selection and comments belong to attempts, etc.)


inst-danger:
<p>Student Install Page</p>

automation/dataseedingapi/src/main/kotlin/com/instructure/dataseeding/api/SubmissionsApi.kt
```diff
@@ -197,7 +197,8 @@ object SubmissionsApi {
                                         url = null,
                                         body = null,
                                         userId = 0,
-                                        grade = null
+                                        grade = null,
+                                        attempt = assignment.attempt!!
```
kristofnemere, line 201: I think this is the only change in the PR we actually need to make it work, but let's have a chat about it

kristofnemere (changes requested):

inst-danger:
<p>Student Install Page</p>

kristofnemere (approved):

PR #1910, opened 2023-03-24
Fix breaking Teacher InboxE2E test

kdeakinstructure:
Fix breaking Teacher InboxE2E test. (Checking subject just like anywhere else, instead of recipient list when asserting a conversation)


inst-danger:
<p>Teacher Install Page</p>

tamaskozmer (approved):

PR #2009, opened 2023-05-31
Fix breaking PeopleE2E test

kdeakinstructure:
Fix breaking PeopleE2E test


inst-danger:
<p>Teacher Install Page</p>

PR #2032, opened 2023-06-20
[MBL-16705][Student] Clear Offline user data on logout

kristofnemere:
Test plan: offline db should be removed on logout, but not when switching user

refs: MBL-16705
affects: Student
release note: none


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

PR #2075, opened 2023-07-28
[MBL-16942][Teacher] - Fix PeopleListFragment accepts few letters as search query

kdeakinstructure:
Fix PeopleListFragment accepts few letters (1, 2) as search query.

refs: MBL-16942
affects: Teacher
release note: none

Test plan: Test 

## Screenshots

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

## Checklist

- [ ] Tested in dark mode
- [ ] Tested in light mode

### DIFF
--- a/apps/teacher/src/main/java/com/instructure/teacher/fragments/PeopleListFragment.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/fragments/PeopleListFragment.kt
@@ -184,8 +184,14 @@ class PeopleListFragment : BaseSyncFragment<User, PeopleListPresenter, PeopleLis
     override fun onQueryTextSubmit(query: String): Boolean = false
 
     override fun onQueryTextChange(newText: String): Boolean {
-        adapter.clear()
-        presenter.searchPeopleList(newText)
+        val searchQuery = newText.trim().takeIf { it.length > 2 }.orEmpty()
+        if(searchQuery.isBlank()) {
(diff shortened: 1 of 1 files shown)


inst-danger:
<p>Teacher Install Page</p>

apps/teacher/src/main/java/com/instructure/teacher/fragments/PeopleListFragment.kt
```diff
@@ -184,8 +184,14 @@ class PeopleListFragment : BaseSyncFragment<User, PeopleListPresenter, PeopleLis
     override fun onQueryTextSubmit(query: String): Boolean = false
 
     override fun onQueryTextChange(newText: String): Boolean {
-        adapter.clear()
-        presenter.searchPeopleList(newText)
+        val searchQuery = newText.trim().takeIf { it.length > 2 }.orEmpty()
```
tamaskozmer, line 187: This logic should be in the Presenter instead of the Fragment. Also there is a much easier way to check the length of the string. You can do this:
```
if(newText.length > 2) {
            presenter.searchPeopleList("")
}
```

tamaskozmer (changes requested):

kdeakinstructure:
Won't do because product agreed that the original way is the expected behaviour.

PR #2125, opened 2023-08-29
[MBL-16965][Student] - Separate announcement list page from discussion list page

kdeakinstructure:
Inherit AnnouncementListPage from DiscussionListPage to make code readable.
Rename editAnnouncementPage to AnnouncementDetailsPage, since it's the detailer page.
Some other minor refactors.

refs: MBL-16965
affects: Student, Teacher
release note: none

### DIFF
--- a/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/pages/AnnouncementDetailsPage.kt
+++ b/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/pages/AnnouncementDetailsPage.kt
@@ -27,9 +27,9 @@ import com.instructure.espresso.scrollTo
 import com.instructure.teacher.R
 
 /**
- * Represents the Edit Announcement page.
+ * Represents the Announcement Details Page.
  */
-class EditAnnouncementPage : BasePage() {
+class AnnouncementDetailsPage : BasePage() {
 
     /**
      * Opens the edit mode for the announcement.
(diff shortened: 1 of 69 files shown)


inst-danger:
<p>Student Install Page</p>

apps/teacher/src/androidTest/java/com/instructure/teacher/ui/pages/AnnouncementDetailsPage.kt
```diff
@@ -27,9 +27,9 @@ import com.instructure.espresso.scrollTo
 import com.instructure.teacher.R
 
 /**
- * Represents the Edit Announcement page.
+ * Represents the Announcement Details Page.
  */
-class EditAnnouncementPage : BasePage() {
+class AnnouncementDetailsPage : BasePage() {
```
tamaskozmer, line 32: This is not necessarily correct. As you can see most of the actions on this page are indeed on the edit announcement screen. The only one which is not is the `openEdit`. It's a bit strange but I can live with that, since it navigates to the edit screen. However the better option would be to leave this as is, remove this one function and use the `DiscussionDetailsPage` (which is the same as an announcement) and use the `openEdit` function from there.

tamaskozmer (changes requested):

apps/teacher/src/androidTest/java/com/instructure/teacher/ui/pages/AnnouncementDetailsPage.kt
kdeakinstructure: We have discussed this via Slack.

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

tamaskozmer (approved):

PR #2143, opened 2023-09-08
[MBL-16877][Student] Keep user logged in while offline

tamaskozmer:
Test plan: 
- Verify that when the user is in offline mode and logged in, we can open the app and the Dashboard will be displayed
- Verify that after an offline login we do a token check the first time the user goes online and if the token is expired we lof out the user.
- This can only be tested by disabling the cache (or waiting 1 hour after login), I have added a patch for this. [no_network_cache.patch](https://github.com/instructure/canvas-android/files/12557932/no_network_cache.patch)

refs: MBL-16877
affects: Student
release note: none

## Checklist

- [x] Tested in light mode


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

kristofnemere (approved):

PR #2158, opened 2023-09-19
[MBL-16327][Teacher] - Use different type of module items in ModulesE2E test in Teacher app

kdeakinstructure:
Use different type of module items in ModulesE2E test in Teacher app.
Add some page object methods.
Extend test with publish/unpublish module and assert how the module items behave, and extend the test with open the module items and assert that their corresponding detailer page is displayed and loaded.

refs: MBL-16327
affects: Teacher
release note: none


inst-danger:
<p>Teacher Install Page</p>

tamaskozmer (approved):

PR #2192, opened 2023-10-16
[MBL-17099][Student] - Modify flank file to apply 'AND' logic for annotations

kdeakinstructure:
Modify flank file to apply 'AND' logic for annotations. (so only those test will be executed which has both the E2E and OfflineTest annotation as well).


inst-danger:
<p>Student Install Page</p>

tamaskozmer (approved):

PR #2213, opened 2023-10-25
[MBL-17138][Parent] Manage Students settings dialogs are overlapping

kristofnemere:
Test plan: in the ticket

refs: MBL-17138
affects: Parent
release note: none


kdeakinstructure (approved):
QA 👍

PR #2242, opened 2023-11-08
[MBL-17175][Parent] Users are seeing error "Uh oh! We're not sure what happened" when opening app or viewing grades

kristofnemere:
Test plan: in the ticket

refs: MBL-17175
affects: Parent
release note: none


inst-danger:
<p>Parent Install Page</p>

tamaskozmer (approved):
QA + 1

PR #2243, opened 2023-11-09
[MBL-17170][Parent] Updated WebView dependency

tamaskozmer:
Test plan: 
- The issue can be reproduced on Android 10 emulator by pressing tab on the login screen while in the first input field. 
- I could also reproduce it by logging in, the app crashed after login.
- Smoke test also the usages of web view screens in dark/light mode, the whole WebView library was updated. The login screen should not be dark (previously it was), that's intentional.

refs: MBL-17170
affects: Parent
release note: Fixed a crash on web content screens on Android 10.

## Checklist

- [x] Tested in dark mode
- [x] Tested in light mode


inst-danger:
<p>Parent Install Page</p>

kdeakinstructure (approved):
QA 👍

inst-danger:
<p>Parent Install Page</p>

hermannakos (approved):

PR #2252, opened 2023-11-15
[MBL-17146][Student] - Workaround to stabilize flaky Inbox E2E tests

kdeakinstructure:
Workaround for flaky inbox swipe tests

refs: MBL-17146
affects: Student
release note: none


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

tamaskozmer (approved):

inst-danger:
<p>Teacher Install Page</p>

automation/espresso/src/main/kotlin/com/instructure/espresso/TestingUtils.kt
```diff
@@ -49,4 +49,21 @@ fun getCurrentDateInCanvasFormat(): String {
     val dayString = expectedDate.dayOfMonth
     val yearString = expectedDate.year
     return "$monthString $dayString, $yearString"
+}
+
+fun retry(
+    times: Int = 3,
+    delay: Long = 1000,
+    block: () -> Unit
```
hermannakos, line 57: We talked about an incremental delay here, that would also be nice to be added.

automation/espresso/src/main/kotlin/com/instructure/espresso/TestingUtils.kt
```diff
@@ -49,4 +49,21 @@ fun getCurrentDateInCanvasFormat(): String {
...
+) {
+    repeat(times - 1) {
+        try {
+            block()
+            return
+        } catch (e: Throwable) {
+            e.printStackTrace()
+        }
+        Thread.sleep(3000)
```
hermannakos, line 66: this delay will always happen, and not just when an exception is thrown. Also this is baked in, so the delay param is not used.

hermannakos (changes requested):

inst-danger:
<p>Student Install Page</p>

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/InboxE2ETest.kt
```diff
@@ -122,12 +124,14 @@ class InboxE2ETest: StudentTest() {
...
-        Log.d(STEP_TAG, "Select the conversations (${seededConversation.subject} and star it." +
-                "Assert that the selected number of conversations on the toolbar is 1 and the conversation is starred.")
-        inboxPage.selectConversations(listOf(seededConversation.subject))
-        inboxPage.assertSelectedConversationNumber("1")
-        inboxPage.clickUnstar()
-        inboxPage.assertConversationNotStarred(seededConversation.subject)
+        retry(times = 10, delay = 3000) {
+            Log.d(STEP_TAG, "Select the conversations (${seededConversation.subject} and star it." +
+                    "Assert that the selected number of conversations on the toolbar is 1 and the conversation is starred.")
+            inboxPage.selectConversations(listOf(seededConversation.subject))
```
hermannakos, line 130: I don't think we need to retry all of this, only the assertion. We make the needed change, and only after that start the retry for the assertion only. This way we could undo previous changes made in earlier attempts.

hermannakos (dismissed):

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/InboxE2ETest.kt
kdeakinstructure: We don't need for sure if it's the API response which is slow, or maybe the automation click/swipe was too fast, that's why I put that there, but I will refactor and we will see on nightly.

automation/espresso/src/main/kotlin/com/instructure/espresso/TestingUtils.kt
kdeakinstructure: This won't always happen, because the return does not let it to happen when it's the 'happy path'. The delay variable thing is valid, thanks.

automation/espresso/src/main/kotlin/com/instructure/espresso/TestingUtils.kt
kdeakinstructure: I have made another version of this function which will have an incremental delay. I'll put both versions to the tests and we can monitor which one is more stabile, because it cannot be tested locally, because the API response is non-deterministical and sometimes it fails locally but sometimes not (there are periods in which it is failing for sure, but mostly it is passing simply).

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

PR #2262, opened 2023-11-27
Nightly - Stabilize SpeedGrader and Todo E2E tests.

kdeakinstructure:
Stabilize teacher SpeedGraderE2ETest and student TodoE2ETest methods.
Add another function parameter for catch branch.


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

automation/espresso/src/main/kotlin/com/instructure/espresso/TestingUtils.kt
```diff
@@ -54,16 +54,18 @@ fun getCurrentDateInCanvasFormat(): String {
 fun retry(
     times: Int = 3,
     delay: Long = 1000,
-    block: () -> Unit
+    block: () -> Unit,
+    catchBlock: (() -> Unit)? = null
```
kristofnemere, line 58: If you keep the block() to be the last parameter, then you dont have to use it as named param where you already used it.

apps/teacher/src/androidTest/java/com/instructure/teacher/ui/e2e/SpeedGraderE2ETest.kt
```diff
@@ -153,7 +158,7 @@ class SpeedGraderE2ETest : TeacherTest() {
         assignmentSubmissionListPage.clickFilterDialogOk()
 
         Log.d(STEP_TAG,"Assert that there is one submission displayed.")
-        assignmentSubmissionListPage.assertHasSubmission(1)
+        retry(times  = 5, delay = 3000, block = { assignmentSubmissionListPage.assertHasSubmission(1) }, catchBlock =  { refresh() })
```
kristofnemere, line 161: I'm not sure why we need the catchBlock, this should work the same if you call the assertion and the refresh in the block I guess.

kristofnemere (approved):

apps/teacher/src/androidTest/java/com/instructure/teacher/ui/e2e/SpeedGraderE2ETest.kt
kdeakinstructure: Because if the test wouldn't fail (because of the API slowness) then why we would refresh the page? I understand that it wouldn't break anything, but if it can be solved that it does not happen when everything is alright and the API is not slow then we don't need to do an unnecessary action.

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

PR #2299, opened 2024-01-10
[MBL-17041][Teacher] - Extend People E2E test with filter by (user) groups

kdeakinstructure:
Extend PeopleE2ETest with filter by (user) groups.
Seed user groups and assign student to them.
Add some page object methods.

refs: MBL-17041
affects: Teacher
release note: none

### DIFF
--- a/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/e2e/PeopleE2ETest.kt
+++ b/apps/teacher/src/androidTest/java/com/instructure/teacher/ui/e2e/PeopleE2ETest.kt
@@ -19,6 +19,7 @@ package com.instructure.teacher.ui.e2e
 import android.util.Log
 import androidx.test.espresso.Espresso
 import com.instructure.canvas.espresso.E2E
+import com.instructure.dataseeding.api.GroupsApi
 import com.instructure.dataseeding.api.SubmissionsApi
 import com.instructure.dataseeding.model.AssignmentApiModel
 import com.instructure.dataseeding.model.CanvasUserApiModel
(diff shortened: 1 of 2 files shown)


inst-danger:
<p>Teacher Install Page</p>

apps/teacher/src/androidTest/java/com/instructure/teacher/ui/e2e/PeopleE2ETest.kt
```diff
@@ -61,6 +63,18 @@ class PeopleE2ETest: TeacherTest() {
         val course = data.coursesList[0]
         val parent = data.parentsList[0]
 
+        Log.d(PREPARATION_TAG,"Seed some group info.")
+        val groupCategory = GroupsApi.createCourseGroupCategory(data.coursesList[0].id, teacher.token)
```
hermannakos, line 67: I would move these to the `StudentTestExtensions.kt` with the other seed functions.

hermannakos (changes requested):

tamaskozmer (approved):

apps/teacher/src/androidTest/java/com/instructure/teacher/ui/e2e/PeopleE2ETest.kt
kdeakinstructure: There will be a major refactor in a separate ticket when the so called 'ApiManager' classes will be introduced soon.

hermannakos: That's awesome, but until then we should still stick with the already used patterns. Otherwise things might get looked over during a major refactor.

hermannakos: I'll drop the CR, but still think we should follow existing patterns.

kdeakinstructure: Thanks. I'll do my best to make that refactor asap.

PR #2304, opened 2024-01-15
[MBL-17312][Student] - Create Offline E2E test for People page

kdeakinstructure:
Implement Offline test for People page.
Stabilize existing ManageOfflineContentE2ETest.
Add some page object methods and test class.

refs: MBL-17312
affects: Student
release note: none

### DIFF
--- a/apps/student/src/androidTest/java/com/instructure/student/ui/e2e/offline/ManageOfflineContentE2ETest.kt
+++ b/apps/student/src/androidTest/java/com/instructure/student/ui/e2e/offline/ManageOfflineContentE2ETest.kt
@@ -18,6 +18,8 @@ package com.instructure.student.ui.e2e.offline
 
 import android.util.Log
 import androidx.test.espresso.Espresso
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.uiautomator.UiDevice
 import com.google.android.material.checkbox.MaterialCheckBox
 import com.instructure.canvas.espresso.FeatureCategory
 import com.instructure.canvas.espresso.OfflineE2E
@@ -47,6 +49,9 @@ class ManageOfflineContentE2ETest : StudentTest() {
         val student = data.studentsList[0]
         val course1 = data.coursesList[0]
         val course2 = data.coursesList[1]
+
+        Log.d(PREPARATION_TAG, "Get the device to be able to perform app-independent actions on it.")
+        val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
 
         Log.d(STEP_TAG,"Login with user: ${student.name}, login id: ${student.loginId}.")
         tokenLogin(student)
@@ -203,6 +208,11 @@ class ManageOfflineContentE2ETest : StudentTest() {
 
         Log.d(PREPARATION_TAG, "Turn off the Wi-Fi and Mobile Data on the device, so it will go offline.")
         turnOffConnectionViaADB()
+        Thread.sleep(10000) //Need to wait a bit here because of a UI glitch that when network state change, the dashboard page 'pops' a bit and it can confuse the automation script.
+        device.waitForIdle()
+        device.waitForWindowUpdate(null, 10000)
+
+        Log.d(STEP_TAG, "Wait for the Dashboard Page to be rendered. Refresh the page.")
         dashboardPage.waitForRender()
 
         Log.d(STEP_TAG, "Select '${course2.name}' course and open 'Grades' menu to check if it's really synced and can be seen in offline mode.")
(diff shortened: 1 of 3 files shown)


inst-danger:
<p>Student Install Page</p>

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/offline/ManageOfflineContentE2ETest.kt
```diff
@@ -203,6 +208,11 @@ class ManageOfflineContentE2ETest : StudentTest() {
 
         Log.d(PREPARATION_TAG, "Turn off the Wi-Fi and Mobile Data on the device, so it will go offline.")
         turnOffConnectionViaADB()
+        Thread.sleep(10000) //Need to wait a bit here because of a UI glitch that when network state change, the dashboard page 'pops' a bit and it can confuse the automation script.
```
hermannakos, line 211: Do we need all these sleeps and waits together? Are 10 secs really necessary? Could we make it so we only wait until the desired state is reached?

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/offline/OfflinePeopleE2ETest.kt
```diff
@@ -0,0 +1,109 @@
...
+        dashboardPage.waitForSyncProgressDownloadStartedNotification()
+        dashboardPage.waitForSyncProgressDownloadStartedNotificationToDisappear()
+
+        Log.d(STEP_TAG, "Wait for the 'Syncing Offline Content' dashboard notification to be displayed, and the to disappear. (It should be displayed after the 'Download Started' notification immediately.)")
+        dashboardPage.waitForSyncProgressStartingNotification()
+        dashboardPage.waitForSyncProgressStartingNotificationToDisappear()
+
+        Log.d(PREPARATION_TAG, "Turn off the Wi-Fi and Mobile Data on the device, so it will go offline.")
+        turnOffConnectionViaADB()
+        Thread.sleep(10000) //Need to wait a bit here because of a UI glitch that when network state change, the dashboard page 'pops' a bit and it can confuse the automation script.
```
hermannakos, line 79: Do we need all these sleeps and waits together? Are 10 secs really necessary? Could we make it so we only wait until the desired state is reached?

hermannakos (changes requested):

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/offline/ManageOfflineContentE2ETest.kt
kdeakinstructure: I have tried to figure it out what is actually happening on the emulator after I turn off the connection, but it still does not clear for me. I have tried many solutions and this is the only way when it is not flaky. If you up to help me figure it out what could be a better solution here, I'm up to it, but currently this is best I could make and this is works for sure, because there are some other offline tests in which this solution has really stabilized the test. As soon as we don't have many offline test, I think we can cope with this, but for the long term, there could be a better (optimal) solution of course.

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/offline/OfflinePeopleE2ETest.kt
kdeakinstructure: See above.

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/offline/OfflinePeopleE2ETest.kt
```diff
@@ -0,0 +1,109 @@
...
+        dashboardPage.waitForRender()
+        dashboardPage.waitForSyncProgressDownloadStartedNotification()
+        dashboardPage.waitForSyncProgressDownloadStartedNotificationToDisappear()
+
+        Log.d(STEP_TAG, "Wait for the 'Syncing Offline Content' dashboard notification to be displayed, and the to disappear. (It should be displayed after the 'Download Started' notification immediately.)")
+        dashboardPage.waitForSyncProgressStartingNotification()
+        dashboardPage.waitForSyncProgressStartingNotificationToDisappear()
+
+        Log.d(PREPARATION_TAG, "Turn off the Wi-Fi and Mobile Data on the device, so it will go offline.")
+        turnOffConnectionViaADB()
```
tamaskozmer, line 78: You could extract this sequence of actions into utility function that can be reused in every offline test. It will be cleaner that way, you avoid a lot of duplications and we can easily change it later when we come up with a better solution for the delays.

tamaskozmer (approved):

apps/student/src/androidTest/java/com/instructure/student/ui/e2e/offline/OfflinePeopleE2ETest.kt
kdeakinstructure: Good idea. I have made a separate ticket for this extracting, since there are other tests in which these are used.
I'll get this done the soonest, thanks!

PR #2336, opened 2024-02-11
[MBL-16673][Student] -  Introduce ApiManager class and extract direct API calls from E2E test files

kdeakinstructure:
Introduce StudentApiManager class.
Extract all the direct API calls from the E2E test classes into StudentApiManager.
Remove lots of duplications.

refs: MBL-16673
affects: Student
release note: none


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

kdeakinstructure:
E2E test run: https://app.bitrise.io/build/f1c36967-6508-4959-90d9-90b4f692378d
(Successful)

Offline E2E test run: https://app.bitrise.io/build/6f0da429-0ddd-4f5e-9ee8-a01efae4dd0f
(Successful)

apps/student/src/androidTest/java/com/instructure/student/ui/utils/StudentApiManager.kt
```diff
@@ -0,0 +1,397 @@
...
+import com.instructure.dataseeding.model.*
+import com.instructure.dataseeding.util.CanvasNetworkAdapter
+import com.instructure.dataseeding.util.Randomizer
+import com.instructure.dataseeding.util.days
+import com.instructure.dataseeding.util.fromNow
+import com.instructure.dataseeding.util.iso8601
+import java.io.File
+import java.util.*
+
+class StudentApiManager {
```
tamaskozmer, line 32: This could be an Object instead of a Class, since everything here is in the companion object.

tamaskozmer, line 32: I wouldn't call this `StudentApiManager`. I guess the most of the API calls could be used in the teacher tests, since these calls are for creating data for the test users. Ideally the role wouldn't matter here and this object would be in the `dataseeding` module.

tamaskozmer (changes requested):

apps/student/src/androidTest/java/com/instructure/student/ui/utils/StudentApiManager.kt
kdeakinstructure: Refactored

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

tamaskozmer (approved):

automation/dataseedingapi/src/main/kotlin/com/instructure/dataseeding/util/ApiManager.kt
```diff
@@ -0,0 +1,389 @@
...
+@file:Suppress("unused")
+
+package com.instructure.dataseeding.util
+
+import com.instructure.dataseeding.api.*
+import com.instructure.dataseeding.model.*
+import java.io.File
+import java.util.*
+
+object ApiManager {
```
hermannakos, line 26: I feel like this object could grow bigger and bigger with no end. It would possible be beneficial to split it into different classes based on the api called. i.e `UserApiManager`, `CourseApiManager` etc. I would do this now rather than later, because it will always be overlooked.

automation/dataseedingapi/src/main/kotlin/com/instructure/dataseeding/util/Extensions.kt
```diff
@@ -48,4 +48,14 @@ val Calendar.iso8601: String
     get() {
         val formatted = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.US).format(Date(this.timeInMillis))
         return formatted.substring(0, 22) + ":" + formatted.substring(22)
-    }
\ No newline at end of file
+    }
+
+@JvmOverloads
+fun Date?.toApiString(timeZone: TimeZone? = null): String {
```
hermannakos, line 54: this is now duplicated. Since the calendar is not used other than fetching the date we could pass the date string instead.

hermannakos (changes requested):

automation/dataseedingapi/src/main/kotlin/com/instructure/dataseeding/util/Extensions.kt
kdeakinstructure: I am using the calendars in some if conditions below.

hermannakos: It is only used once, in `ApiManager.kt::107`

automation/dataseedingapi/src/main/kotlin/com/instructure/dataseeding/util/ApiManager.kt
kdeakinstructure: Actually the goal was to extract all the api calls into one file/class, to reduce code duplication significantly. So this was intentional to be in one class. I don't really think this will endlessly grow in the future since the most API calls we are using in the tests are mostly already in it.  The current size of it is less than 400 lines, I don't feel this as big, there are some much more bigger classes already in our code base (e.g. MockCanvas with 2000+ lines).

hermannakos: On second thought it might defeat the purpose of this `ApiManager`. I would look for a unified way of handling these seeding calls. If you have some logic in the `ApiManager` you don't want to handle in the tests then it might be beneficial to separate it into different managers. But if the `ApiManager` would only be a middle man and just pass calls to the api, it would make sense to call the api directly.

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

kristofnemere (approved):

hermannakos (approved):

PR #2535, opened 2024-08-23
Inbox compose attachment handling

domonkosadam:
Implemented on Compose Screen:

Context picker
- Recipient picker
- Individual send
- Subject field
- Body field
- **Attachment handling**

To be implemented:
- Reply, Forward handling
- Update list screen if a new message composed
- Search recipient on Compose Screen


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/InboxComposeViewModel.kt
```diff
@@ -308,4 +346,28 @@ class InboxComposeViewModel @Inject constructor(
...
+            WorkInfo.State.SUCCEEDED -> AttachmentStatus.UPLOADED
+            WorkInfo.State.FAILED -> AttachmentStatus.FAILED
+            WorkInfo.State.ENQUEUED -> AttachmentStatus.UPLOADING
+            WorkInfo.State.RUNNING -> AttachmentStatus.UPLOADING
+            WorkInfo.State.BLOCKED -> AttachmentStatus.FAILED
+            WorkInfo.State.CANCELLED -> AttachmentStatus.FAILED
+        }
+    }
+
+    private fun downloadFileToDevice(attachment: Attachment) {
```
tamaskozmer, line 361: I would extract this into it's own class. It might be reused in other places as well (for example downloading attachments from the inbox details) and it would also make testing easier for the ViewModel.

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/InboxComposeViewModel.kt
```diff
@@ -40,6 +49,22 @@ class InboxComposeViewModel @Inject constructor(
         loadContexts()
     }
 
+    fun updateAttachments(uuid: UUID?, workInfo: WorkInfo) {
```
tamaskozmer, line 52: You should add tests for the new features in the ViewModel. Also for the new Actions as well in the `handleAction` function.

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/composables/AttachmentCard.kt
```diff
@@ -47,21 +51,28 @@ import com.bumptech.glide.integration.compose.GlideImage
...
 @Composable
 fun AttachmentCard(
-    attachment: Attachment,
-    status: AttachmentStatus,
-    onRemove: () -> Unit,
-    context: Context
+    attachmentCardItem: AttachmentCardItem,
+    context: Context,
+    onSelect: () -> Unit,
+    onRemove: () -> Unit
```
tamaskozmer, line 64: I think this should have an optional modifier as a parameter.

tamaskozmer (changes requested):

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/InboxComposeViewModel.kt
```diff
@@ -40,6 +49,22 @@ class InboxComposeViewModel @Inject constructor(
...
+        if (workInfo.state == WorkInfo.State.SUCCEEDED) {
+            viewModelScope.launch {
+                uuid?.let { uuid ->
+                    val attachmentEntities = attachmentDao.findByParentId(uuid.toString())
+                    val status = workInfo.state.toAttachmentCardStatus()
+                    attachmentEntities?.let { attachmentList ->
+                        _uiState.update { it.copy(attachments = it.attachments + attachmentList.map { AttachmentCardItem(it.toApiModel(), status) }) }
+                        attachmentDao.deleteAll(attachmentList)
+                    } ?: context.toast(R.string.errorUploadingFile)
```
kristofnemere, line 61: this kind of toast is weird for me in the viewmodel, I would have an event for these in InboxComposeViewModelAction and handle in the fragment

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/composables/AttachmentCard.kt
```diff
@@ -47,21 +51,28 @@ import com.bumptech.glide.integration.compose.GlideImage
...
 
 @OptIn(ExperimentalGlideComposeApi::class)
 @Composable
 fun AttachmentCard(
-    attachment: Attachment,
-    status: AttachmentStatus,
-    onRemove: () -> Unit,
-    context: Context
+    attachmentCardItem: AttachmentCardItem,
+    context: Context,
```
kristofnemere, line 62: I think you don't need context as a paramter, you could get it inside like LocalContext.current

libs/pandautils/src/test/java/com/instructure/pandautils/features/inbox/compose/InboxComposeViewModelTest.kt
```diff
@@ -384,6 +386,6 @@ class InboxComposeViewModelTest {
     //endregion
 
     private fun getViewModel(): InboxComposeViewModel {
-        return InboxComposeViewModel(context, inboxComposeRepository)
+        return InboxComposeViewModel(context, inboxComposeRepository, attachmentDao)
```
kristofnemere, line 389: You should test the new logics

kristofnemere (changes requested):

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/composables/AttachmentCard.kt
domonkosadam: It will be fixed with implementing the Conversation Details screen.

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/composables/AttachmentCard.kt
domonkosadam: It will be fixed with implementing the Conversation Details screen.

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/InboxComposeViewModel.kt
```diff
@@ -308,4 +346,28 @@ class InboxComposeViewModel @Inject constructor(
             else -> ""
         }
     }
+
+    private fun WorkInfo.State.toAttachmentCardStatus(): AttachmentStatus {
```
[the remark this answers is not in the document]
domonkosadam: It will be fixed with implementing the Conversation Details screen.

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

libs/pandautils/src/main/java/com/instructure/pandautils/utils/FileDownloader.kt
```diff
@@ -0,0 +1,22 @@
+package com.instructure.pandautils.utils
+
+import android.app.DownloadManager
+import android.content.Context
+import android.net.Uri
+import android.os.Environment
+import com.instructure.canvasapi2.models.Attachment
+
+class FileDownloader(private val context: Context) {
```
kristofnemere, line 9: I would just provide DownloadManager instead of this class, or are we gonna use this elsewhere? If you want to keep this, please give it a more specific name that describes better what it does - like AttachmentFileDownloader or something like that

kristofnemere (changes requested):

libs/pandautils/src/main/java/com/instructure/pandautils/utils/FileDownloader.kt
tamaskozmer: I agree with the first part. I think it doesn't need a more specific name, I would keep this more generic like this. However in that case I would pass the `fileName`, `url` and `contentType` instead of an `Attachment` object so it can be reused easily.

tamaskozmer (approved):

libs/pandautils/src/main/java/com/instructure/pandautils/utils/FileDownloader.kt
domonkosadam: We might use this class on the Conversation Details screen as well. I think it's easier to handle downloads in ViewModels by injecting the downloader class itself.

Wouldn't be a better solution to have multiple functions with different parameters?
```Kotlin
class FileDownloader(private val context: Context) {
    fun downloadFileToDevice(attachment: Attachment) {
        downloadFileToDevice(attachment.url, attachment.filename, attachment.contentType)
    }

    fun downloadFileToDevice(
        downloadURL: String?,
        filename: String?,
        contentType: String?
    ) {
        downloadFileToDevice(Uri.parse(downloadURL), filename, contentType)
    }

    fun downloadFileToDevice(
        downloadURI: Uri,
        filename: String?,
        contentType: String?
    ) {
        val downloadManager = context.getSystemService(DownloadManager::class.java)

        val request = DownloadManager.Request(downloadURI)
        request
            .setNotificationVisibility(DownloadManager.Request.VISIBILITY_VISIBLE_NOTIFY_COMPLETED)
            .setTitle(filename)
            .setMimeType(contentType)
            .setDestinationInExternalPublicDir(Environment.DIRECTORY_DOWNLOADS, "$filename")

        downloadManager.enqueue(request)
    }
}
```

tamaskozmer: This can be good as well, but I would keep only the first one, and we can add the second one when we need to call it with an URI.

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

libs/pandautils/src/main/java/com/instructure/pandautils/utils/FileDownloader.kt
```diff
@@ -0,0 +1,38 @@
...
+    fun downloadFileToDevice(attachment: Attachment) {
+        downloadFileToDevice(attachment.url, attachment.filename, attachment.contentType)
+    }
+
+    fun downloadFileToDevice(
```
kristofnemere, line 14: I think this method is not necessary, the uri parse could be in the below method, we gonna call it with a string url typically

kristofnemere (approved):
QA+1

PR #2546, opened 2024-08-30
[MBL-17828][Teacher] Cannot open XML files

domonkosadam:
# Cannot open XML files

This bug was related to the Samsung HTML Viewer application.
When files are retrieved from a FileCache, their names and extensions are replaced with IDs. Since some external applications may rely on this information, we can rename the retrieved files to their original names and extensions so they will be accessible to external applications as well.

Test plan:
- Upload an XML file to user Files
- Open it from the Teacher app with the default Samsung HTML Viewer

refs: MBL-17828
affects: Teacher
release note: XML files can be opened from user Files

## Screenshots

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist

- [ ] Tested in light mode


inst-danger:
<p>Teacher Install Page</p>

tamaskozmer (approved):

PR #2568, opened 2024-09-19
Remove Multidex usages

MGaetan89:
Since the min SDK is at least 21, it is no longer necessary to use the Multidex library.

See the following for more info: https://developer.android.com/build/multidex#mdex-on-l


MGaetan89:
I see the builds fail on my PR. Is there something I have to fix, or is it the same failures as on `master`?

tamaskozmer:
Hi @MGaetan89. I have reviewed your PR, and you are right, we don't need multidex anymore, it's a legacy part that remained in our codebase for a long time. Currently we have some issues with the open source builds so they can't run on your PR. I ran our E2E tests for both our apps for the proposed changes on a different branch and they were successful, so I will merge your PR. 

Thank you for your contribution.

PR #2590, opened 2024-10-08
[MBL-17974][Student] - Handle 0 value in assignment reminder custom dialog

kdeakinstructure:
refs: MBL-17974
affects: Student
release note: Handle 0 value in assignment reminder custom dialog

## Checklist

- [x] Tested in dark mode
- [x] Tested in light mode

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/features/assignments/reminder/CustomReminderDialog.kt
+++ b/apps/student/src/main/java/com/instructure/student/features/assignments/reminder/CustomReminderDialog.kt
@@ -77,7 +77,7 @@ class CustomReminderDialog : DialogFragment() {
     }
 
     private fun updateButtonState(button: Button) {
-        button.isEnabled = binding.choices.checkedRadioButtonId != -1 && binding.quantity.text.isNotEmpty()
+        button.isEnabled = binding.choices.checkedRadioButtonId != -1 && binding.quantity.text.isNotEmpty() && binding.quantity.text.toString() != "0"
     }
 
     companion object {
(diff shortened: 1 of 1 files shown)


inst-danger:
<p>Student Install Page</p>

apps/student/src/main/java/com/instructure/student/features/assignments/reminder/CustomReminderDialog.kt
```diff
@@ -77,7 +77,7 @@ class CustomReminderDialog : DialogFragment() {
     }
 
     private fun updateButtonState(button: Button) {
-        button.isEnabled = binding.choices.checkedRadioButtonId != -1 && binding.quantity.text.isNotEmpty()
+        button.isEnabled = binding.choices.checkedRadioButtonId != -1 && binding.quantity.text.isNotEmpty() && binding.quantity.text.toString() != "0"
```
tamaskozmer, line 80: While this solves the issue in the ticket you can still type for example "0000" and it will be accepted. You should check the numerical value safely with `binding.quantity.text.toString().toIntOrNull().orDefault() > 0` I think in this case we won't even need the text.isNotEmpty condition.

tamaskozmer (changes requested):

apps/student/src/main/java/com/instructure/student/features/assignments/reminder/CustomReminderDialog.kt
kdeakinstructure: Good advice, thanks!

inst-danger:
<p>Student Install Page</p>

tamaskozmer (approved):

tamaskozmer:
QA + 1

PR #2608, opened 2024-10-21
[MBL-17987][Student] Clicking alerts regarding a Discussion mention gives an error

tamaskozmer:
Test plan: In the ticket

refs: MBL-17987
affects: Student
release note: Fixed a bug where discussion mention alerts couldn't be opened.

## Checklist

- [x] Tested in dark mode
- [x] Tested in light mode


inst-danger:

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

domonkosadam (changes requested):
I've tested it with the attached account and the same error occurred with some of the mention notifications.

tamaskozmer:
> I've tested it with the attached account and the same error occurred with some of the mention notifications.

Yes, those are expected, that doesn't work on iOS either because there is no way to route to the discussion. We will contact an other team with that. If the item in the video works than it's okay.

PR #2610, opened 2024-10-22
[MBL-18007][Teacher] Calendar events assigned to multiple sections show duplicates

tamaskozmer:
Test plan: In the ticket. This is also happening in the new Parent app, check my comment on the ticket.

refs: MBL-18007
affects: Teacher
release note: none

## Checklist

- [ ] Tested in dark mode
- [ ] Tested in light mode

### DIFF
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/features/calendar/CalendarRepository.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/features/calendar/CalendarRepository.kt
@@ -16,24 +16,66 @@
 package com.instructure.pandautils.features.calendar
 
 import com.instructure.canvasapi2.models.CanvasContext
+import com.instructure.canvasapi2.models.Plannable
+import com.instructure.canvasapi2.models.PlannableType
 import com.instructure.canvasapi2.models.PlannerItem
 import com.instructure.canvasapi2.utils.DataResult
+import com.instructure.canvasapi2.utils.toDate
 import com.instructure.pandautils.room.calendar.entities.CalendarFilterEntity
 
-interface CalendarRepository {
+abstract class CalendarRepository {
 
-    suspend fun getPlannerItems(
+    protected var canvasContexts: List<CanvasContext> = emptyList()
+
+    abstract suspend fun getPlannerItems(
         startDate: String,
         endDate: String,
         contextCodes: List<String>,
         forceNetwork: Boolean
     ): List<PlannerItem>
 
-    suspend fun getCanvasContexts(): DataResult<Map<CanvasContext.Type, List<CanvasContext>>>
+    abstract suspend fun getCanvasContexts(): DataResult<Map<CanvasContext.Type, List<CanvasContext>>>
 
-    suspend fun getCalendarFilterLimit(): Int
+    abstract suspend fun getCalendarFilterLimit(): Int
 
-    suspend fun getCalendarFilters(): CalendarFilterEntity?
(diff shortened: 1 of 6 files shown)


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:

libs/pandautils/src/main/java/com/instructure/pandautils/features/calendar/CalendarRepository.kt
```diff
@@ -16,24 +16,66 @@
...
+                    submissionState = null,
+                    newActivity = false,
+                    plannerOverride = null
+                )
+            }
+        }
+    }
 
-    suspend fun updateCalendarFilters(calendarFilterEntity: CalendarFilterEntity)
+    protected fun List<PlannerItem>.mapContextName(): List<PlannerItem> {
```
domonkosadam, line 71: If I understood correctly the logic of this function, calling this extension function requires to call getCanvasContexts() function before. Wouldn't be a better idea to display this constraint in the code as well? Maybe passing canvasContexts as parameter would be a better approach to avoid future bugs.

libs/pandautils/src/main/java/com/instructure/pandautils/features/calendar/CalendarRepository.kt
```diff
@@ -16,24 +16,66 @@
...
+
+    abstract suspend fun getCalendarFilters(): CalendarFilterEntity?
 
-    suspend fun getCalendarFilterLimit(): Int
+    abstract suspend fun updateCalendarFilters(calendarFilterEntity: CalendarFilterEntity)
 
-    suspend fun getCalendarFilters(): CalendarFilterEntity?
+    protected fun List<Plannable>.toPlannerItems(): List<PlannerItem> {
+        return mapNotNull { plannable ->
+            val contextType = if (plannable.courseId != null) CanvasContext.Type.COURSE.apiString else CanvasContext.Type.USER.apiString
```
domonkosadam, line 47: Plannable has a groupId attribute as well. What's the reason for not handling that type?

domonkosadam:
QA + 1

libs/pandautils/src/main/java/com/instructure/pandautils/features/calendar/CalendarRepository.kt
tamaskozmer: There are no groups in Teacher and Parent and these are only used by those two apps.

libs/pandautils/src/main/java/com/instructure/pandautils/features/calendar/CalendarRepository.kt
tamaskozmer: It is a bit hacky, because the backend doesn't return the context for some items. We already used the `canvasContexts` before like this to find the correct course for To Dos in this function `List<Plannable>.toPlannerItems()`. Generally that is always called before getting the calendar events. I can add a comment to clarify that as well.

domonkosadam (approved):

inst-danger:

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

PR #2730, opened 2025-01-30
[MBL-18207][Teacher] a11y speedgrader bottom tab selection focus

andrasmaczak:
Test plan: See ticket

refs: MBL-18207
affects: Teacher
release note: none


inst-danger:

inst-danger:
<p>Teacher Install Page</p>

apps/teacher/src/main/java/com/instructure/teacher/view/SubmissionContentView.kt
```diff
@@ -820,7 +822,9 @@ class SubmissionContentView(
         mBottomViewPager.addOnPageChangeListener(object : ViewPager.OnPageChangeListener {
             override fun onPageScrolled(position: Int, positionOffset: Float, positionOffsetPixels: Int) {}
 
-            override fun onPageSelected(position: Int) {}
+            override fun onPageSelected(position: Int) {
```
domonkosadam, line 825: The TalkBack reader does not jump to the Files content unlike in the first 2 case.

domonkosadam, line 825: I'm not even sure if the original logic should be changed, @hermannakos? I think it worths to discuss it with Product.

My opinion is that this solution is way more confusing than the original one, as it works unpredictable in 2 of 3 cases. It works well with "Grades", but when I select "Comments" it reads the entire visible comment part then the next items will be the same comments one by one. When I select "Files" it will jump to the empty content when there is no uploaded file to the submission, but if there are files attached to the submission TalkBack will remain on the tab.

I checked Google Drive's tab layout as an example and it does not jumps to the content.

andrasmaczak: I would vote for reverting this, there is no clean solution and this one is also unreliable.

PR #2758, opened 2025-02-04
Fix offline file handling

domonkosadam:
Fix finding: Embedded pdf file can only be opened once in offline mode + Also, embedded pictures on pages does not load.


inst-danger:
<p>Student Install Page</p>

tamaskozmer:
QA + 1

PR #2766, opened 2025-02-04
[MBL-18279][All] Settings screen a11y improvements

tamaskozmer:
Test plan: The ticket is about the old settings screen, so that is not relevant, but I found an issue in the new settings screen as well, when the talkback announces the theme change buttons separately from the texts under them so removed the announcements from the labels (those are not relevant, because they are not clickable and announce the same as the icon)

refs: MBL-18279
affects: All
release note: none

## Checklist

- [ ] Tested in light mode
- [ ] A11y checked


inst-danger:

inst-danger:
<p>Teacher Install Page</p>

kristofnemere (approved):

PR #2772, opened 2025-02-05
Fix inbox recipients update

domonkosadam:
RC finding: When selecting a recipient on the inbox compose message screen, and then change the course to a course in which that recipient is not in, the recipient name will remain in the recipient list, However the 'send' button will remain disabled (which prevents the user from sending messages to a particular course with recipients who are not enrolled to that course and it's good), but on the UI it could be misleading.


inst-danger:

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

tamaskozmer (approved):

PR #2785, opened 2025-02-07
[MBL-16700][Student] Inbox message can display on incorrect day on Notifications page when deleted by different user

tamaskozmer:
Test plan: It's on old ticket so the test user in the ticket won't work. Repro steps are written in the ticket. 
- If you have a student that has recently received messages, you can test with them. Delete a message with an other user present in that conversation and notice that the conversations notification gets updated in the Notifications screen. After checking this you can check the new build and see if the day is correct and not updated. If you don't have such user you can test with my test student tomorrow (I created the message today, so you need to wait for tomorrow and delete it with other user, so the changes take effect)

refs: MBL-16700
affects: Student
release note: none

## Checklist

- [ ] Tested in light mode


inst-danger:

hermannakos (approved):

hermannakos (approved):
QA + 1

PR #2807, opened 2025-02-17
[MBL-18603][Parent] Tapping Alert Settings toggles too fast may can result 'unexpected error' bottom message

kristofnemere:
Test plan: in the ticket

refs: MBL-18603
affects: Parent
release note: none


inst-danger:

PR #2810, opened 2025-02-17
[MBL-18604] 'Save' button disabled state on alert settings dialog is not accessible in dark mode

kristofnemere:
Test plan: in the ticket

refs: MBL-18604
affects: Parent
release note: none


inst-danger:

PR #2816, opened 2025-02-18
[MBL-18606][Parent] We don't have any message if the user tries to upload but exceeded it's storage quota

kristofnemere:
Test plan: in the ticket

refs: MBL-18606
affects: Parent
release note: none


tamaskozmer (approved):

kdeakinstructure (approved):
QA 👍

PR #2864, opened 2025-03-03
[MBL-16366][Teacher] Migrate FileDownloadService 

tamaskozmer:
Test plan: This service was only used when we were deeplinking to a canvas file from the web. And the only think it did is to download the file and clicking the notification did nothing. I guess this was not the intended behavior, but it was broken a long time ago, so I just replaced it with our FileDownloadWorker. Testing can be with done the following steps.
- Enable deep linking to instructure domains in the app settings.
- Download the attached deep link html file to the ticket (or you can create your own. This will only work with my test user)
- Verify that the file is downloaded with the correct name.

refs: MBL-16366
affects: Teacher
release note: none

## Checklist

- [x] Tested in light mode


inst-danger:

inst-danger:
<p>Teacher Install Page</p>

hermannakos (approved):
QA + 1

PR #2873, opened 2025-03-04
[MBL-18047][Teacher] Added option to navigate to module headers from links.

tamaskozmer:
I had to do this a bit hacky way, because of this legacy RecyclerView used here. There is no way that we can get the positions of the headers, so we just subtract one from the first item's position in it's group.

Test plan: In the ticket

refs: MBL-18047
affects: Teacher
release note: none

## Checklist

- [x] Tested in light mode


inst-danger:
<p>Teacher Install Page</p>

kristofnemere (approved):

PR #2883, opened 2025-03-10
[MBL-18669][Student][Teacher] Terms of Use page 'Webpage not available'

tamaskozmer:
Test plan: In the ticket. Actually it's quite hard to reproduce this in the Student because once you login the terms of use is requested and it will be cached, so it will open correctly. In the teacher we only request it on the LegalDialog, so you can test there. Just go offline after a new login and before opening the legal dialog.

refs: MBL-18669
affects: Student, Teacher
release note: none

## Checklist

- [x] Tested in light mode


inst-danger:

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

kdeakinstructure (approved):
QA 👍

kdeakinstructure:
Btw it was quite easy to reproduce this in the Student as well because we are supporting "Change User" functionality in offline, so if you tap on "Change User" then go offline on the login page, select a user (could be the same you were logged in recently) and then navigate to Legal -- Terms of Use, it will show the No Internet Connection dialog.

kristofnemere (approved):

PR #2927, opened 2025-03-28
[CLX-1098][Horizon] Segmented control

tamaskozmer:


libs/horizon/src/main/java/com/instructure/horizon/horizonui/molecules/SegmentedControl.kt
```diff
@@ -0,0 +1,281 @@
...
+    data class NoIcon(override val checkmark: Boolean = false) : SegmentedControlIconPosition(checkmark)
+    data class Start(override val checkmark: Boolean = false, @DrawableRes override val iconRes: Int = R.drawable.check) :
+        SegmentedControlIconPosition(checkmark, iconRes)
+
+    data class End(override val checkmark: Boolean = false, @DrawableRes override val iconRes: Int = R.drawable.check) :
+        SegmentedControlIconPosition(checkmark, iconRes)
+}
+
+@Composable
+fun SegmentedControl(
```
domonkosadam, line 73: I think current state should be stored outside of the component. There are two reasons for that:

1. It might be needed to set selected value programatically.
2. If we embed the component to a lazy view the selected value will be lost as you can see in the screen recording

https://github.com/user-attachments/assets/f9952a6d-a841-4373-932a-df83f0b75a63

domonkosadam (changes requested):

domonkosadam (approved):

PR #2929, opened 2025-03-28
[CLX-1100][Horizon] Avatar

tamaskozmer:


libs/horizon/src/main/java/com/instructure/horizon/horizonui/molecules/Avatar.kt
```diff
@@ -0,0 +1,386 @@
...
+                )
+            }
+
+            is AvatarType.Picture -> {
+                GlideImage(
+                    model = type.pictureUrl,
+                    contentDescription = null,
+                    modifier = Modifier
+                        .size(size.size)
+                        .clip(RoundedCornerShape(500.dp)),
```
domonkosadam, line 113: I think we should use the CircleShape instead

libs/horizon/src/main/java/com/instructure/horizon/horizonui/molecules/Avatar.kt
```diff
@@ -0,0 +1,386 @@
...
+            is AvatarType.Initials -> {
+                Text(
+                    text = type.initials.uppercase(),
+                    color = color.contentColor,
+                    style = size.initialsTextStyle,
+                )
+            }
+
+            is AvatarType.Picture -> {
+                GlideImage(
```
domonkosadam, line 108: We could add "Avatar" content description just like in the Icon case.

libs/horizon/src/main/java/com/instructure/horizon/horizonui/molecules/Avatar.kt
```diff
@@ -0,0 +1,386 @@
...
+    size: AvatarSize = AvatarSize.LARGE,
+    color: AvatarColor = AvatarColor.NORMAL
+) {
+    var avatarModifier = modifier
+        .background(
+            color = color.backgroundColor,
+            shape = RoundedCornerShape(500.dp)
+        )
+    if (type !is AvatarType.Picture) avatarModifier =
+        avatarModifier.border(HorizonBorder.level2(color.borderColor), shape = RoundedCornerShape(500.dp))
```
domonkosadam, line 83: I think we should use the CircleShape instead

PR #2934, opened 2025-04-02
[CLX-1105][Horizon] Cards

tamaskozmer:


domonkosadam:
In `FileDrop.kt` file there are some invalid `ButtonColor` attributes so the project cannot be built.

tamaskozmer:
> In FileDrop.kt file there are some invalid ButtonColor attributes so the project cannot be built.
Yes, I found this out and fixed it on my other branch, but will fix here too so it can be built and tested.

inst-danger:
<p>Student Install Page</p>

domonkosadam (approved):

PR #2936, opened 2025-04-02
[MBL-18746][All] Calendar crashes fixed

andrasmaczak:
Test plan:
We could not reproduce this crash, so there is no way to test it  by us.

refs: MBL-18746
release note: None


inst-danger:

inst-danger:
<p>Teacher Install Page</p>

kristofnemere (approved):

PR #2937, opened 2025-04-02
[MBL-18753][Parent] Grading Period filter complaints

kristofnemere:
Test plan: grades by default should be filtered by the current grading period (if theres no current then All Periods), should be grouped by Due Date by default, but should be saved when changed

refs: MBL-18753
affects: Parent
release note: In the Grades Filter, the default grading period is now set to the current one, and the sorting preference is saved.

## Checklist

- [x] Approve from product

### DIFF
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/features/grades/GradesRepository.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/features/grades/GradesRepository.kt
@@ -22,6 +22,7 @@ import com.instructure.canvasapi2.models.Course
 import com.instructure.canvasapi2.models.CourseGrade
 import com.instructure.canvasapi2.models.Enrollment
 import com.instructure.canvasapi2.models.GradingPeriod
+import com.instructure.pandautils.features.grades.gradepreferences.SortBy
 
 
 interface GradesRepository {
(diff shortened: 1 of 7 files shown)


inst-danger:

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

libs/pandautils/src/main/java/com/instructure/pandautils/features/grades/GradesRepository.kt
```diff
@@ -32,4 +33,6 @@ interface GradesRepository {
     suspend fun loadEnrollments(courseId: Long, gradingPeriodId: Long?, forceRefresh: Boolean): List<Enrollment>
     suspend fun loadCourse(courseId: Long, forceRefresh: Boolean): Course
     fun getCourseGrade(course: Course, studentId: Long, enrollments: List<Enrollment>, gradingPeriodId: Long?): CourseGrade?
+    fun getSortBy(): SortBy?
```
hermannakos, line 36: I don't think this belongs in the repository.

hermannakos (changes requested):

libs/pandautils/src/main/java/com/instructure/pandautils/features/grades/GradesRepository.kt
kristofnemere: The viewmodel is in the common package, so this way I can define the sorting logic by app separately, also I think setting/getting preferences can belong to the repository.

hermannakos (approved):

PR #2974, opened 2025-04-24
Create Compose WebView for url loading.

tamaskozmer:
This is just an implementation of the CanvasWebView that can be used in Jetpack Compose. Don't need to test anything, since this is currently not used anywhere, only in the Horizon UI, but should be also merged to master.

Note: We could add the same callbacks for the ComposeCanvasWebViewWrapper as well, it can be beneficial to simplify things, but I didn't want to do that now, because it would need some existing production code to change.


inst-danger:

inst-danger:
<p>Teacher Install Page</p>

PR #2975, opened 2025-04-24
[CLX-1190][Horizon] External link & external tool learning object.

tamaskozmer:


inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

domonkosadam:
During QA I found two interesting behavior, both of them are visible on the following recording:

- When I opened an external content, I had to click on a button to open it in a new tab. I think it's quite confusing as a user.
- After opening and closing the external tab, I quickly navigated forward two modules and then back, the external tab automatically opened.


https://github.com/user-attachments/assets/dd16bff4-e975-4cdf-9d17-bf030b08cfcd

libs/horizon/src/main/java/com/instructure/horizon/features/moduleitemsequence/content/lti/ExternalToolViewModel.kt
```diff
@@ -0,0 +1,87 @@
...
+            _uiState.update {
+                it.copy(openExternallyLoading = false)
+            }
+        }
+    }
+
+    private fun setPreviewError() {
+        viewModelScope.launch {
+            // We need a small delay here because the pageFinished callback is called just after the page error and we don't want to overwrite the error state
+            delay(20)
```
domonkosadam, line 75: I don't really like this delay, isn't there a better way for this?

tamaskozmer: The other option is to disregard the onPageFinished callback and don't set the `previewState` to success when it already has an error. On a second thought this might be better, but I'm not sure.

inst-danger:
<p>Teacher Install Page</p>

inst-danger:

domonkosadam (approved):

PR #3056, opened 2025-06-16
[MBL-18929][All] Inbox signautre loads previous user signature if composing message with any FAB button

kristofnemere:
Test plan: Test all cases, even what already worked since I came up with a complete different solution.

refs: MBL-18929
affects: Student, Parent, Teacher
release note: Fixed a caching issue related to inbox signature.


inst-danger:
<p>Parent Install Page</p>

kdeakinstructure (approved):
QA 👍

PR #3076, opened 2025-06-23
[MBL-18721][Teacher] SpeedGrader grading

hermannakos:
Test plan: Check every grading type.

refs: MBL-18721
affects: Teacher
release note: none

## Checklist

- [x] Tested in dark mode
- [ ] A11y checked

### DIFF
--- a/apps/teacher/src/main/java/com/instructure/teacher/fragments/ViewImageFragment.kt
+++ b/apps/teacher/src/main/java/com/instructure/teacher/fragments/ViewImageFragment.kt
@@ -212,7 +212,7 @@ class ViewImageFragment : BaseCanvasFragment(), ShareableFile {
                 .load(it)
                 .listener(requestListener)
                 .into(binding.photoView)
-        } ?: {
+        } ?: run {
             load(url) {
                 Glide.with(requireContext())
                     .asBitmap()
--- a/libs/pandautils/src/main/java/com/instructure/pandautils/features/speedgrader/grade/grading/SpeedGraderGradingScreen.kt
+++ b/libs/pandautils/src/main/java/com/instructure/pandautils/features/speedgrader/grade/grading/SpeedGraderGradingScreen.kt
@@ -141,22 +141,24 @@ fun SpeedGraderGradingContent(uiState: SpeedGraderGradingUiState) {
                         .fillMaxWidth()
                         .padding(top = 16.dp)
                 ) {
-                    OutlinedButton(
-                        enabled = uiState.enteredScore != null || uiState.excused,
-                        onClick = { uiState.onScoreChange(null) },
-                        modifier = Modifier.weight(1f),
-                        border = BorderStroke(
-                            1.dp,
-                            LocalCourseColor.current.copy(alpha = if (uiState.enteredScore != null || uiState.excused) 1f else 0.5f)
-                        ),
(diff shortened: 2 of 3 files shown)


inst-danger:
<p>Parent Install Page</p>

kristofnemere (changes requested):
- [x] When clicking on ‘Write score here’, the cursor is not visible.
- [x] The slider and buttons do not match the design.
- [ ] When the keyboard closes, the entire bottom sheet also closes.
- [ ] Large numbers can be entered as a score, which causes the app to freeze.
- [x] Switching between portrait and landscape orientation behaves oddly, but this is a known issue for now.
- [ ] In Complete/Incomplete grading, there is a radio button option for ‘No Grade’, but it is not selectable.

inst-danger:
<p>Teacher Install Page</p>

kdeakinstructure (changes requested):
Speedgrader Grading QA Findings:

- [ ] I'm not always able to reproduce constantly but sometimes the grading drops an unexpected error with various scenarios. I managed to get one when I typed too many numbers manually to the grade, and even when I just scrolled the grade line, but it was occurred something in other cases, even when clicked on excuse or no grade buttons. (See attached 1. and 2. video) ---> I'm suspicious that it will constantly triggering this unexpected error if you tap on the exact same place where the grade line is current positioning, so just tap on the | line and you will get the unexpected error probably. But there are other cases as well just like i mentioned above the manual input field too long, clicking on buttons, etc.

- [ ] Grade line does not go back to start after clicking on 'No Grade' or 'Excuse' buttons. (I think it correlates with a 'similar' issue that the grade line is only refreshed after "losing focus" from the 'Write percentage here' input box and technically when we just clicking one of the buttons the focus loss event won't be triggering since there wasn't even a focus on the input field. ) (See attached 3. video)

- [ ] 'No Grade' radio button cannot be tapped at complete/incomplete grading type assignments. (See attached 4. video)

- [x] Grade line is disappearing when writing a value which is more than the maximum manually. (See attached 5. video)

- [x] 'Days late' seems like to be a constant value (with '1'), but it does not makes sense when there wasn't even a due date or if the due date is not expired yet.

- [x] 'Status' value is not always consistent with the grading line and manually written value (For example, it does not makes sense to me that if we grade a submission the status remains 'None'

- [x] Percentage and point grading type assignment grading line has different 'style' then the other (other are having a 'dotted' one and even has sound when moving my finger through it, despite the percentage and point grading types). (See attached 1. and 2. screenshots) (I did not find the 'dotted' one on the design).

- [ ] Rubrics cannot be seen yet on the 'Grade & Rubric' tab. (Maybe it wasn't the
 scope yet)


https://github.com/user-attachments/assets/6b4062f1-26b3-4b59-9a72-59e6cd42307b

https://github.com/user-attachments/assets/4cdcf9e5-748d-4c86-a867-12e8a15b8843


https://github.com/user-attachments/assets/28e2bd49-5e34-4420-9cea-77f00dd898ad


https://github.com/user-attachments/assets/867b5020-bc78-4e2a-a836-43845d3fbb61



https://github.com/user-attachments/assets/457aba31-0f06-4253-af33-ce3cb2a3173a

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

kdeakinstructure (approved):
QA 👍

PR #3082, opened 2025-06-26
[CLX-1811][Horizon] Inbox Details

domonkosadam:
refs: CLX-1811
affects: Horizon
release note: none

### DIFF
--- a/libs/horizon/src/main/java/com/instructure/horizon/features/inbox/details/HorizonInboxDetailsScreen.kt
+++ b/libs/horizon/src/main/java/com/instructure/horizon/features/inbox/details/HorizonInboxDetailsScreen.kt
@@ -74,7 +74,11 @@ import com.instructure.horizon.horizonui.organisms.inputs.textarea.TextAreaState
 import com.instructure.horizon.horizonui.platform.LoadingState
 import com.instructure.horizon.horizonui.platform.LoadingStateWrapper
 import com.instructure.pandautils.compose.composables.ComposeCanvasWebViewWrapper
+import com.instructure.pandautils.compose.composables.ComposeEmbeddedWebViewCallbacks
 import com.instructure.pandautils.room.appdatabase.entities.FileDownloadProgressState
+import com.instructure.pandautils.utils.ThemePrefs
+import com.instructure.pandautils.utils.getActivityOrNull
+import com.instructure.pandautils.utils.launchCustomTab
 import com.instructure.pandautils.utils.toFormattedString
 import java.util.Date
 
@@ -262,7 +266,16 @@ private fun HorizonInboxHtmlContent(
     content: String,
     modifier: Modifier = Modifier
 ) {
-    ComposeCanvasWebViewWrapper(content, modifier)
+    val activity = LocalContext.current.getActivityOrNull()
+
+    ComposeCanvasWebViewWrapper(
+        content,
+        modifier,
+        embeddedWebViewCallbacks = ComposeEmbeddedWebViewCallbacks(
+            shouldLaunchInternalWebViewFragment = { _ -> true },
+            launchInternalWebViewFragment = { url -> activity?.launchCustomTab(url, ThemePrefs.brandColor) }
+        ),
+    )
 }
 
 @Composable
--- a/libs/horizon/src/main/java/com/instructure/horizon/features/inbox/list/HorizonInboxListViewModel.kt
+++ b/libs/horizon/src/main/java/com/instructure/horizon/features/inbox/list/HorizonInboxListViewModel.kt
@@ -148,7 +148,7 @@ class HorizonInboxListViewModel @Inject constructor(
                                 recipientId -> recipients.firstOrNull { it.stringId == recipientId.toString() }
                             }?.map { it.name }?.joinToString(", ").orEmpty(),
                             courseId = conversation.contextCode?.substringAfter("course_")?.toLongOrNull(),
-                            date = conversation.lastAuthoredMessageSent,
+                            date = conversation.lastMessageSent,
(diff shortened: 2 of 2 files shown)


inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Horizon Install Page</p>

inst-danger:
<p>Student Install Page</p>

libs/horizon/src/main/java/com/instructure/horizon/features/inbox/details/HorizonInboxDetailsScreen.kt
```diff
@@ -16,21 +16,321 @@
...
+                Text(
+                    text = title,
+                    style = HorizonTypography.labelLargeBold,
+                    color = HorizonColors.Surface.institution(),
+                    maxLines = 2
+                )
+            }
+        },
+        navigationIcon = {
+            Image(
```
tamaskozmer, line 130: I think this should be an IconButton instead, we always used that for navigational elements, and it would make the touch feedback look better. It's also an IconButton on the designs.

libs/horizon/src/main/java/com/instructure/horizon/features/inbox/details/HorizonInboxDetailsScreen.kt
```diff
@@ -16,21 +16,321 @@
...
+
+        item.attachments.forEach { attachment ->
+            HorizonSpace(SpaceSize.SPACE_8)
+
+            val fileState = if (attachment.downloadState == FileDownloadProgressState.STARTING || attachment.downloadState == FileDownloadProgressState.IN_PROGRESS) {
+                FileDropItemState.InProgress(
+                    fileName = attachment.name,
+                    progress = attachment.downloadProgress,
+                    onActionClick = { attachment.onCancelDownloadClick(attachment.id) },
+                    onClick = { attachment.onDownloadClick(attachment) }
```
tamaskozmer, line 237: I don't think we need onClicks for these items, only for the actions, if you want to keep that for the `NoLongerEditable` item it's fine, but we only use that for previews where applicable.

tamaskozmer, line 237: There shouldn't be an `onDownloadClick` ever possible for an in progress item.

libs/horizon/src/main/java/com/instructure/horizon/features/inbox/details/HorizonInboxDetailsScreen.kt
```diff
@@ -16,21 +16,321 @@
...
+                        HorizonDivider()
+                    }
+                }
+            }
+        }
+    }
+}
+
+@Composable
+private fun HorizonInboxDetailsItem(
```
tamaskozmer, line 183: I think we need to completely differentiate between the message items and the announcement items here and also in the UiState, because the announcements can contain rich content and they should be shown in a WebView.

libs/horizon/src/main/java/com/instructure/horizon/features/inbox/details/HorizonInboxDetailsViewModel.kt
```diff
@@ -0,0 +1,358 @@
...
+
+            _uiState.update {
+                it.copy(
+                    loadingState = it.loadingState.copy(isLoading = false)
+                )
+            }
+        } catch {
+            _uiState.update {
+                it.copy(
+                    loadingState = it.loadingState.copy(isLoading = false, snackbarMessage = "Error")
```
tamaskozmer, line 98: This shouldn't be a hardcoded string.

libs/horizon/src/main/java/com/instructure/horizon/features/inbox/details/HorizonInboxDetailsViewModel.kt
```diff
@@ -0,0 +1,358 @@
...
+                    title = announcement.title.orEmpty(),
+                    titleIcon = R.drawable.ic_announcement,
+                    items = listOf(
+                        HorizonInboxDetailsItem(
+                            author = announcement.author?.displayName.orEmpty(),
+                            date = announcement.postedDate ?: Date(),
+                            content = announcement.message.orEmpty(),
+                            attachments = emptyList()
+                        )
+                    ) + topic.views.map { message ->
```
tamaskozmer, line 165: I see that in this case we also show the announcement replies. This can be tricky if we want to switch to WebView, but since we have a minimalistic UI I think it wouldn't be hard to just concatenate these into 1 HTML and show them in a WebView.

tamaskozmer (changes requested):
- [ ] Announcement icon's color is different on the designs.
- [ ] There is a white line under the reply block when the content is scrolled down.
- [ ] Images are not displayed in the announcements. Announcements can contain rich content, so those should be displayed in a WebView.
- [ ] This might be an edge case, but it could be nice to invalidate the inbox details cache after refreshing on the list. I sent some new messages for a thread, refreshed the list, and it showed up, but after opening the details the new messages weren't there because it loaded that thread from the cache.
- [ ] The app is logging me out when I try to download an attachment. (sent you the details for this on Slack)
- [ ] I think it would look much better to transform the send button into a loading indicator when sending a reply and refreshing. See the submission comments or assignment submission for example.

inst-danger:
<p>Horizon Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Horizon Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Horizon Install Page</p>

PR #3083, opened 2025-06-26
[CLX-1436][Horizon] Set up app localization.

tamaskozmer:
Test plan: -

refs: CLX-1436
affects: Horizon
release note: none


inst-danger:
<p>Horizon Install Page</p>

PR #3102, opened 2025-07-09
[MBL-18990][Teacher] Multi-Page PDF Submissions Cannot Be Scrolled

kristofnemere:
Test plan: check if pdf content works fine, also students can be swiped in speedgrader (outside of pdf content)

refs: MBL-18990
affects: Teacher
release note: none


inst-danger:
<p>Parent Install Page</p>

PR #3194, opened 2025-08-11
[PFS-25747][All] Disable Individual Recipient for Chechbox Teachers/Students

dhiraj14:
Test plan:

refs: PFS-25747
affects: Student, Teacher
release note: none

## Checklist

- [ ] Run E2E test suite
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] A11y checked
- [ ] Approve from product

### DIFF
--- a/apps/student/src/main/java/com/instructure/student/di/InboxModule.kt
+++ b/apps/student/src/main/java/com/instructure/student/di/InboxModule.kt
@@ -24,10 +24,12 @@ import com.instructure.canvasapi2.apis.GroupAPI
 import com.instructure.canvasapi2.apis.InboxApi
 import com.instructure.canvasapi2.apis.ProgressAPI
 import com.instructure.canvasapi2.apis.RecipientAPI
+import com.instructure.canvasapi2.managers.EnrollmentManager
 import com.instructure.canvasapi2.managers.InboxSettingsManager
 import com.instructure.pandautils.features.inbox.compose.InboxComposeRepository
 import com.instructure.pandautils.features.inbox.list.InboxRepository
 import com.instructure.pandautils.features.inbox.list.InboxRouter
+
 import com.instructure.student.features.inbox.compose.StudentInboxComposeRepository
 import com.instructure.student.features.inbox.list.StudentInboxRepository
 import com.instructure.student.features.inbox.list.StudentInboxRouter
@@ -42,7 +44,10 @@ import dagger.hilt.android.components.ViewModelComponent
 class InboxFragmentModule {
 
     @Provides
-    fun providesInboxRouter(activity: FragmentActivity, fragment: Fragment): InboxRouter {
+    fun providesInboxRouter(
+        activity: FragmentActivity, 
+        fragment: Fragment
+    ): InboxRouter {
         return StudentInboxRouter(activity, fragment)
     }
 }
--- a/apps/student/src/main/java/com/instructure/student/di/InboxModule.kt
+++ b/apps/student/src/main/java/com/instructure/student/di/InboxModule.kt
@@ -24,12 +24,10 @@ import com.instructure.canvasapi2.apis.GroupAPI
 import com.instructure.canvasapi2.apis.InboxApi
 import com.instructure.canvasapi2.apis.ProgressAPI
 import com.instructure.canvasapi2.apis.RecipientAPI
-import com.instructure.canvasapi2.managers.EnrollmentManager
 import com.instructure.canvasapi2.managers.InboxSettingsManager
 import com.instructure.pandautils.features.inbox.compose.InboxComposeRepository
 import com.instructure.pandautils.features.inbox.list.InboxRepository
 import com.instructure.pandautils.features.inbox.list.InboxRouter
-
 import com.instructure.student.features.inbox.compose.StudentInboxComposeRepository
 import com.instructure.student.features.inbox.list.StudentInboxRepository
 import com.instructure.student.features.inbox.list.StudentInboxRouter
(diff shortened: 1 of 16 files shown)


inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

apps/student/src/main/java/com/instructure/student/di/InboxModule.kt
```diff
@@ -42,7 +44,10 @@ import dagger.hilt.android.components.ViewModelComponent
 class InboxFragmentModule {
 
     @Provides
-    fun providesInboxRouter(activity: FragmentActivity, fragment: Fragment): InboxRouter {
+    fun providesInboxRouter(
```
tamaskozmer, line 47: These are unnecessary changes and unused imports and should be reverted, so the PR wouldn't be cluttered with irrelevant changes.

apps/student/src/main/java/com/instructure/student/features/inbox/list/StudentInboxRouter.kt
```diff
@@ -27,11 +28,16 @@ import com.instructure.pandautils.features.inbox.details.InboxDetailsFragment
...
 import com.instructure.student.activity.NavigationActivity
 import com.instructure.student.router.RouteMatcher
 import com.instructure.student.router.RouteMatcher.openMedia
+import javax.inject.Inject
 
-class StudentInboxRouter(private val activity: FragmentActivity, private val fragment: Fragment) : InboxRouter {
+class StudentInboxRouter(
+    private val activity: FragmentActivity, 
+    private val fragment: Fragment
+) : InboxRouter {
```
tamaskozmer, line 40: These are unnecessary changes and unused imports and should be reverted, so the PR wouldn't be cluttered with irrelevant changes.

apps/teacher/src/main/java/com/instructure/teacher/di/InboxModule.kt
```diff
@@ -44,7 +46,10 @@ import dagger.hilt.android.components.ViewModelComponent
 class InboxFragmentModule {
 
     @Provides
-    fun providesInboxRouter(activity: FragmentActivity, fragment: Fragment): InboxRouter {
+    fun providesInboxRouter(
```
tamaskozmer, line 49: These are unnecessary changes and unused imports and should be reverted, so the PR wouldn't be cluttered with irrelevant changes.

apps/teacher/src/main/java/com/instructure/teacher/features/inbox/list/TeacherInboxRouter.kt
```diff
@@ -39,8 +41,12 @@ import com.instructure.teacher.fragments.EmptyFragment
 import com.instructure.teacher.router.RouteMatcher
 import com.instructure.teacher.router.RouteMatcher.openMedia
 import com.instructure.teacher.utils.setupBackButtonAsBackPressedOnly
+import kotlinx.coroutines.launch
 
-class TeacherInboxRouter(private val activity: FragmentActivity, private val fragment: Fragment) : InboxRouter {
+class TeacherInboxRouter(
+    private val activity: FragmentActivity, 
+    private val fragment: Fragment
+) : InboxRouter {
```
tamaskozmer, line 49: These are unnecessary changes and unused imports and should be reverted, so the PR wouldn't be cluttered with irrelevant changes.

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/InboxComposeViewModel.kt
```diff
@@ -185,6 +188,35 @@ class InboxComposeViewModel @Inject constructor(
         }
     }
 
+    private fun checkAndApplyFeatureFlagRestrictions() {
+        viewModelScope.launch {
+            val isTeacherApp = context.packageName.contains("teacher")
```
tamaskozmer, line 193: This is not how we should differentiate between app types in common code. In these cases we use classes/interfaces called `Behavior`, where we have an interface for example in this case it would be `InboxComposeBehavior` in the common code, and it has different implementation in the specific apps module, and we provide the implementation for the specific apps in their own module through Hilt. In this case you would probably only need to do 1 implementation for this behavior class which would have a function `shouldRestrictStudentAccess` and it can have a default value as false, and the Teacher implementation would implement the feature flag chacking logic. You can find a good example about how this pattern works by searching for the `SettingsBehavior` interface.

libs/pandautils/src/main/java/com/instructure/pandautils/utils/FeatureFlagProvider.kt
```diff
@@ -54,7 +54,7 @@ class FeatureFlagProvider(
         return checkEnvironmentFeatureFlag(FEATURE_FLAG_OFFLINE) && !apiPrefs.canvasForElementary
     }
 
-    private suspend fun checkEnvironmentFeatureFlag(featureFlag: String): Boolean {
+    suspend fun checkEnvironmentFeatureFlag(featureFlag: String): Boolean {
```
tamaskozmer, line 57: Since this function was made public it is necessary to write unit tests for it.

apps/student/src/main/java/com/instructure/student/di/InboxModule.kt
tamaskozmer: I suggest that you just revert all these changes for the files where I added this comment.

tamaskozmer (changes requested):

dhiraj14:
@tamaskozmer Thanks for the review and feedbacks, I have applied the changes. Could you please have a look again?
Thanks!

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

PR #3196, opened 2025-08-13
[PFS-25667][All]Hide Conversation Deletion Button and Reply All for all apps when restricted_student_access is enabled

dhiraj14:
Test plan:

refs: 
affects: PFS-25748, PFS-25749
release note: none

## Checklist

- [ ] Run E2E test suite
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] A11y checked
- [ ] Approve from product


inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

apps/parent/src/main/java/com/instructure/parentapp/features/main/MainActivity.kt
```diff
@@ -198,6 +203,16 @@ class MainActivity : BaseCanvasActivity(), OnUnreadCountInvalidated, Masqueradin
         toast(R.string.fileQuotaExceeded)
     }
 
+    private fun fetchFeatureFlags() {
+        lifecycleScope.launch {
+            try {
+                featureFlagProvider.fetchEnvironmentFeatureFlags()
```
tamaskozmer, line 209: I would move this to the `SplashViewModel.loadinitialData`, that's where we do the initial data loading for the app.

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/details/InboxDetailsBehavior.kt
```diff
@@ -20,4 +20,6 @@ import android.content.Context
 
 open class InboxDetailsBehavior {
     open fun getShowBackButton(context: Context): Boolean = true
+    open suspend fun shouldRestrictDeleteConversation(): Boolean = false
+    open suspend fun shouldRestrictReplyAll(): Boolean = false
```
tamaskozmer, line 24: Same here. We don't really need these functions if the implementation is the same for all 3 apps.

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/compose/InboxComposeViewModel.kt
```diff
@@ -192,15 +192,17 @@ class InboxComposeViewModel @Inject constructor(
     private fun checkAndApplyFeatureFlagRestrictions() {
         viewModelScope.launch {
             val shouldRestrict = inboxComposeBehavior.shouldRestrictStudentAccess()
+            val shouldHideSendIndividual = inboxComposeBehavior.shouldHideSendIndividual()
+            
             if (shouldRestrict) {
```
tamaskozmer, line 197: I don't understand this logic. Please explain it to me why is this needed. It seems like we only update the UiState to `sendIndividual` true when the `shouldRestrict` is true. It will be true for every app when the feature flag is enabled. However we use the `shouldHideSendIndividual` to determine if we should hide that switch. This will be only true for the teacher app. So what is the point of setting the `sendIndividual` true in the case of Parent and Student when that's not hidden and the users can change that.

If we should always hide the "send individual" checkbox for all 3 apps maybe we don't even need the `InboxComposeBehavior` at all and can move all the logic to the common code, but maybe I am missing some requirements here.

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/details/composables/InboxDetailsScreen.kt
```diff
@@ -290,8 +290,9 @@ private fun InboxDetailsContentView(
 }
 
 @Composable
-private fun AppBarMenu(conversation: Conversation?, actionHandler: (InboxDetailsAction) -> Unit) {
+private fun AppBarMenu(uiState: InboxDetailsUiState, actionHandler: (InboxDetailsAction) -> Unit) {
```
tamaskozmer, line 293: I would avoid passing the whole `uiState` here if only just a couple of properties needed. You can just pass those 2 boolean flags instead.

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/list/InboxViewModel.kt
```diff
@@ -40,7 +42,8 @@ import javax.inject.Inject
 class InboxViewModel @Inject constructor(
     private val inboxRepository: InboxRepository,
     private val resources: Resources,
-    private val inboxEntryItemCreator: InboxEntryItemCreator
+    private val inboxEntryItemCreator: InboxEntryItemCreator,
+    private val inboxDetailsBehavior: InboxDetailsBehavior
```
tamaskozmer, line 46: `InboxViewModel` is the ViewModel for the Inbox List. We shouldn't inject behavior classes that are responsible for a different screens behavior. Alternatively if it is needed we can create a generic behavior class for these restrictions and inject it to all inbox related ViewModels, but as I mentioned in an other comment we might not even need that.

tamaskozmer (changes requested):

inst-danger:
<p>Student Install Page</p>

libs/pandautils/src/main/java/com/instructure/pandautils/features/inbox/details/InboxDetailsViewModel.kt
```diff
@@ -92,20 +100,27 @@ class InboxDetailsViewModel @Inject constructor(
...
-                negativeButton = context.getString(R.string.cancel),
-                onPositiveButtonClick = {
-                    deleteConversation(action.conversationId)
-                    _uiState.update { it.copy(confirmationDialogState = ConfirmationDialogState()) }
-                },
-                onNegativeButtonClick = {
-                    _uiState.update { it.copy(confirmationDialogState = ConfirmationDialogState()) }
+            is InboxDetailsAction.DeleteConversation -> {
+                viewModelScope.launch {
+                    val shouldRestrict = checkRestrictStudentAccessFlag()
```
tamaskozmer, line 105: Wouldn't it be easier to just save this flag as a class variable when we already initializing this call in the `checkAndAppyFeatureFlagRestrictions`? In that case we won't need a coroutine here.

apps/parent/src/main/java/com/instructure/parentapp/features/splash/SplashViewModel.kt
```diff
@@ -90,6 +90,13 @@ class SplashViewModel @Inject constructor(
                 Pendo.endSession()
             }
 
+            // Fetch environment feature flags for app functionality
+            try {
+                featureFlagProvider.fetchEnvironmentFeatureFlags()
```
tamaskozmer, line 95: This might not be related to this feature, but we rely on a feature flag in line 86, so it would be better to initialize the feature flags before that. I would move it under this block

```
val theme = repository.getTheme()
theme?.let { _events.send(SplashAction.ApplyTheme(it)) }
```

tamaskozmer (changes requested):

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

apps/teacher/src/main/java/com/instructure/teacher/features/inbox/details/TeacherInboxDetailsBehavior.kt
```diff
@@ -19,7 +19,9 @@ package com.instructure.teacher.features.inbox.details
 import android.content.Context
 import com.instructure.pandautils.features.inbox.details.InboxDetailsBehavior
 import com.instructure.teacher.utils.isTablet
+import javax.inject.Inject
 
-class TeacherInboxDetailsBehavior: InboxDetailsBehavior() {
+class TeacherInboxDetailsBehavior @Inject constructor() : InboxDetailsBehavior() {
```
tamaskozmer, line 24: Changes to this class can be reverted.

inst-danger:
<p>Parent Install Page</p>

PR #3204, opened 2025-08-15
[MBL-19122][All] Calendar Discussion Checkpoints

andrasmaczak:
Test plan: 
Create a graded discussion with checkpoints
Open calendar in all 3 apps, item should be visible

refs: MBL-19122
release note:

## Checklist

- [x] Run E2E test suite
- [x] Tested in dark mode
- [x] Tested in light mode
- [ ] A11y checked
- [ ] Approve from product

### DIFF
--- a/apps/parent/src/main/java/com/instructure/parentapp/features/calendar/ParentCalendarRepository.kt
+++ b/apps/parent/src/main/java/com/instructure/parentapp/features/calendar/ParentCalendarRepository.kt
@@ -90,13 +90,29 @@ class ParentCalendarRepository(
                     .mapContextName()
             }
 
+            val calendarSubAssignments = async {
+                calendarEventApi.getCalendarEvents(
+                    false,
+                    CalendarEventAPI.CalendarEventType.SUB_ASSIGNMENT.apiName,
+                    startDate,
+                    endDate,
+                    contextCodes,
+                    restParams
+                ).depaginate {
+                    calendarEventApi.next(it, restParams)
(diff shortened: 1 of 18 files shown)


inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Student Install Page</p>

kristofnemere (approved):
QA+1

inst-danger:
<p>Student Install Page</p>

kdeakinstructure (changes requested):
QA Findings:

Teacher and Parent:

- [x] 'Reply to topic' should be used instead of 'Reply to Topic' according to the design
- [x] '-' is missing after the graded discussion's name and the corresponding checkpoint's name (It's in the design and I also think it's needed to 'separate' the discussion's name from the checkpoints' names)
- [x] In the design the 2nd checkpoints name is 'Additional replies' but in these 2 apps we are using 'Required replies' instead of that. (Note: In the student, the string name is corrent as it's 'Additional replies' there).

- [x] TEACHER only: On the 'Due dates' screen (opening 'Due' section on the graded discussion's details page) I think it would be better if the multiple due dates would be ordered by due date ASCENDING. Currently the latest due date will be listed as 1st but I think it would be better to see the earliest due date on the top.

Student:

- [x] Discussion checkpoint strings looks like they're "under" the discussion's name instead of "being on the same level" as the discussion's name just like on the designs (and in the other 2 apps).

PR #3234, opened 2025-09-05
[CLX-2539] Programs on dashboard

tamaskozmer:
Test plan: Check that programs information is present on the Dashboard and the navigation works correctly.

refs: CLX-2539
affects: Student
release note: none

## Checklist

- [x] Tested in light mode

### DIFF
--- a/libs/horizon/src/main/java/com/instructure/horizon/features/learn/LearnScreen.kt
+++ b/libs/horizon/src/main/java/com/instructure/horizon/features/learn/LearnScreen.kt
@@ -199,7 +199,9 @@ private fun LearnScreenWrapper(
                         )
                     }
                     val programDetailsState by programDetailsViewModel.state.collectAsState()
-                    ProgramDetailsScreen(programDetailsState, state.onCourseSelected)
+                    ProgramDetailsScreen(programDetailsState, { courseId ->
+                        state.onProgramCourseSelected(state.selectedLearningItem.program.id, courseId)
+                    })
                 }
 
                 else -> {
--- a/libs/horizon/src/main/java/com/instructure/horizon/features/learn/LearnUiState.kt
+++ b/libs/horizon/src/main/java/com/instructure/horizon/features/learn/LearnUiState.kt
@@ -25,7 +25,7 @@ data class LearnUiState(
     val learningItems: List<LearningItem> = emptyList(),
     val selectedLearningItem: LearningItem? = null,
     val onSelectedLearningItemChanged: ((LearningItem) -> Unit) = {},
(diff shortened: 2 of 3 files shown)


inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Student Install Page</p>

libs/horizon/src/main/java/com/instructure/horizon/features/dashboard/DashboardScreen.kt
```diff
@@ -209,12 +235,28 @@ private fun HomeScreenTopBar(uiState: DashboardUiState, mainNavController: NavCo
     }
 }
 
+@Composable
+private fun DashboardProgramItem(
+    programUiState: DashboardProgramUiState,
+    onProgramClick: () -> Unit
+) {
+    Column {
+        Text(text = programUiState.name, style = HorizonTypography.h2)
```
domonkosadam, line 244: Shouldn't be the program's header clickable as it is with courses?

libs/horizon/src/main/java/com/instructure/horizon/features/dashboard/DashboardScreen.kt
```diff
@@ -251,6 +297,41 @@ private fun DashboardCourseItem(
...
 
+@Composable
+private fun ProgramsText(
+    programs: List<DashboardCourseProgram>,
+    onProgramClick: (String) -> Unit
+) {
+    val programsAnnotated = buildAnnotatedString {
+        programs.forEachIndexed { i, program ->
+            if (i > 0) append(", ")
+            withLink(
```
domonkosadam, line 308: I know we are not focusing an a11y, but I think it would be better to embed the text in a button. I think the ripple effect would be beneficial as well.

tamaskozmer: I couldn't do it without compromising i18n guidelines. If we want to make it a button we need to split the string into multiple views instead of using a parameterised string. If you have any good solutions for this I am open to suggestions.

libs/horizon/src/main/java/com/instructure/horizon/features/dashboard/DashboardScreen.kt
tamaskozmer: There is a `My Program` button that is clickable and navigates to the program. This is how it's done on the web as well.

PR #3265, opened 2025-09-25
[MBL-19372][Parent] - Fix Alert Settings dialog message appearance

kdeakinstructure:
Fix Alert Settings error message to be shown in every case, regardless of type and the size of the given number.

refs: MBL-19298
affects: Parent
release note: Always show Alert Settings error message regardless of type and number size.


inst-danger:
<p>Parent Install Page</p>

inst-danger:
<p>Teacher Install Page</p>

inst-danger:
<p>Student Install Page</p>

adamNagy56 (changes requested):
- [x] As we discussed, when saving a value, any leading zeros should be removed so that only the valid number is stored and displayed.

See attached video.

https://github.com/user-attachments/assets/6be50320-ed4d-4293-82f1-4a19a99cb76d

apps/parent/src/main/java/com/instructure/parentapp/features/alerts/settings/AlertSettingsScreen.kt
```diff
@@ -469,15 +469,18 @@ private fun ThresholdDialog(
                 )
             )
             val errorText = when {
-                (percentage.toIntOrNull() ?: 100) <= min ->
+                percentage.toBigDecimalOrNull() == null && percentage.isNotEmpty() ->
```
kristofnemere, line 472: You forget about this line
`val enabled = percentage.toIntOrNull().orDefault() in (min + 1)..<max`
thats why it allows to save invalid values. 

TBH since this is just a percentage, I would simply allow max 3 charachters in the text field, that would solve the issue you found. To do so just add this in the onValueChange of the TextField

`onValueChange = {
                    if (it.length <= 3){
                        percentage = it
                    }
                }`

kristofnemere (changes requested):

inst-danger:
<p>Parent Install Page</p>

adamNagy56 (approved):

PR #3620, opened 2026-04-02
[CLXR-455][Horizon] Implement offline POC

domonkosadam:
refs: CLXR-455
affects: Student
release note: none


libs/horizon/src/main/java/com/instructure/horizon/features/dashboard/widget/course/DashboardCourseViewModel.kt
```diff
@@ -42,9 +42,12 @@ import javax.inject.Inject
 @HiltViewModel
 class DashboardCourseViewModel @Inject constructor(
     @ApplicationContext private val context: Context,
-    private val repository: DashboardCourseRepository,
-    private val dashboardEventHandler: DashboardEventHandler
-): ViewModel() {
+    private val getDashboardCoursesUseCase: GetDashboardCoursesUseCase,
+    private val dashboardEventHandler: DashboardEventHandler,
+    networkStateProvider: NetworkStateProvider,
```
tamaskozmer, line 48: You can use composition instead. Creating a class that encapsulates the behavior of the HorizonOfflineViewModel, in that case you won't need to pass all the parameters to the ViewModels instead just pass that new class.

libs/horizon/src/main/java/com/instructure/horizon/features/dashboard/DashboardViewModel.kt
```diff
@@ -40,7 +46,10 @@ class DashboardViewModel @Inject constructor(
     private val apiPrefs: ApiPrefs,
     private val themePrefs: ThemePrefs,
     private val localeUtils: LocaleUtils,
-    private val dashboardEventHandler: DashboardEventHandler
+    private val dashboardEventHandler: DashboardEventHandler,
+    private val networkStateProvider: NetworkStateProvider,
+    private val featureFlagProvider: FeatureFlagProvider,
+    private val getLastSyncedAtUseCase: GetLastSyncedAtUseCase,
 ) : ViewModel() {
```
tamaskozmer: Shouldn't we use HorizonOfflineViewModel here as well?

libs/horizon/src/main/java/com/instructure/horizon/domain/usecase/GetLastSyncedAtUseCase.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.instructure.horizon.database.dao.HorizonSyncMetadataDao
+import com.instructure.horizon.database.entity.SyncDataType
+import com.instructure.pandautils.domain.usecase.BaseUseCase
+import javax.inject.Inject
+
+class GetLastSyncedAtUseCase @Inject constructor(
+    private val syncMetadataDao: HorizonSyncMetadataDao,
+) : BaseUseCase<SyncDataType, Long?>() {
+
+    override suspend fun execute(params: SyncDataType): Long? = syncMetadataDao.getLastSyncedAt(params)
```
tamaskozmer: If I remember correctly we agreed with Ákos to have the params as an inner class of the Use Case as a convention. I am not sure if this is applicable where we have a single parameter only, but I see some occurrences that suggest this, for example `GetAuthenticatedSessionUseCase`. What do you think @hermannakos ?

libs/horizon/src/main/java/com/instructure/horizon/data/repository/ProgramRepository.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.instructure.horizon.data.datasource.ProgramLocalDataSource
+import com.instructure.horizon.data.datasource.ProgramNetworkDataSource
+import com.instructure.horizon.offline.OfflineSyncRepository
+import com.instructure.pandautils.utils.FeatureFlagProvider
+import com.instructure.pandautils.utils.NetworkStateProvider
+import javax.inject.Inject
+
+class ProgramRepository @Inject constructor(
+    private val networkDataSource: ProgramNetworkDataSource,
+    private val localDataSource: ProgramLocalDataSource,
```
tamaskozmer, line 28: Why don't we have a data source interface like in the current offline implementation?

tamaskozmer (approved):

libs/horizon/src/main/java/com/instructure/horizon/domain/usecase/GetLastSyncedAtUseCase.kt
hermannakos: Yeah, I would go with the inner class for consistency at least. We didn't follow this convention at first since we didn't have that, so there might be some where it is not followed

PR #3642, opened 2026-04-15
Fixed Parent build.

tamaskozmer:


kdeakinstructure (approved):
Could you check please if there is a unit test for this (like you mentioned there is in the Teacher)

kristofnemere (approved):

PR #3650, opened 2026-04-17
[MBL-19926][Student] InstUI Design System Foundation

hermannakos:
## Summary

Introduces the `libs/instui` module — a design token system and initial Compose component library for the InstUI design system, synced with [instructure-ui](https://github.com/instructure/instructure-ui) (web).

### Token Generation (3 layers, auto-generated from instructure-ui v11.7.1)
- **Primitives** — colors, sizes, font weights, font families, opacities
- **Semantics** — theme-aware light/dark colors, layout sizes, typography, elevation
- **Components** — 65 per-component token files (Text, Heading, Button, Pill, List, etc.) with composed `TextStyle` objects, `@Composable` color accessors, and `sp` font sizes

### Compose Components (hand-written, thin wrappers)
- `InstUITheme` — MaterialTheme wrapper with InstUI defaults
- `Text` / `Heading` — typography composables using generated TextStyles
- `Pill` — status badge with 5 variants (Default, Info, Error, Success, Warning)
- `ListItem` — leading/content/trailing/bottom slot layout
- `SectionHeader` — collapsible header with trailing slot
- `Separator` — themed horizontal divider
- `Card` — elevated surface with token-based border radius and padding
- `TopBar` — navigation bar with title, subtitle, and icon slots
- `GradesScreenPreview` — dummy preview assembling all components per Figma spec

### Other
- `BorderStyle` enum mapping CSS border styles to Compose `PathEffect`
- Font files: Atkinson Hyperlegible Next, Inclusive Sans, Lato
- NGC added to translation projects

## Test plan

1. Open `libs/instui` in Android Studio
2. Run Compose Previews in `Text.kt`, `Pill.kt`, `Card.kt`, `ListItem.kt`, `SectionHeader.kt`, `TopBar.kt`
3. Verify light and dark mode previews render correctly
4. Open `GradesScreenPreview.kt` and verify the full Grades screen preview matches the Figma design
5. Run `./gradlew :instui:compileDebugKotlin` — should pass
6. Run `cd libs/instui/scripts && npm install && npm run build` — should regenerate all token files

- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] A11y checked

refs: MBL-19926
affects: Student
release note: none (foundation work, no user-facing changes)


libs/instui/build.gradle.kts
```diff
@@ -0,0 +1,93 @@
...
+        consumerProguardFiles("consumer-rules.pro")
+    }
+
+    buildTypes {
+        debug {
+            isMinifyEnabled = false
+        }
+        release {
+            isMinifyEnabled = false
+            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
```
kristofnemere, line 39: We should add proguard-rules.pro and consumer-rules.pro files

libs/instui/src/main/java/com/instructure/instui/compose/InstUITheme.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.compose.material3.LocalTextStyle
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import com.instructure.instui.token.component.InstUIText
+import com.instructure.instui.token.semantic.InstUISemanticColors
+
+@Composable
+fun InstUITheme(content: @Composable () -> Unit) {
+    MaterialTheme {
```
kristofnemere, line 29: We should pass a colorscheme and maybe typography

libs/instui/src/main/java/com/instructure/instui/compose/container/Card.kt
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    androidx.compose.material3.Card(
+        modifier = modifier,
+        shape = shape,
+        colors = CardDefaults.cardColors(containerColor = backgroundColor),
+        elevation = CardDefaults.cardElevation(defaultElevation = elevation.dp),
+        border = border,
+    ) {
+        Column(modifier = Modifier.padding(contentPadding)) {
```
kristofnemere: Content is wrapped in a Column this will crash if a caller puts a LazyColumn inside Card {}. Use Box(Modifier.padding(contentPadding)) instead

libs/instui/src/main/java/com/instructure/instui/compose/indicator/Pill.kt
```diff
@@ -0,0 +1,177 @@
...
+    val borderColor = when (variant) {
+        PillVariant.Default -> InstUIPill.baseBorderColor
+        PillVariant.Info -> InstUIPill.infoBorderColor
+        PillVariant.Error -> InstUIPill.errorBorderColor
+        PillVariant.Success -> InstUIPill.successBorderColor
+        PillVariant.Warning -> InstUIPill.warningBorderColor
+    }
+
+    val shape = RoundedCornerShape(InstUIPill.borderRadius)
+    val textStyle = TextStyle(
```
kristofnemere: TextStyle(fontFamily, fontWeight, fontSize) is allocated on every recomposition. The token generator already emits composed TextStyle objects for other components (e.g. InstUIText.content. Should expose a textStyle composite too, or wrap in remember

libs/instui/src/main/java/com/instructure/instui/compose/list/SectionHeader.kt
```diff
@@ -0,0 +1,122 @@
...
+    title: String,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    trailing: (@Composable () -> Unit)? = null,
+) {
+    Row(
+        modifier = modifier
+            .fillMaxWidth()
+            .clickable(onClick = onClick)
+            .padding(horizontal = 16.dp, vertical = 8.dp),
```
kristofnemere: Horizontal/vertical padding is hardcoded (16.dp, 8.dp) while all other components use tokens. Should use InstUILayoutSizes.Spacing.Padding.container_sm and InstUILayoutSizes.Spacing.SpaceSm.spaceSm

libs/instui/scripts/build-instui-icons.js
```diff
@@ -0,0 +1,358 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.PathFillType
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.graphics.vector.PathParser
+import androidx.compose.ui.unit.dp
+import ${BASE_PACKAGE}.InstUIIcons
+
+val InstUIIcons.${category}.${kotlinName}: ImageVector by lazy {
+    ImageVector.Builder(
```
kristofnemere, line 194: Directional icons (ArrowLeft, ArrowRight, Back, ArrowOpenEnd/Start, chevrons) need autoMirror = true in the ImageVector.Builder call, otherwise they point the wrong way in RTL locales

kristofnemere (changes requested):

libs/instui/src/main/java/com/instructure/instui/compose/indicator/Pill.kt
```diff
@@ -0,0 +1,176 @@
...
+) {
+    val textColor = when (variant) {
+        PillVariant.Default -> InstUIPill.baseTextColor
+        PillVariant.Info -> InstUIPill.infoTextColor
+        PillVariant.Error -> InstUIPill.errorTextColor
+        PillVariant.Success -> InstUIPill.successTextColor
+        PillVariant.Warning -> InstUIPill.warningTextColor
+    }
+
+    val borderColor = when (variant) {
```
tamaskozmer: This solution is not too elegant, and much more harder to maintain and expand in the future. For example if you will have a new property depending on a variant you will add a new condition. For a new variant you would need to maintain all the conditions. See how this is done in the Horizon design system. Enum values can have properties as well, or you can make these sealed classes also.

tamaskozmer: The difference is that these are composable colors instead of the non-composable colors in Horizon. We can either generate the dark/light theme dependent colors in a different way to make them non-composable or use the enum like this:

```
enum class PillType(
      val shapeColor: @Composable () -> Color,                                                                                                                                                                                          
      val textColor: @Composable () -> Color,                                             
      val filledTextColor: @Composable () -> Color,                                                                                                                                                                                     
      val iconColor: @Composable () -> Color                                                                                                                                                                                            
  ) {                                                                                                                                                                                                                                   
      DEFAULT(                                                                                                                                                                                                                          
          shapeColor = { HorizonColors.Surface.inversePrimary() },                                                                                                                                                                      
          textColor = { HorizonColors.Text.body() },                                                                                                                                                                                    
          filledTextColor = { HorizonColors.Text.surfaceColored() },
          iconColor = { HorizonColors.Icon.default() }                                                                                                                                                                                  
      ),                                                                                                                                                                                                                                
      ...
  }   
```

or

```
data class PillColors(                                                                                                                                                                                                                
      val shapeColor: Color,                                                                                                                                                                                                            
      val textColor: Color,                                                                                                                                                                                                             
      val filledTextColor: Color,                                                                                                                                                                                                       
      val iconColor: Color                                                                                                                                                                                                              
  )                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                        
  enum class PillType(val colors: @Composable () -> PillColors) {                                                                                                                                                                       
      DEFAULT(colors = {
          PillColors(                                                                                                                                                                                                                   
              shapeColor = HorizonColors.Surface.inversePrimary(), 
              textColor = HorizonColors.Text.body(),                                                                                                                                                                                    
              filledTextColor = HorizonColors.Text.surfaceColored(),
              iconColor = HorizonColors.Icon.default()                                                                                                                                                                                  
          )                                                        
      }),   
```

tamaskozmer: We should use this approach everywhere in the new design system.

libs/instui/src/main/java/com/instructure/instui/compose/list/ListItem.kt
```diff
@@ -0,0 +1,198 @@
...
+ *     leading = { Icon(...) },
+ *     bottom = { Pill(text = "Missing", variant = PillVariant.Error) },
+ * )
+ * ```
+ */
+@Composable
+fun ListItem(
+    title: String,
+    modifier: Modifier = Modifier,
+    titleStyle: TextStyle = InstUIHeading.titleCardMini,
```
tamaskozmer: I think we don't need this.

libs/instui/src/main/java/com/instructure/instui/compose/list/SectionHeader.kt
```diff
@@ -0,0 +1,102 @@
...
+            text = title,
+            style = TextStyle(
+                fontFamily = InstUIToggleDetails.fontFamily,
+                fontWeight = InstUIToggleDetails.fontWeight,
+                fontSize = InstUIToggleDetails.fontSizeSmall,
+            ),
+            color = InstUIToggleDetails.textColor,
+            modifier = Modifier.weight(1f),
+        )
+        if (trailing != null) {
```
tamaskozmer: In the design system we have an icon here instead of a general composable. We can go this way, but in that case I would ask Marci to modify the design as well.

libs/instui/src/main/java/com/instructure/instui/compose/list/Separator.kt
```diff
@@ -0,0 +1,71 @@
...
+ * ```
+ * Separator()
+ * Separator(modifier = Modifier.padding(horizontal = 16.dp))
+ * ```
+ */
+@Composable
+fun Separator(
+    modifier: Modifier = Modifier,
+    color: Color = InstUISemanticColors.Background.Divider.base(),
+    thickness: Dp = 1.dp,
```
tamaskozmer, line 47: I think this should be tokenized as well.

libs/instui/src/main/java/com/instructure/instui/compose/navigation/TopBar.kt
```diff
@@ -0,0 +1,117 @@
...
+ * ```
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun TopBar(
+    title: String,
+    modifier: Modifier = Modifier,
+    subtitle: String? = null,
+    containerColor: Color = InstUISemanticColors.Background.brand(),
+    contentColor: Color = InstUISemanticColors.Text.onColor(),
```
tamaskozmer: This should be `inverseColor` instead.

libs/instui/src/main/java/com/instructure/instui/compose/indicator/Pill.kt
```diff
@@ -161,13 +165,11 @@ private fun PillWithIconPreview() {
...
                 variant = PillVariant.Error,
                 icon = {
-                    Box(
-                        modifier = Modifier
-                            .size(12.dp)
-                            .background(
-                                InstUISemanticColors.Icon.error(),
-                                RoundedCornerShape(50),
-                            )
+                    Icon(
```
tamaskozmer: I think we should have our own Icon composable that could take for example sizing properties from tokens. In that case we could remove also that hardcoded 12.dp.

tamaskozmer (changes requested):

libs/instui/src/main/java/com/instructure/instui/compose/InstUITheme.kt
hermannakos: We're not using M3 but our own design system, so no need for them.

kristofnemere (approved):

tamaskozmer (approved):

PR #3679, opened 2026-05-04
[TC-007] - Prevent response body leakage in RestRetryInterceptor

kdeakinstructure:
Prevent response body leakage in RestRetryInterceptor
Instead, only print it internally and throw only statuscode and content-type.

### DIFF
--- a/automation/dataseedingapi/src/main/kotlin/com/instructure/dataseeding/util/RestRetryInterceptor.kt
+++ b/automation/dataseedingapi/src/main/kotlin/com/instructure/dataseeding/util/RestRetryInterceptor.kt
@@ -40,8 +40,11 @@ object RestRetryInterceptor : Interceptor {
 
         if (response.failed) {
             val code = response.code
+            val contentType = response.body?.contentType()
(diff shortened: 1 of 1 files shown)


automation/dataseedingapi/src/main/kotlin/com/instructure/dataseeding/util/RestRetryInterceptor.kt
```diff
@@ -40,8 +40,11 @@ object RestRetryInterceptor : Interceptor {
 
         if (response.failed) {
             val code = response.code
+            val contentType = response.body?.contentType()
             val body = response.body?.string()
-            throw RuntimeException("status code: $code\nbody: $body")
+            println("RestRetryInterceptor: all retries exhausted. status=$code, content-type=$contentType, body=$body")
```
hermannakos: Don't have this `println` at all.

hermannakos (changes requested):

kdeakinstructure:
Will be reopened in new repo.

PR #3680, opened 2026-05-04
[All] PKCE OAuth login + encrypt credentials at rest

hermannakos:
Test plan:

**PKCE login**
1. Build Student / Teacher / Parent and sign in with a test account on a hosted Canvas instance.
2. With WebView debugging (`chrome://inspect`), confirm the auth URL contains both `code_challenge=…` and `code_challenge_method=S256`.
3. Confirm the token exchange succeeds and the user lands in the app.
4. Background and reopen the app to confirm refresh-token flow still works.
5. Sign out and sign back in to verify a different `code_challenge` is generated each time.

**Credential encryption at rest**
1. Upgrade install: install a build that predates this PR, sign in, then upgrade to a build with this PR. Confirm the app does not require re-authentication (lazy migration of plaintext token + signed-in-user blob succeeds on first read).
2. Inspect `/data/data/<pkg>/shared_prefs/canvas-kit-sp.xml` and `signedInUsersList.xml` (debuggable build, `adb shell run-as`). After login, all credential-shaped values (`token`, `accessToken`, `refreshToken`, `clientSecret`, `notoriousToken`, plus the `SignedInUser` JSON blob) should begin with `enc1:` and be opaque base64.
3. Sign out, sign back in with a different account, then switch back via the previous-users picker to confirm decryption works across the multi-account flow.
4. Force-stop and relaunch — confirm session persistence is intact.

release note: Hardened OAuth login with PKCE and encrypted authentication credentials at rest.

## Checklist

- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Test in landscape mode and/or tablet
- [ ] A11y checked
- [ ] Approve from product

### DIFF
--- a/libs/canvas-api-2/build.gradle
+++ b/libs/canvas-api-2/build.gradle
@@ -175,6 +175,7 @@ dependencies {
         exclude group: 'com.google.firebase', module: 'firebase-analytics'
     }
     implementation Libs.FIREBASE_CONFIG
+    implementation Libs.FIREBASE_CRASHLYTICS
 
     implementation Libs.HILT
     ksp Libs.HILT_COMPILER
--- a/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/apis/OAuthAPI.kt
+++ b/libs/canvas-api-2/src/main/java/com/instructure/canvasapi2/apis/OAuthAPI.kt
@@ -49,6 +49,7 @@ object OAuthAPI {
                 @Field("client_secret") clientSecret: String,
                 @Field("code") oAuthRequest: String,
                 @Field(value = "redirect_uri", encoded = true) redirectURI: String,
+                @Field("code_verifier") codeVerifier: String? = null,
                 @Field("grant_type") grantType: String = "authorization_code"): Call<OAuthTokenResponse>
 
         @GET("/login/session_token")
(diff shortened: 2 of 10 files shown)


hermannakos:
Thanks for the review. Status on each comment:

**1. `SecureStringPref.onClear()` empty body** — leaving as-is. `PrefManager.clearPrefs()` ends with `editor.clear().apply()` which wipes the whole prefs file, so per-delegate cleanup is redundant. The sibling `StringPref.onClear()` is empty for the same reason; this matches existing convention.

**2. `pkce` save/restore in `onSaveInstanceState`** — leaving as-is. The activity has no `configChanges` set, so a recreation re-runs `onCreate` → `beginSignIn` → `mobileVerify` → `buildAuthenticationUrl`, which generates a fresh `PkcePair` and reloads the WebView with a new challenge URL. The WebView is also destroyed on recreation, so no in-flight redirect can outlive the verifier. The "null verifier sent to /token" path can't actually trigger.

**3. Silent decrypt failures** — fixed in `c16e61bb6`. Replaced with `FirebaseCrashlytics.recordException(...)` in both `PreviousUsersUtils.get()` and `getSignedInUser()` so the failure is visible in production rather than just logcat.

**4. `isAvailable()` side effect** — fixed in `c16e61bb6`. Method had no callers; removed it. Same commit also routes the codec's `encrypt`/`decrypt`/`getOrCreateKey` failure paths to Crashlytics for the same observability reason as #3.

